PR #37, opened 2018-04-12
make link optional

quarterto:
nothing is sacred
 🐿 v2.8.0

### DIFF
--- a/packages/x-teaser/src/Link.jsx
+++ b/packages/x-teaser/src/Link.jsx
@@ -1,8 +1,15 @@
 const h = require('@financial-times/x-engine');
 
-const BaseLink = ({url, title, attrs, children}) => <a href={url} title={title} {...attrs}>
-	{children}
-</a>;
+const BaseLink = ({url, title, attrs, children}) => {
+	attrs.href = url;
+	if(title) {
+		attrs.title = title;
+	}
+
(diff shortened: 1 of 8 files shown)


packages/x-teaser/src/Title.jsx
```diff
@@ -1,13 +1,16 @@
 const h = require('@financial-times/x-engine');
-const Link = require('./Link.jsx');
+const Link = require('./Link');
 
-module.exports = ({ title, alternativeTitle, useAlternativeTitle, url, indicators }) => {
+module.exports = ({ title, alternativeTitle, useAlternativeTitle, url, indicators, ...props }) => {
```
i-like-robots: did not know this was a thing, how convenient

packages/x-teaser/src/Link.jsx
```diff
@@ -0,0 +1,14 @@
+const h = require('@financial-times/x-engine');
+
+const BaseLink = ({url, title, attrs, children}) => <a href={url} title={title} {...attrs}>
+	{children}
+</a>;
+
+module.exports = ({ elementOverrides = {}, ...props }) => {
```
i-like-robots: We have a `customSlot` property, what about `customElements`?

PR #53, opened 2018-05-11
Teaser metadata updates

i-like-robots:
Normalises the metadata sub-component props and context props. I've also finally added support for checking the context properly (both ID and label) for the prefix/suffix and link.

Related: https://codepen.io/i-like-robots/full/oddBmj/


components/x-teaser/stories/index.js
```diff
@@ -1,5 +1,9 @@
 const { Teaser } = require('../');
 
+if (window && !/\.ft\.com$/.test(window.location.hostname)) {
```
quarterto: `ReferenceError: window is not defined`

goes to show how careful we need to be about assuming environments 😬

i-like-robots: Strange it didn't blow up my local build 🤷‍♂️ 

Easy to fix at least

PR #63, opened 2018-06-04
is that even a thing?

quarterto:
🐿 v2.8.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "scripts": {
     "lerna": "lerna",
     "postinstall": "lerna bootstrap",
-    "rebootstrap": "lerna clean --yes && lerna bootstrap --no-ci",
+    "rebootstrap": "lerna clean --yes && lerna bootstrap --ci=false",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -2,7 +2,7 @@
   "scripts": {
     "lerna": "lerna",
     "postinstall": "lerna bootstrap",
-    "rebootstrap": "lerna clean --yes && lerna bootstrap",
+    "rebootstrap": "lerna clean --yes && lerna bootstrap --no-ci",
```
chee: is it!??!

quarterto: nope

chee:
`--no-ci`

[source](https://github.com/lerna/lerna/blob/d7e33c6e422934a37302f5e78e226423e3abcf04/integration/lerna-bootstrap.test.js#L19)

quarterto:
...huh. didn't work in my testing. maybe the [`isCI` check](https://github.com/lerna/lerna/blob/5b01599/core/cli/index.js#L58) is overriding that

forcing travis onto npm 5.6 is fine for now

PR #65, opened 2018-06-14
Handlebars redux

i-like-robots:
This refactors the `x-handlebars` package into a generic component loader capable of requiring both x-dash packages and locally defined modules.

Usage:

```handlebars
<!-- load an x-dash package with named export -->
{{{x package="x-teaser" component="Teaser"}}}

<!-- load a local module with default export -->
{{{x local="components/my-thing"}}
```

It supports 3 arguments: `package` (to load a Node modules dependency from the `@financial-times` scope) or `local` to load a locally defined module.

If the loaded module has named exports then the specific function to use may be specified with the `component` argument.

Unlike the previous iteration the Handlebars context (`this` or `.`) does not have to be explicitly defined.

I've also removed the dependency on `hyperons` so that apps can specify their own runtime via `x-engine`.


packages/x-handlebars/concerns/resolve-peer.js
```diff
@@ -0,0 +1,2 @@
+const path = require('path');
+module.exports = (moduleId) => path.join(process.cwd(), 'node_modules', moduleId);
```
quarterto: should this be in a common dependency of x-handlebars and x-engine? or exported from x-engine?

packages/x-handlebars/readme.md
```diff
@@ -25,33 +25,51 @@ If you are building an Express app using [n-ui][n-ui] you can register the helpe
 const express = require('@financial-times/n-ui');
 
 const app = express({
-	helpers: require('@financial-times/x-handlebars')
+	helpers: {
+		// existing helpers
+		...require('@financial-times/x-handlebars')
```
quarterto: object rest/spread is in node 10 but you've said it's node 6 compatible. i know this is just an example but baseline es6 syntax would be clearer i think

packages/x-handlebars/index.js
```diff
@@ -1,3 +1,40 @@
...
+const { render } = require('@financial-times/x-engine');
+const extend = require('./concerns/extend');
+const resolvePeer = require('./concerns/resolve-peer');
+const resolveLocal = require('./concerns/resolve-local');
+
+// This is a regular function expression so that the template context may be shared as "this"
+const x = function ({ hash }) {
+	let moduleId;
+
+	if (hash.hasOwnProperty('package')) {
```
quarterto: this should throw early if both `package` and `local` are missing

packages/x-handlebars/index.js
```diff
@@ -1,3 +1,40 @@
-module.exports = {
-	'x-teaser': require('./helpers/teaser')
+const { render } = require('@financial-times/x-engine');
+const extend = require('./concerns/extend');
```
quarterto: is this necessary if the first argument is a blank object anyway?

packages/x-handlebars/concerns/resolve-peer.js
i-like-robots: If we need it again then I think making it a part of x-engine would be sensible.

packages/x-handlebars/index.js
i-like-robots: Sorry I don't follow? I don't mind switching this to vanilla `Object.assign`

quarterto: `// much faster but much less safe Object.assign() which always creates a new obj`

the only call to this is using a new obj as the first argument anyway

i-like-robots: Ah, yes yes...

quarterto: oh i was actually thinking just go with `Object.assign` for clarity 🤷‍♀️

quarterto: brain is not working today

i-like-robots: Sure, happy to change

PR #78, opened 2018-07-10
X gift article

fenglish:
screen shot 2018-09-18 at 14 28 57

x-gift-article listens a custom event `xDash.giftArticle.activate` to activate.
**How to activate in storybook**
```js
const target = document.getElementById('storybook-preview-iframe').contentWindow.document.body;
target.dispatchEvent(new CustomEvent('xDash.giftArticle.activate'));
```

⚠️  **Need to dispatch the custom event above manually every time move to other gift article story at the moment**


components/x-gift-article/.gitignore
```diff
@@ -0,0 +1 @@
+dist
```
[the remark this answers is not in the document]
quarterto: i didn't want to prescribe any particular structure for a component, but i did want to provide a sensible default. i'm not tied to this though, could you open an issue please?

tools/x-workbench/register-components.js
```diff
@@ -1,4 +1,7 @@
+// TODO Need to change the relative path to '@financial-times/x-gift-article/stories'
+// after releasing new version of x-dash with x-gift-article
 const components = [
+	require('../../components/x-gift-article/stories'),
```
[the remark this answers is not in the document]
quarterto: i'm not sure what you mean?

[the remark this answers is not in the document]
quarterto: oh, instead of using the npm/lerna linked packages? it could but if we ever move the folders around we'd have to update the paths. with lerna links the actual location of the packages doesn't matter, `lerna bootstrap` will be able to locate and link them no matter what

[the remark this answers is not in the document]
dan-searle: Ah, ok. I'm lerna learner still.

components/x-gift-article/.gitignore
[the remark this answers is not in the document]
dan-searle: Sure, no problem.

components/x-gift-article/src/lib/props-composer.js
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	getDefault() {
+		return  {
+			title: this.title,
+			isFreeArticle: this.isFreeArticle,
+			isGift: this.isFreeArticle ? false : true,
+			url: this.isFreeArticle ? this.urls.nonGift : this.urls.dummy,
+			urlType: this.isFreeArticle ? this.urlTypes.nonGift : this.urlTypes.dummy,
+			mailtoLink: this.isFreeArticle ? this.mailtoLinks.nonGift : undefined,
```
benbarnett: Maybe `mailtoUrl`?

components/x-gift-article/src/RadioButtonsSection.jsx
```diff
@@ -0,0 +1,41 @@
...
+import styles from './GiftArticle.css';
+
+const boldTextClassName = styles.bold;
+const radioSectionClassNames = [
+	'o-forms__group',
+	'o-forms__group--inline',
+	styles['radio-button-section']
+].join(' ');
+
+export default ({ isGift, displayGiftUrlSection, displayNonGiftUrlSection }) => (
```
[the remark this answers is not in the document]
benbarnett: The types constants file could be something like:

// lib/constants.js
```js
export const SHARE_TYPE_GIFT = 'gift';
export const SHARE_TYPE_NON_GIFT = 'non-gift';
```

then consumers can just use:

```js
import { SHARE_TYPE_GIFT, SHARE_TYPE_NON_GIFT } from 'lib/constants';

// e.g.

<input checked={ shareType === SHARE_TYPE_GIFT } />
```

components/x-gift-article/src/Form.jsx
```diff
@@ -0,0 +1,46 @@
...
+			}
+
+			<UrlSection
+				type={ props.type }
+				isGift={ props.isGift }
+				isGiftUrlCreated={ props.isGiftUrlCreated }
+				isFreeArticle={ props.isFreeArticle }
+				url={ props.url }
+				urlType={ props.urlType }
+				credit={ props.credit }
```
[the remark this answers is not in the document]
fenglish: 👍

components/x-gift-article/src/GiftArticle.jsx
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+}));
+
+const BaseTemplate = (props) => {
+	if (!propsComposer) {
+		propsComposer = props.composer;
+	}
+
+	if (!hasAttempetedToGetAllowance && !props.isFreeArticle) {
```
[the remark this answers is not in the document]
fenglish: 🙈

components/x-gift-article/src/lib/props-composer.js
fenglish: 👍

components/x-gift-article/readme.md
```diff
@@ -0,0 +1,44 @@
...
+
+Property                  | Type    | Required | Note
+--------------------------|---------|----------|----
+`isFreeArticle`           | Boolean | yes      | Only non gift form is displayed when this value is `true`.
+`articleUrl`              | String  | yes      | Canonical URL
+`articleTitle`            | String  | yes      |
+`articleId`               | String  | yes      | Content UUID
+`sessionId`               | String  | yes      | This is needed to get a gift url.
+`showMobileShareLinks`    | Boolean | no       | This value will be `false` when `nativeShare` is `true`.
+`isCopySupported`         | Boolean | no       | This value will be `false` when `nativeShare` is `true`. Pass whether the browser is supported the copy command or not. e.g. => `document.queryCommandSupported && document.queryCommandSupported('copy')`
```
benbarnett: I wonder if the component could work this out itself?

components/x-gift-article/readme.md
```diff
@@ -0,0 +1,44 @@
...
+### Properties
+
+Property                  | Type    | Required | Note
+--------------------------|---------|----------|----
+`isFreeArticle`           | Boolean | yes      | Only non gift form is displayed when this value is `true`.
+`articleUrl`              | String  | yes      | Canonical URL
+`articleTitle`            | String  | yes      |
+`articleId`               | String  | yes      | Content UUID
+`sessionId`               | String  | yes      | This is needed to get a gift url.
+`showMobileShareLinks`    | Boolean | no       | This value will be `false` when `nativeShare` is `true`.
```
benbarnett: I think that to save complexity, we don't need this option to interact with `nativeShare`. If a consumer wants to turn these buttons on and also have the native share option, that's probably ok?

components/x-gift-article/src/GiftArticle.jsx
```diff
@@ -0,0 +1,117 @@
...
+	emailNonGiftUrl() {
+		tracking.emailLink('nonGiftLink', propsComposer.urls.nonGift);
+	},
+
+	hideCopyConfirmation() {
+		return propsComposer.hideCopyConfirmation();
+	},
+
+	shareByNativeShare() {
+		// TODO display native share
```
benbarnett: 👍  this can just be an event we emit, something like:

```js
document.body.dispatchEvent(new CustomEvent('xdash.giftArticle.share', { detail: { content: { uuid: 'abc-123', url: 'giftArticleUrl' }}, bubbles: true}));
```

(can fix this properly when we integrate though)

components/x-gift-article/src/MobileShareButtons.jsx
```diff
@@ -0,0 +1,71 @@
...
+				</a>
+			</li>
+			<li className={ buttonClassNames } data-share="linkedin">
+				<a className={ linkedinClassNames } rel="noopener" href={ mobileShareLinks.linkedin } data-trackable="linkedin">
+					LinkedIn<span className="o-share__text">(opens new window)</span>
+				</a>
+			</li>
+			<li className={ whatsappButtonClassNames } data-share="whatsapp">
+				<a className={ whatsappClassNames } rel="noopener" href={ mobileShareLinks.whatsapp } data-trackable="whatsapp">
+					Whatsapp<span className="o-share__text">(opens new window)</span>
```
benbarnett: Probably need a space after the name and before the `<span>` (here and in all buttons on this component)

components/x-gift-article/src/lib/api.js
```diff
@@ -0,0 +1,63 @@
...
+					ftSessionSecure: sessionId
+				})
+			});
+
+			const body = response.ok ? await response.json() : {};
+			if (body.errors) {
+				throw new Error(`Failed to get gift article link: ${body.errors.join(', ')}`);
+			}
+
+			return Object.assign({}, body, { redemptionLimit: REDEMPTION_LIMIT });
```
benbarnett: You could use the shorter [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Spread_in_object_literals) here: 

```js
return {
  ...body,
  redemptionLimit: REDEMPTION_LIMIT
}
```

components/x-gift-article/src/lib/props-composer.js
```diff
@@ -0,0 +1,137 @@
...
+			articleUrl: this.articleUrl,
+			sessionId: this.sessionId,
+			showCopyButton: this.showCopyButton,
+			showShareButtons: this.showMobileShareLinks,
+			showNativeShareButton: this.showNativeShareButton,
+			mobileShareLinks: this.mobileShareLinks
+		};
+		const additionalProps = this.isFreeArticle ? this.showNonGiftUrlSection() : this.showGiftUrlSection();
+
+		return Object.assign({}, fundamentalProps, additionalProps);
```
benbarnett: same here about the object spread syntax

components/x-gift-article/src/GiftArticle.jsx
fenglish: I will add this when we integrate to app. Thank you!

components/x-gift-article/src/lib/api.js
fenglish: Oh, I see! thank you!!:simple_smile:

components/x-gift-article/src/GiftArticle.jsx
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+}));
+
+const BaseTemplate = (props) => {
+	if (!propsComposer) {
+		propsComposer = props.composer;
+	}
+
+	document.body.addEventListener('xDash.giftArticle.activate', () => {
```
fenglish: @Financial-Times/x-dash Is there a naming convention for x-dash custom events? `xDash.giftArticle.activate` is ok for you?

components/x-gift-article/src/GiftArticle.jsx
```diff
@@ -0,0 +1,110 @@
...
+
+	shareByNativeShare() {
+		// TODO display native share
+	}
+
+}));
+
+const BaseTemplate = (props) => {
+	if (!propsComposer) {
+		propsComposer = props.composer;
```
quarterto: x-dash components should be free from side effects. this would mean that when you render x-gift-article multiple times on a page (this can happen in the App, which renders three articles at once for the swiping panes), they would all get the same `propsComposer`, is that what you want?

components/x-gift-article/src/GiftArticle.jsx
quarterto: we would prefer x-dash components not to register global event handlers. this should be the responsibility of the app that consumes the component. however, there's currently no pattern for exposing this kind of functionality to the consumer, so we'd need to discuss how to do that.

components/x-gift-article/src/GiftArticle.jsx
```diff
@@ -0,0 +1,103 @@
...
+	}
+}));
+
+const BaseTemplate = (props) => {
+	document.body.addEventListener('xDash.giftArticle.activate', () => {
+		props.isFreeArticle ?
+			props.actions.getShorterNonGiftUrl() : props.actions.getAllowance();
+	})
+
+	return props.isLoading ? <Loading/> : <Form {...props}/>;
```
quarterto: i can't see anywhere where `isLoading` is set, should we be setting in actions?

components/x-gift-article/src/GiftArticle.jsx
quarterto: i'm going to add some kind of "external actions" support to x-interaction, see Financial-Times/x-dash#163 for proposal.

components/x-gift-article/src/GiftArticle.jsx
benbarnett: Oh doesn't this come from x-interaction? https://github.com/Financial-Times/x-dash/blob/master/components/x-interaction/src/InteractionRender.jsx#L18

i-like-robots:
x-gift-article v0.0.2 was released to npm yesterday 🎉

components/x-gift-article/src/lib/share-link-actions.js
```diff
@@ -6,29 +6,30 @@ function createMailtoUrl (articleTitle, shareUrl) {
...
-
-	el.select();                                    // Select the <textarea> content
-	document.execCommand('copy');                   // Copy
-	document.body.removeChild(el);                  // Remove the <textarea> element
-
-	if (selected) {                                 // If a selection existed before copying
-		document.getSelection().removeAllRanges();    // Unselect everything on the HTML document
-		document.getSelection().addRange(selected);   // Restore the original selection
+function copyToClipboard (targetClass = '.js-gift-article__copy-target') {
+	const inputEl = document.querySelector(targetClass);
```
quarterto: this won't work in the App, where there will usually be three gift article components rendered on the page because of article prerendering. this will select the first `.js-gift-article__copy-target` which will be effectively random. instead, you could pass through `event.target` from the action, and use that to find the surrounding `.js-gift-article__copy-target`? (using `Element.prototype.closest` or similar)

fenglish: @quarterto Thank you for your comment! I always forget about App... 🙈  I will do what you suggested 👍

fenglish:
@benbarnett `Element.prototype.closest` is not supported on IE11. However, It works ok on next-article on IE11. Is there the polyfill for `closest`?

benbarnett:
@fenglish good point about polyfill. I think that's ok but perhaps the component should indicate that it makes use of it?

@quarterto do you have any thoughts on this?

components/x-gift-article/src/lib/share-link-actions.js
```diff
@@ -0,0 +1,40 @@
...
+	const subject = encodeURIComponent(articleTitle);
+	const body = encodeURIComponent(shareUrl);
+
+	return `mailto:?subject=${subject}&body=${body}`;
+}
+
+
+function copyToClipboard (event) {
+
+	const urlSection = event.target.closest('fieldset > .js-gift-article__url-section');
```
i-like-robots: I'm struggling to parse this... is it trying to find the closest fieldset parent element or element matching the class name? Either way, is it possible to avoid the combinator?

fenglish: 🙀 Sorry! the combinator is unnecessary!!

benbarnett:
This is now on FT.com article pages 🎉

fenglish:
Thank you for all your help!! 🙇‍♀️

components/x-gift-article/src/GiftArticle.jsx
quarterto: rather than dispatching an event the app should use `x-interaction` to replace this action

PR #59, opened 2018-05-25
x-follow-button

kateloschinina:
Using concept name as button text:
screenshot 2019-02-06 at 16 29 09
screenshot 2019-02-06 at 16 29 03

Using generic button text:
screenshot 2019-02-06 at 16 28 49
screenshot 2019-02-06 at 16 28 55

Currently in use directly by `next-myft-page` and `ft-app`, and as a dependency of `x-topic-search`.

PR #91, opened 2018-07-12
ESLint and Prettier in Harmony

### LINKED ISSUE #85: JSLint and Prettier
We currently install Prettier as a dependency but it is only run in an ad-hoc manner. We should update this to run precommit/prepush, either in fix mode or to highlight issues.

At the same time it would be useful to add JSLint to highlight any silly mistakes or obvious code errors.

https://prettier.io/docs/en/eslint.html

i-like-robots:
Fixes #85 

I haven't enabled the prettier plugin because it needs to do quite a bit of work to make itself happy... one for another PR!

I've tried to fix the eslint errors where possible to do so.


athloi.js
```diff
@@ -134,7 +134,6 @@ module.exports = ({tasks, prompt, addPrompt}) => ({
 		async run(options) {
 			const port = parseInt(url.parse(openUrls.storybook).port, 10);
 			if(await tcpPortUsed.check(port)) {
-				console.log();
```
quarterto: this was there deliberately. i like some whitespace in my cli output 🙂

components/x-interaction/src/InteractionClass.jsx
```diff
@@ -16,13 +16,15 @@ export class InteractionClass extends Component {
 			// additional render will be scheduled.
 			Promise.resolve().then(() => {
 				this.setState({ [loading]: true }, () => {
+					/* eslint no-console:off */
 					console.log('didsetloading', this.state[loading]);
```
quarterto: these shouldn't be here but they're removed in #59

tools/x-docs/src/components/search/index.js
```diff
@@ -62,9 +65,9 @@ export default class Search extends Component {
 		</div>;
 	}
 
-	search = ev => {
+	search (ev) {
```
quarterto: this will need to be bound now

packages/x-logo/index.js
```diff
@@ -72,25 +72,20 @@ const xPoints = ({x, y, width, height, thickness}) => {
 }
 
 export default class XLogo extends Component {
-	static defaultProps = {
```
quarterto: these missing default props are breaking the build

quarterto:
was just about to approve but noticed the base branch is `matth/migration-guide-tweaks` which was merged, can you update to master plz

i-like-robots:
Woah, good spot!

quarterto:
i mean, it had all the _letters_ of `master` in it, just a few others as well

PR #92, opened 2018-07-16
Update cross-linked dependencies to use relative file references

i-like-robots:
So we can use `npm install` without worrying about any extra symlinking steps. Lerna should rewrite these references on publish with the given version numbers.

 🐿 v2.10.0


i-like-robots:
Looks like this functionality was never merged into Lerna (https://github.com/lerna/lerna/pull/425) although it is the default behaviour of NPM 5+.

Will close those in favour of a PR to migrate to Athloi.

PR #93, opened 2018-07-16
document adding a component to storybook

quarterto:
this is a previously undocumented requirement that caused @fenglish some trouble


fenglish:
@quarterto Thank you for adding the information!!

PR #94, opened 2018-07-16
Correct code error on x-dash-engine readme example

maxcbc:


quarterto:
good spot, thanks @maxcbc

PR #98, opened 2018-07-17
move most gitignore stuff to toplevel, fixes #88

### LINKED ISSUE #88: Main .gitignore to include some component folder patterns
I understand and agree with the intention of not being too prescriptive about component folder structures, but I think there are some common paths, that are prescribed, that could be added to the main `.gitignore` file and remove the need for individual component `.gitignore` files in many cases:

```
components/*/.DS_Store
components/*/dist
components/*/node_modules
components/*/package-lock.json
```

quarterto:
@dan-searle most patterns in your suggested `.gitignore` were already covered by our toplevel one, so i've added `dist` and cleaned things up a bit


tools/x-docs/.gitignore
```diff
@@ -1,9 +1,4 @@
-# Project dependencies
-# https://www.npmjs.org/doc/misc/npm-faq.html#should-i-check-my-node_modules-folder-into-git
 .cache
-node_modules
 bower_components
```
i-like-robots: Move to top level?

quarterto: i didn't want to encourage this sort of thing

i-like-robots:
Needs a rebase but +11

PR #99, opened 2018-07-18
Switch to FT athloi

### LINKED ISSUE #90: Monorepo management tooling
So far we have been using Lerna to manage the installation and publishing of the multiple packages in this repo. We chose Lerna because it is the de-facto monorepo management tool available for Node and it is used by many very popular projects.

However, so far we have been disappointed and oftentimes frustrated by it. It is relatively slow and when it breaks - which it does too often to give us confidence - it usually does so silently. This has been a common sight:

screen shot 2018-07-19 at 15 47 13

The problems can be very hard to debug because Lerna has an extremely complicated architecture. It is not a project we can easily dig into and help to support. For these reasons we have investigated the alternatives, including:

- https://www.npmjs.com/package/@microsoft/rush
- https://www.npmjs.com/package/oao
- http://boltpkg.com/

All 3 are solid projects and differ little in configuration and usage. But, they all use Yarn and/or hoisting to minimise the size and number of packages installed. This is a good idea on paper but in practice it causes havoc because Node's module resolution algorithm depends so much on context, and that context is lost when not managing packages on an individual basis.

All of the tools discussed above include deep integration with Git with many checks and balances to ensure releases are valid, change logs are updated etc. We release on CI so these complicated features are of no use to us at this time.

As far as I can tell there are 3 jobs that a monorepo management tool needs to do for us:

- Install external dependencies and cross-link local packages
- Publish all packages in lock-step
- Run a given script in the context of every package

A script can be run in the context of each package with a few lines of bash, for example to run `npm install`:

```sh
targets=$(find {components,packages,tools}/x-*/package.json)

for $item in $targets; do
    target=$(dirname $item)
    (cd $target && exec npm install)
done
```

However because packages may depend on each other several types of task must be performed _in order_. This means finding the cross linked dependencies and sorting them. This is [not too hard](https://gist.github.com/i-like-robots/79cc1c647a941dbdeeebd28e2806f8a1) but it is important to see problems through this lens.

Some monorepo projects are managed using a set of custom scripts. This is transparent but we have always had issues maintaining homebrew tooling in Next and people tend to feel especially burnt when it fails and they have to spend time yak shaving. For this reason if we do decide to roll-our-own then it would be sensible to keep it to the absolute minimum. Some examples of this approach include:

- [cycle.js](https://github.com/cyclejs/cyclejs/tree/master/.scripts)
- [React](https://github.com/facebook/react/tree/master/scripts)

Thoughts?

i-like-robots:
Removes Lerna in favour of FT flavour Athloi. Fixes #90.

See the [athloi docs](https://github.com/Financial-Times/athloi) for more information.


package.json
```diff
@@ -1,43 +1,28 @@
...
+    "clean": "git clean -fxdi",
+    "build": "athloi run build",
     "jest": "jest -c jest.config.js",
     "test": "npm run lint && npm run jest",
     "lint": "eslint . --ext=js,jsx",
-    "start": "athloi",
     "precommit": "secret-squirrel",
     "commitmsg": "secret-squirrel-commitmsg",
-    "postcheckout": "athloi cleanup"
+    "run-workbench": "(cd tools/x-workbench && npm start)",
```
quarterto: something i built for a side project last night, but would be useful here is `@quarterto/open-once-listening` which would let you do

```
"prerun-workbench": "open-once-listening http://local.ft.com:9001 &",
"prerun-docs": "open-once-listening http://local.ft.com:8080 &",
```

PR #103, opened 2018-07-19
ordering for docs pages

quarterto:


i-like-robots:
Lint error by the looks of it!

PR #109, opened 2018-07-24
add fetchMock property to story modules

fenglish:


tools/x-workbench/.storybook/build-story.js
```diff
@@ -56,6 +57,11 @@ function buildStory ({package: pkg, dependencies, component: Component, knobs, s
 	storybook.add(story.title, () => {
 		const props = createProps(story.data, story.knobs, knobs);
 
+		if (story.fetchMock) {
+			fetchMock.restore(); // to isolate the mocks to each story
```
benbarnett: Thinking about it more, this may cause us trouble if we have multiple stories with mocks, as they're all working with the same instance of `fetchMock`. I've not used it before but we might want to consider [`fetchMock.sandbox()`](http://www.wheresrhys.co.uk/fetch-mock/api#sandbox)

fenglish:
@i-like-robots x-gift-article PR has the almost same change. However, Is it ok if we separate this PR from x-gift-article? This fetch mock module is not only for x-gift-article and this PR includes readme change. I think it will be clearer if we keep this PR separate from x-gift-article PR.

i-like-robots:
That's absolutely fine, thanks for clearing it up.

PR #127, opened 2018-08-07
X eventpromo

ludrob:
This is a port of https://github.com/Financial-Times/n-eventpromo to x-dash.

- Eventpromos are manage by live.ft.com, see: https://live.ft.com/Events/2018/FT-Future-of-Retail-Summit
- We use https://github.com/Financial-Times/next-eventpromo-api to sync/get the data from their system
- Currently, eventpromos look like this (look just after the article content): https://www.ft.com/content/b47de80a-9989-11e8-9702-5946bae86e6d
- The data available for the eventpromo can be accessed here: https://www.ft.com/eventpromo/api

Note: we called this 'eventpromo' to prevent any issues with 'event' and make it easily searchable.


jest.config.js
```diff
@@ -1,4 +1,6 @@
 module.exports = {
 	testMatch: ['**/__tests__/**/*.test.js'],
-	testPathIgnorePatterns: ['/node_modules/', '/bower_components/']
+	testPathIgnorePatterns: ['/node_modules/', '/bower_components/'],
+	verbose: true,
+	testURL: "http://localhost/",
```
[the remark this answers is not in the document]
i-like-robots: This change is in a few branches that I've seen, I'm not sure why it isn't in master yet! But thank you for flagging.

components/x-eventpromo/stories/index.js
```diff
@@ -0,0 +1,35 @@
...
+		return text('Dates', data.dates);
+	},
+	eventpromoId() {
+		return text('Eventpromo Id', data.eventpromoId);
+	},
+	eventpromoLink() {
+		return text('Eventpromo link', data.eventpromoLink);
+	},
+	eventpromoTitle() {
+		return text('Eventpromo title', data.eventpromoTitle);
```
i-like-robots: can these props not be named `id`, `link` (or `url`) and `title` rather than requiring the `eventpromo` prefix?

components/x-eventpromo/stories/eventpromo.js
```diff
@@ -0,0 +1,27 @@
...
+};
+
+exports.knobs = [
+	'eventpromoId',
+	'eventpromoLink',
+	'eventpromoTitle',
+	'dates',
+	'image1',
+	'image2',
+	'image3',
```
i-like-robots: It's not clear to me what the function of the different images is, do they have specific uses?

components/x-eventpromo/src/Meta.jsx
```diff
@@ -0,0 +1,12 @@
+import { h } from '@financial-times/x-engine';
+
+const Meta = ({location, dates}) => {
+	return (
+		<div className="event-promo__meta">
+			<p>{location}</p>
```
i-like-robots: Should this be an `<address>` element or is it just a word or two?

components/x-eventpromo/src/Image.jsx
```diff
@@ -0,0 +1,16 @@
...
+import Link from './Link';
+
+const Image = ({linkUrl, imageUrl}) => {
+
+	const baseSrc = `https://www.ft.com/__origami/service/image/v2/images/raw/${imageUrl}?source=next&amp;fit=cover&amp;compression=best&amp;`;
+	const imgSrc = `${baseSrc}width=340 340w,${baseSrc}width=400 740w`;
+
+	return (
+		<Link url={linkUrl} attrs={{ 'className': 'js-event-promo--animate', 'data-trackable': 'event-promo'}}>
+			<img className="js-event-promo--animate event-promo-inarticle__img fade-0" srcSet={imgSrc} alt="" data-n-image-lazy-load-js="" role="presentation" styles={{size: '(max-width: 46.24em) 100vw, 45vw'}} />
```
i-like-robots: Not sure what's going on with the `styles` property here, do you mean `sizes`?

i-like-robots: Because the image has no content for an assistive device to read it may be best to hide it from them by adding `aria-hidden="true"` and `tabindex="-1"` to the link

components/x-eventpromo/src/ImagesContainer.jsx
```diff
@@ -0,0 +1,16 @@
+import { h } from '@financial-times/x-engine';
+import Image from './Image';
+
+const ImagesContainer = ({eventpromoLink, image1, image2, image3}) => {
+
+	return (
+		<div className="event-promo-inarticle__blocks event-promo-inarticle__img-block">
+			<span className="event-promo__control js-event-promo__control" aria-label="control animation"></span>
```
i-like-robots: Should this be an interactive element such as a button?

components/x-eventpromo/src/Meta.jsx
ludrob: It's just free text, it can be anything. 
Usually it's something like, "London", "New York", "ExCeL London", etc

components/x-eventpromo/stories/eventpromo.js
ludrob: I'll add a bit more details in the PR description.
The eventpromos always have exactly 3 images, used to create and fade in/out animation.
They all have the same function.

components/x-eventpromo/src/ImagesContainer.jsx
ludrob: it's the play/pause button overlay-ed on top of the animated images, @Thurst-ft , any reason we should not make this a button?

components/x-eventpromo/readme.md
```diff
@@ -0,0 +1,16 @@
...
+
+This module is compatible with Node 6+ and is distributed on npm.
+
+```bash
+npm install --save @financial-times/x-eventpromo @financial-times/x-engine
+```
+
+This module also requires [`x-engine`][engine] as a peer dependency. The Engine module is used to inject your chosen runtime into the component. Please read the `x-engine` documentation first if you are consuming `x-` components for the first time in your application.
+
+[engine]: https://github.com/Financial-Times/x-dash/tree/master/packages/x-engine
```
i-like-robots: Usage guidelines and a list of accepted properties needs to be added, see [the readme blueprint](https://github.com/Financial-Times/x-dash/blob/master/private/blueprints/component/readme.md) or check x-teaser for reference.

components/x-eventpromo/src/Footer.jsx
```diff
@@ -0,0 +1,21 @@
...
+	return (
+		<div className="event-promo-inarticle__cta">
+			<div className="event-promo-inarticle__btn-block">
+				<Link url={url} attrs={{
+					'className': 'event-event-promo-inarticle__btn',
+					'data-trackable': 'event-promo'
+				}}>Register now</Link>
+			</div>
+			<div className="event-promo-inarticle__brand">
+				<span>Presented by</span>
```
i-like-robots: I think this `<span>` may be removed

components/x-eventpromo/src/ImagesContainer.jsx
```diff
@@ -0,0 +1,16 @@
+import { h } from '@financial-times/x-engine';
+import Image from './Image';
+
+const ImagesContainer = ({link, image1, image2, image3}) => {
```
i-like-robots: Perhaps this should be an array of image URLs?

components/x-eventpromo/src/Link.jsx
```diff
@@ -0,0 +1,9 @@
+import { h } from '@financial-times/x-engine';
+
+const Link = ({url, attrs = {}, children}) => {
+	return (
+		<a href={url} {...attrs}>{children}</a>
```
i-like-robots: Does this need to be a HOC (could components implement the `<a>` anchor element themselves?)

components/x-eventpromo/src/Details.jsx
```diff
@@ -0,0 +1,23 @@
+import { h } from '@financial-times/x-engine';
+import Footer from './Footer';
+import Link from './Link';
+import Meta from './Meta';
+
+const Details = ({link, title, ...props}) => {
+
+	return (
+		<div className="event-promo-inarticle__blocks">
```
i-like-robots: I think this needs a heading element (even if hidden and referenced using an `aria-labelledby` attribute) to identify this content

ludrob:
I should now have addressed all the comments on the PR.
My react is a bit rusty and my accessibility skills subpar so do not hesitate to educate me :)

Styles are still sitting in the n-eventpromo repo (at least for now).
We need to decide what we do with them (css in js, outside js, keep in separate repo, etc).
Question: is it acceptable to merge this in the current state and deal with styles later?
Note that the class names should all be used either by styles or js.

components/x-eventpromo/src/Details.jsx
```diff
@@ -0,0 +1,23 @@
+import { h } from '@financial-times/x-engine';
+import Footer from './Footer';
```
Thurst-ft: Not sure about this being named 'Footer' but not sure what else to call it

components/x-eventpromo/src/Image.jsx
```diff
@@ -0,0 +1,25 @@
+import { h } from '@financial-times/x-engine';
+
+const Image = ({ linkUrl, imageUrl }) => {
+	const baseSrc = `https://www.ft.com/__origami/service/image/v2/images/raw/${imageUrl}?source=next&amp;fit=cover&amp;compression=best&amp;`;
+	const imgSrc = `${baseSrc}width=340 340w,${baseSrc}width=400 740w`;
+
+	return (
+		<a href={linkUrl}
```
Thurst-ft: formattting looks a bit odd

components/x-eventpromo/src/Meta.jsx
Thurst-ft: It is likely to just be a place name and City at most but I don't see any harm in using <address>

__tests__/__snapshots__/snapshots.test.js.snap
```diff
@@ -1,5 +1,131 @@
...
+    />
+    <a
+      aria-hidden="true"
+      className="js-event-promo--animate"
+      data-trackable="event-promo"
+      href="https://live.ft.com/Events/2018/FT-Brexit-and-Beyond-Summit?segmentId=some-segment-id&amp;variant=false"
+      tabIndex="-1"
+    >
+      <img
+        alt=""
```
Thurst-ft: alt should have event  eventTitle as value

components/x-eventpromo/src/Image.jsx
```diff
@@ -0,0 +1,25 @@
...
+						className="js-event-promo--animate"
+						data-trackable="event-promo"
+						aria-hidden="true"
+						tabIndex="-1"
+		>
+			<img
+				className="js-event-promo--animate event-promo-inarticle__img fade-0"
+				srcSet={imgSrc} alt=""
```
Thurst-ft: alt needs a value probably eventTitle and possibly the strapline

components/x-eventpromo/stories/eventpromo.js
```diff
@@ -0,0 +1,27 @@
...
+	'dates': '12 November - 14 November 2018',
+	'id': 'fancyEventId',
+	'images': [
+		'https://live.ft.com/var/ftlive/storage/images/events/2018/ft-brexit-and-beyond-summit/888258-7-eng-GB/FT-Brexit-and-Beyond-Summit.png',
+		'https://www.ft.com/__assets/creatives/better-promo/audiance_clapping.jpg',
+		'https://www.ft.com/__assets/creatives/better-promo/break_out.jpg',
+	],
+	'link': 'https://live.ft.com/Events/2018/FT-Brexit-and-Beyond-Summit?segmentId=some-segment-id&amp;variant=false',
+	'location': 'planet earth',
+	'title': 'FT Brexit and Beyond Summit'
```
Thurst-ft: Should we also inlcude the strapline in to this iteration?

components/x-eventpromo/src/Details.jsx
```diff
@@ -0,0 +1,23 @@
...
+import Meta from './Meta';
+
+const Details = ({link, title, ...props}) => {
+
+	return (
+		<div className="event-promo-inarticle__blocks" aria-labelledby="details-header">
+			<h1 id="details-header" style={{display: 'none'}}>Event details and information</h1>
```
i-like-robots: This isn't the main page headline. Without knowing the context it will sit in it's hard to choose the right level though, but as it's used for labeling rather than a landmark (display none will remove it from the accessibility tree) a `<div>` will probably do, `<div hidden>` would also work.

components/x-eventpromo/src/Image.jsx
i-like-robots: The image is marked as for presentation only so blank alt text is recommended, see https://www.w3.org/WAI/tutorials/images/decorative/

ludrob:
I've now merged everything (style changes) into x-eventpromo, this PR now has everything needed to render the x-eventpromo component.

tools/x-workbench/package.json
```diff
@@ -5,7 +5,7 @@
   "description": "",
   "main": "register-components.js",
   "scripts": {
-    "start": "start-storybook -p 9001 -c .storybook -s static -h local.ft.com",
```
i-like-robots: Please leave this in, there are some components that will fail due to CORS issues without it.

ludrob:
@i-like-robots 
I've updated and prettified everything, happy for us to create a pre-release?
@Thurst-ft 
Strapline changes have now been added to this PR

quarterto:
From discussion with @ludrob, gonna close this, development is ongoing in `n-eventpromo` and any `x-` version would be ported from that

PR #129, opened 2018-08-08
reinstate data-content-id

quarterto:
likely needed for syndication icon

### DIFF
--- a/__tests__/__snapshots__/snapshots.test.js.snap
+++ b/__tests__/__snapshots__/snapshots.test.js.snap
@@ -43,7 +43,6 @@ exports[`@financial-times/x-styling-demo renders a default Styling x-styling-dem
 exports[`@financial-times/x-teaser renders a Hero Article x-teaser 1`] = `
 <div
   className="o-teaser o-teaser--article o-teaser--hero o-teaser--has-image o-teaser--highlight js-teaser"
-  data-id=""
 >
   <div
     className="o-teaser__content"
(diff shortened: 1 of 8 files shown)


i-like-robots:
Just a note: The content ID is already a data attribute on the container so perhaps that should be removed so that this is canonical?

i-like-robots:
Note that implementers must also include the o-teaser syndication styles (which i have failed to do on search) https://github.com/Financial-Times/o-teaser/blob/16ea9398e306d052c838b551e87f34e06384518b/src/scss/_variables.scss#L8

leggsimon:
I'm not sure whether we use origami's syndication styles...

i-like-robots:
From what i can see either implementation (`data-id` on container or `data-content-id` on title link) should work OK:

https://github.com/Financial-Times/n-syndication/blob/master/src/js/config.js#L25-L32

I can see as well that syndication styles are also included in `n-syndication` so I'll raise the o-teaser PR to remove it from there.

i-like-robots:
So we worked out how `n-syndication` works. It appears the intention is that components and apps do not need to know about the presence of the `n-syndication` script and as such it implements a set of specific rules to target different elements on the page.

So... we have the option to do what we want, either:

- we maintain the current compatibility with n-teaser and merge this
- we add a new rule to `n-syndication` if we think the `[data-id]` implementation is "better"

Personally prefer the latter because hoisting the ID to the top-level element increases its visibility, but it's such a minor detail I'm not going to sweat it either way. Whatever the decision the link between the two needs to be documented.

leggsimon:
I'm happy with the latter. `x-teaser` is being used on the search page right? We can test it out there?

quarterto:
okay, imma close this then

PR #130, opened 2018-08-08
Display subheadings on the sidebar

kateloschinina:
## What is does
Displays subheadings on the sidebar, like this:
screen shot 2018-08-08 at 22 22 40

screen shot 2018-08-13 at 18 47 47

## How it works
Gets markdown information via query using `allMarkdownRemark` function, and then displays it in relevant places.
 - Any depth is displayed, let's hope if stays visually appealing
 - Every subheading is a link to corresponding part of the text (transitions are smoooooth)
 - We can see subheadings of the active heading only

### DIFF
--- a/tools/x-docs-2/src/components/sidebar/module-list.jsx
+++ b/tools/x-docs-2/src/components/sidebar/module-list.jsx
@@ -1,8 +1,11 @@
 import React from 'react';
 import { Link } from 'gatsby';
 
+import GithubSlugger from 'github-slugger';
+
 const createUrlFromName = (name) => {
-	return name.toLowerCase().replace(/ /g, '-').replace(/[?!:%$'.,"…:]/g, '');
+	const slugger = new GithubSlugger();
+	return slugger.slug(name)
 }
 
 const buildListItem = (item, currentNode) => {
--- a/tools/x-docs-2/src/components/sidebar/module-list.jsx
+++ b/tools/x-docs-2/src/components/sidebar/module-list.jsx
@@ -29,18 +29,25 @@ const buildListItem = (item, currentNode) => {
 	);
 }
 
-const buildSubheadingsList = (menu, currentNode) => {
-	return menu.headings.map(item => {
-		if (item.depth === 2) {
-			return buildListItem(item, currentNode);
-		} else if (item.depth === 3) {
+const recursiveBulletsGenerator = (item, prevDepth, currentNode) => {
+	if (item.depth > 1) {
+		if (item.depth > prevDepth) {
 			return (
 				<ul key={item.value}>
-					{ buildListItem(item, currentNode) }
+					{ recursiveBulletsGenerator(item, prevDepth + 1, currentNode) }
 				</ul>
 			);
-		}
-		return null;
+		} else return buildListItem(item, currentNode); 
+	} else return null;
+}
+
+const buildSubheadingsList = (menu, currentNode) => {
+	return menu.headings.map(item => {
+		return (
+			<ul key={item.value}>
+				{ recursiveBulletsGenerator(item, 2, currentNode) }
+			</ul>
+		);
 	});
 }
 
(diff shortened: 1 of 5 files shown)


i-like-robots:
Niiiiice. Sorry about the data changes yesterday, I thought I'd been quick enough to get them in first!

Shall we pair on this for a bit today?

@quarterto what do you think about displaying the subheadings like this?

quarterto:
i like this a lot! thanks @kateloschinina ❤️

tools/x-docs-2/src/components/sidebar/module-list.jsx
```diff
@@ -1,26 +1,28 @@
 import React from 'react';
 import { Link } from 'gatsby';
 
+import GithubSlugger from 'github-slugger';
+
 const createUrlFromName = (name) => {
-	return name.toLowerCase().replace(/ /g, '-').replace(/[?!:%$'.,"…:]/g, '');
+	const slugger = new GithubSlugger();
```
i-like-robots: if you hoist this into module scope you'll only need to contruct a single instance =]

tools/x-docs-2/src/components/sidebar/module-list.jsx
```diff
@@ -29,15 +31,15 @@ const buildSubheadingsList = (menu, currentNode) => {
...
-		<li className="site-sidebar__item site-sidebar__item--title">{title}</li>
-		{data.map(({ node }, i) => (
+		<li className="site-sidebar__item site-sidebar__item--title">{title || pageContext.source}</li>
+		{ data.map(({ node }, i) => (
 			<li key={`list-${i}`} className="site-sidebar__item site-sidebar__item--link">
 				<Link to={node.path} exact activeClassName="is-active">
 					{node.context.title}
 				</Link>
-				{ menu && node.path === document.location.pathname ? (
+				{ menu && menu.headings.length > 0 && pageContext && pageContext.source && pageContext.title && node.path === `/${pageContext.source}/${pageContext.title}` ? (
```
i-like-robots: If we were to change the logic we use to build slugs this would break. Instead of re-constructing the slug here see if you get access to the `location` prop passed to the page instead =]

tools/x-docs-2/src/components/sidebar/module-list.jsx
kateloschinina: The problem is that when it constructs an instance and then applies it to same string it checks it for uniqueness, like this:
```
const slugger = new GithubSlugger();
slugger.slug('url one') // url-one
slugger.slug('url one') // url-one-2
```
So we need to create a new instance every time =(

tools/x-docs-2/src/components/sidebar/module-list.jsx
kateloschinina: That makes sense =)

tools/x-docs-2/src/components/sidebar/module-list.jsx
i-like-robots: Ah OK, definitely worth adding a comment to explain this!

tools/x-docs-2/src/components/sidebar/module-list.jsx
```diff
@@ -1,17 +1,65 @@
...
+	// const slugger = new GithubSlugger();
+	// slugger.slug('url one') // url-one
+	// slugger.slug('url one') // url-one-2
+	// therefore there is a need to create a new class instance every time when the function is applied
+	return slugger.slug(name)
+}
+
+const reactToClick = (e, link) => {
+	e.preventDefault();
+	document.querySelector(`#${link}`).scrollIntoView({
```
i-like-robots: Niiice, but we can go further! The event itself gives us a reference to the link, and the link gives us the ID =]

https://developer.mozilla.org/en-US/docs/Web/API/Event/target
https://developer.mozilla.org/en-US/docs/Web/API/HTMLHyperlinkElementUtils/hash

tools/x-docs-2/src/components/sidebar/module-list.jsx
```diff
@@ -1,17 +1,65 @@
...
+		behavior: 'smooth'
+	});
+}
+
+const buildListItem = (item, currentNode) => {
+	const url = createUrlFromName(item.value);
+	return (
+		<Link to={`${currentNode.path}#${url}`} key={item.value} onClick={(e) => reactToClick(e, url)}>
```
i-like-robots: No need for the path here, the hash - or fragment - works alone ✨ 

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-href

kateloschinina: Wow, that is incredible!

tools/x-docs-2/src/components/sidebar/module-list.jsx
kateloschinina: magic!

PR #131, opened 2018-08-09
Document how to configure `n-ui` to support `x-engine` .jsx files on the client side

maxcbc:
# Document how to configure n-ui to support x-engine .jsx files on the client side

## What
_**This is a documentation change.**_
This adds documentation describing how to configure [n-ui](https://github.com/Financial-Times/n-ui) to compile `x-engine` based `.jsx` files for use on the client side.

## Why
Because [n-ui](https://github.com/Financial-Times/n-ui) means that consuming projects webpack configs are a step removed from them, it means that these project require a slightly different configuration.

## Actual process for supporting on n-ui
In order to configure n-ui to compile x-dash based `.jsx` files you will also need to add the following to `n-ui-build.config.js`.
 ```js
// n-ui-build.config.js
const xEngine = require('@financial-times/x-engine/src/webpack');
 module.exports = {
	plugins: [
		xEngine()
	],
	pragma: 'h'
};
```
 You will also need to import `h` from `x-engine` at the top of each `.jsx` file.
```javascript
import {h} from '@financial-times/x-engine'
```

### DIFF
--- a/packages/x-engine/readme.md
+++ b/packages/x-engine/readme.md
@@ -119,6 +119,26 @@ export default const TeaserList = (props) => (
 );
 ```
 
+#### Client-side through n-ui
+In order to configure n-ui to compile x-dash based `.jsx` files you will also need to add the following to `n-ui-build.config.js`.
+
+```js
+// n-ui-build.config.js
+const xEngine = require('@financial-times/x-engine/src/webpack');
+
+module.exports = {
+	plugins: [
+		xEngine()
+	],
+	pragma: 'h'
+};
+```
+
+You will also need to import `h` from `x-engine` at the top of each `.jsx` file.
+```javascript
+import {h} from '@financial-times/x-engine'
+```
+
 ## FAQ
 
 ### This sounds complicated… is it a magic black box?


packages/x-engine/readme.md
```diff
@@ -119,6 +119,26 @@ export default const TeaserList = (props) => (
...
+	plugins: [
+		xEngine()
+	],
+	pragma: 'xEngine.h'
+};
+```
+
+You will also need to import `x-engine` at the top of each `.jsx` file.
+```javascript
+import xEngine from 'x-engine'
```
i-like-robots: xEngine doesn't have a default export so I believe this should be `* as xEngine` or use named imports such as `import { h } from '@financial-times/x-engine'`

packages/x-engine/readme.md
```diff
@@ -119,6 +119,26 @@ export default const TeaserList = (props) => (
...
+
+```js
+// n-ui-build.config.js
+const xEngine = require('@financial-times/x-engine/src/webpack');
+
+module.exports = {
+	plugins: [
+		xEngine()
+	],
+	pragma: 'xEngine.h'
```
i-like-robots: This may just be `'h'`

packages/x-engine/readme.md
i-like-robots: Confirmed: could we change this to advise using named imports?

maxcbc: Will do this, though n-ui makes it work anyway.

PR #137, opened 2018-08-14
Remove obsolete syndication props

### LINKED ISSUE #136: Remove canBeSyndicated and canBeDistributed props from x-teaser
These properties are not used by the teaser component, only via `n-syndication` which pulls this information from an API. At the moment these "indicator" props append class names to each teaser but these are currently just wasted bytes.

These will also need to be removed from [the data store](https://github.com/Financial-Times/next-es-interface/blob/master/server/transformers/plugins/concerns/base-teaser.js#L27-L28).

leggsimon:
Fixes #136 

These properties are no longer used

### DIFF
--- a/components/x-teaser/Props.d.ts
+++ b/components/x-teaser/Props.d.ts
@@ -131,7 +131,6 @@ export interface Media {
 
 export interface Indicators {
 	canBeDistributed: 'yes' | 'no' | 'verify';
-	canBeSyndicated: 'yes' | 'no' | 'verify' | 'withContributorPayment';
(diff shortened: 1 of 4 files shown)


leggsimon:
Not sure about updating this mention of syndication from the README 
> ### Indicators
>
>Teasers display content but our content items are also decorated with hints and traits which further describe them. Content items may only be available to premium subscribers, excluded from syndication licenses, or selected as an editor's choice. 

https://github.com/Financial-Times/x-dash/blob/master/components/x-teaser/readme.md#L37

PR #138, opened 2018-08-14
add testURL property

leggsimon:
Adds the `testURL` property to the Jest config. To fix the following error with the latest version of JSDom. (Further details https://github.com/facebook/jest/issues/6766)
```
FAIL
● Test suite failed to run

SecurityError: localStorage is not available for opaque origins at Window.get localStorage [as localStorage] (node_modules/jsdom/lib/jsdom/browser/Window.js:257:15)
      at Array.forEach (<anonymous>)
```

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,5 +1,5 @@
 module.exports = {
-	testURL: 'http://localhost/',
+	testURL: 'https://local.ft.com/',
 	testMatch: ['**/__tests__/**/*.test.js'],
 	testPathIgnorePatterns: ['/node_modules/', '/bower_components/']
 };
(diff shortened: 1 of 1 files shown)


quarterto:
do we want to do it this way, or using `testEnvironment: 'node'`?

quarterto:
actually let's do it like this but make it `https://local.ft.com` to also squash the video cors warning? i think that should work

leggsimon:
will being `https` have bad effects? 🤔

quarterto:
who knows!

i-like-robots:
Hopefully JSDOM would only care about that if we were mocking location APIs etc

i-like-robots:
For now it literally needs a URL because localstorage is not available on `about:blank` 😆

Issue #124, opened 2018-08-01
Allow teaser to accept title and standfirst markup


i-like-robots:
The original `n-teaser` component allowed implementers to render unescaped values for the `title` and `standfirst` properties. This functionality has recently been removed due to concerns with rendering content sourced from 3rd parties (https://github.com/Financial-Times/n-teaser/pull/195).

However, there are cases where this feature is desirable - for example highlighting matching terms on the search page - and so it may be useful to allow an option to toggle it on.

quarterto:
could we just make people who have this use case inject their own `Title` component? might need first class support for that as with `Link`

i-like-robots:
Here is a temporary workaround from @chee:

```js
const { h } = require('@financial-times/x-engine');

const wrap = (html) => { dangerouslySetInnerHTML: { __html: html }  };

const teaser = { … };

const htmlTeaser = {
 ...teaser,
 standfirst: h('span', wrap(teaser.standfirst))
};
```

i-like-robots:
Fixed in #139

PR #139, opened 2018-08-20
Don't coerce `displayTitle` to a string

chee:
so the title can be supplied as a component, as per https://github.com/Financial-Times/x-dash/issues/124#issuecomment-414339312

if the feature of string coercion was not documented, then this would be a patch version.

--

edit: @i-like-robots says it's possible the whitespace isn't needed anymore at all. i'll confirm this locally and supply another commit if that's the case.


i-like-robots:
Just need to update the snapshots then +1

chee:
snoopshoots are up to doot

PR #140, opened 2018-08-21
Add space after relative time prefix

adgad:
Add a space after the prefix text for relative timestamps, as they are currently bunched together.

Is currently: 

screen shot 2018-08-21 at 14 04 54


components/x-teaser/src/RelativeTime.jsx
```diff
@@ -21,7 +21,7 @@ export default ({ publishedDate, firstPublishedDate }) => {
 
 	return isRecent(relativeDate) ? (
 		<div className={`o-teaser__timestamp o-teaser__timestamp--${status}`}>
-			{status ? <span className="o-teaser__timestamp-prefix">{status}</span> : null}
+			{status ? <span className="o-teaser__timestamp-prefix">{status} </span> : null}
```
i-like-robots: `{` ${status} `}`

PR #149, opened 2018-08-31
Include package lock files

### LINKED ISSUE #148: Commit package-lock.json
Now that we are approaching some stability I think it's about time we start committing the `package-lock.json` file generated each time we run `npm install`.

We can use this file to speed up installs both locally and on CI (`npm ci`) and improve dependency caching.

i-like-robots:
Fixes #148 

In this PR I have committed the lockfile for each package. Testing locally having lockfiles available reduces the install time from 122 seconds to under 40 seconds.

This has required the removal of Secret Squirrel as keeping it would require the addition more than 22000 rules! (https://github.com/Financial-Times/secret-squirrel/issues/95)

I have also taken the opportunity to try out the `npm ci` command which promises to be [even faster](https://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable) but this has meant bumping the Node version to 10+ for the CI environment.


i-like-robots:
Closing in favour of #150

PR #150, opened 2018-08-31
Package lock again

i-like-robots:
Following on from #149 this is my second attempt to commit package lock files. It works OK locally and (eventually) on CI using Node 8.11 and NPM 5.6.

However, I have so far not gotten it to work with Node 10.9 and NPM 6.2 either locally or on CI, my machine or otherwise.

The process for creating them was as follows:

1. run `make clean` to remove all installed modules
2. remove the `--no-package-lock` from the install make task and top-level postinstall script
3. run `make install`
4. Add `--no-save` flag to the install make task and top-level postinstall script to prevent future installs updating the lockfile
5. commit

To update the lock files these are the steps I have been taking:

1. run `make clean`
2. run `athloi exec rm package-lock.json` to remove all the lock files
3. run `athloi exec npm i` to regenerate them


i-like-robots:
This has been tested by several more folks with decent results so I'll merge this. If it does begin to cause any issues then it is straightforward to undo.

i-like-robots:
OK, 3... 2... 1...

PR #155, opened 2018-09-05
Docs with Gatsby 2

i-like-robots:
There's certainly no point looking at the diff for this so here are the highlights and then check out the branch to try it out locally!

- Gatsby V1 => V2 🆕 
- Docs folder now in the top level of the repository  📖 
- Closely follows the [new Gatsby docs site][1] for easy reference 👯 
- Build time reduced by nearly 70% ⚡️  
- Trimmed ~300 dependencies 🗑 

Landing page:

x-docs landing

Module listing:

x-docs module list

Module page:

x-docs module page

Component demos:

x-docs demos

Thank you very much to Katya and Matt for fighting the hard fights and battling the really tricky stuff with Gatsby.

[1]: https://github.com/gatsbyjs/gatsby/tree/master/www


tools/x-docs/src/data/docs-menu.yml
```diff
@@ -0,0 +1,22 @@
...
+    - title: Create a component
+      link: /docs/creating-a-component
+- title: Applications
+  items:
+    - title: Quick start
+      link: /docs/apps
+    - title: Linking components
+      link: /docs/apps/linking
+    - title: Setup
+      link: /docs/apps/setup
```
i-like-robots: This menus via config pattern is copied over from the Gatsby docs. Auto building the menu might be preferable as per the V1 implementation but that requires more code so 🤷‍♂️

PR #157, opened 2018-09-07
Why we don't use Yarn

i-like-robots:
Opening this PR up for reference and to answer the question; "why don't you use Yarn?"

Firstly, Yarn is a pretty good tool these days! It's a single dependency and can be very fast indeed. Yarn supports workspaces so it works with multi-package repos out of the box.

However, some of its design choices are incompatible with npm and module resolution can be challenging.

The good stuff 👍 :

- Initial install and lock file creation is ~50% faster than npm
- With a Yarn lock file fresh installs without any cached modules take around half that time again
- Running an install with cached modules can take only a few seconds
- Workspaces hoist and dedupe dependencies at the top level which saves a lot of disk space

The bad stuff 👎 :

- Using the Node module resolution algorithm against a flattened dependency tree makes resolving the correct version of a module impossible in several situations. For instance Gatsby and Storybook require different versions of Webpack (1.x vs 3.x) and React (15.x vs 16.x) but only the lowest denominator will be hoisted. This means requiring the `webpack` dependency in the workbench configuration code will incorrectly resolve to 1.x and requiring `react` in the docs site source code will incorrectly resolve to 16.x.
- Interlinked dependencies use a protocol incompatible with npm  (`link:` rather than `file:`) which means apps cannot directly consume components being developed locally.

If you do clone this branch you'll find that it does install and the components and packages will build. I have been able to modify the Storybook configuration to ensure it loads the correct dependencies but Gatsby will not build for a quite a long list of reasons! As demonstrated some of these problems _are solvable_ but others are not without bigger changes.

For example to fetch the version of Webpack which Storybook needs you could do something like this:

```diff
- const webpack = require('webpack');
+ const webpackResolution = require.resolve('webpack', {
+	paths: [require.resolve('@storybook/react')]
+ });
+ const webpack = require(webpackResolution);
```

But not all of these resolutions are patchable in this manner.


zhenyulin:
resolving versions of common child dependencies is a real pain in yarn, in npm install, it looks like ti would update the package.json file adding fields like `requiredBy` to link a specific version of dep to a parent dep, while yarn isn't doing anything in this space and yarn.lock isn't helping the module import; wonder if this actually makes npm slower than yarn?

PR #162, opened 2018-09-07
X eventpromo styles

ludrob:
Ported n-eventpromo styles to x-eventpromo
- cleaned up useless classes
- use state to control animations


components/x-eventpromo/src/Footer.scss
```diff
@@ -0,0 +1,83 @@
...
+
+.brand {
+  display: flex;
+  align-items: center;
+  @include oTypographySans($scale: -2);
+  span {
+	padding-right: $ep-half-space;
+  }
+  &::after {
+	display: inarticle-block;
```
i-like-robots: typo

components/x-eventpromo/src/Image.jsx
```diff
@@ -0,0 +1,29 @@
...
+	return (
+		<a href={linkUrl}
+						data-trackable="event-promo"
+						aria-hidden="true"
+						tabIndex="-1"
+		>
+			<img
+				className={classNames}
+				srcSet={imgSrc} alt=""
+				data-n-image-lazy-load-js=""
```
i-like-robots: The `data-n-image-lazy-load-js` attribute should be added by the lazy loading script so not needed here

components/x-eventpromo/src/Image.scss
```diff
@@ -0,0 +1,58 @@
...
+}
+.fade-2 {
+	opacity: 0;
+	z-index: 30;
+	animation: fade 20s 6s infinite;
+}
+.paused {
+	animation-play-state: paused;
+}
+@keyframes fade {
```
i-like-robots: Worth checking if this animation name gets processed or not. If not this needs a better namespaced name.

components/x-eventpromo/src/ImagesContainer.jsx
```diff
@@ -0,0 +1,39 @@
+import { h, Component } from '@financial-times/x-engine';
+import Image from './Image';
+import styles from './ImagesContainer.scss';
+
+class ImagesContainer extends Component {
+
+	constructor(props, ...args) {
+		super(props, ...args);
```
i-like-robots: What is the purpose of the rest arg here? Can there be anything other than `props`?

components/x-eventpromo/src/styles/bootstrap.scss
```diff
@@ -0,0 +1,10 @@
+@import "n-ui-foundations/main";
+@import 'o-buttons/main';
+
+$o-fonts-is-silent: false;
+@import 'o-fonts/main';
+
+$ep-standard-space: 20px;
+$ep-half-space: $ep-standard-space/2;
+
+@include oFontsInclude(MetricWeb, $weight: light);
```
i-like-robots: What is the intention of this bootstrap? If apps were to include this component should they include this? They will already have the fonts setup etc.

components/x-eventpromo/src/lib/animation-control.js
```diff
@@ -0,0 +1,18 @@
...
+	aniCntr.addEventListener('click', () => {
+		aniEls.forEach((item) => {
+			item.classList.toggle('paused');
+		});
+
+		aniCntr.classList.toggle('event-promo__control--paused');
+	});
+
+	return true;
+};
```
i-like-robots: What's this file for?

components/x-eventpromo/src/styles/bootstrap.scss
ludrob: Indeed, this may only be needed for storybook and demos so definitely something to review. 
So far I was literally trying to reproduce what we have in n-eventpromo like for like in x-dash, without too much consideration for the final integration.

components/x-eventpromo/src/lib/animation-control.js
ludrob: a mistake :( 
That a copy of /lib file from n-eventpromo, not used here... I'll delete them.

components/x-eventpromo/src/Image.scss
quarterto: can't find any documentation on this but inspecting the output, looks like the animation names are scoped

components/x-eventpromo/src/ImagesContainer.jsx
quarterto: react also passes `context` through here

PR #164, opened 2018-09-13
x-article-save-button

dan-searle:
screen shot 2018-09-13 at 11 59 19

Journey Optimisation are building this component as a first step towards x-dashing the myFT timeline feed (for the App and ft.com).

PR raised very early. Feedback always welcome!

### DIFF
--- a/components/x-article-save-button/package.json
+++ b/components/x-article-save-button/package.json
@@ -7,10 +7,9 @@
   "browser": "dist/ArticleSaveButton.es5.js",
   "style": "dist/ArticleSaveButton.css",
   "scripts": {
-    "prepare": "npm run build",
+    "prepare": "bower install && npm run build",
     "build": "node rollup.js",
-    "start": "node rollup.js --watch",
-    "postinstall": "bower install"
+    "start": "node rollup.js --watch"
   },
   "keywords": [],
   "author": "",
--- a/components/x-article-save-button/package.json
+++ b/components/x-article-save-button/package.json
@@ -10,14 +10,16 @@
   "scripts": {
     "prepare": "npm run build",
     "build": "node rollup.js",
-    "start": "node rollup.js --watch"
+    "start": "node rollup.js --watch",
+    "postinstall": "bower install"
   },
   "keywords": [],
   "author": "",
   "license": "ISC",
   "devDependencies": {
     "@financial-times/x-rollup": "file:../../packages/x-rollup",
-    "node-sass": "^4.9.2"
+    "node-sass": "^4.9.2",
+    "bower": "^1.7.9"
   },
   "dependencies": {
     "@financial-times/x-engine": "file:../../packages/x-engine",
--- a/components/x-article-save-button/package.json
+++ b/components/x-article-save-button/package.json
@@ -16,7 +16,8 @@
   "author": "",
   "license": "ISC",
   "devDependencies": {
-    "@financial-times/x-rollup": "file:../../packages/x-rollup"
+    "@financial-times/x-rollup": "file:../../packages/x-rollup",
+    "node-sass": "^4.9.2"
   },
   "dependencies": {
     "@financial-times/x-engine": "file:../../packages/x-engine",
--- a/components/x-article-save-button/package.json
+++ b/components/x-article-save-button/package.json
@@ -1,7 +1,6 @@
 {
   "name": "@financial-times/x-article-save-button",
   "version": "0.0.0",
-  "private": true,
   "description": "",
   "main": "dist/ArticleSaveButton.cjs.js",
   "module": "dist/ArticleSaveButton.esm.js",
--- a/components/x-article-save-button/package.json
+++ b/components/x-article-save-button/package.json
@@ -1,0 +1,26 @@
+{
+  "name": "@financial-times/x-article-save-button",
+  "version": "0.0.0",
+  "private": true,
+  "description": "",
+  "main": "dist/ArticleSaveButton.cjs.js",
+  "module": "dist/ArticleSaveButton.esm.js",
+  "browser": "dist/ArticleSaveButton.es5.js",
+  "style": "dist/ArticleSaveButton.css",
+  "scripts": {
+    "prepare": "npm run build",
+    "build": "node rollup.js",
+    "start": "node rollup.js --watch"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "devDependencies": {
+    "@financial-times/x-rollup": "file:../../packages/x-rollup"
+  },
+  "dependencies": {
+    "@financial-times/x-engine": "file:../../packages/x-engine",
+    "@financial-times/x-interaction": "file:../x-interaction",
+    "classnames": "^2.2.6"
+  }
+}
--- a/components/x-article-save-button/package.json
+++ b/components/x-article-save-button/package.json
@@ -21,7 +21,6 @@
   },
   "dependencies": {
     "@financial-times/x-engine": "file:../../packages/x-engine",
-    "@financial-times/x-interaction": "file:../x-interaction",
     "classnames": "^2.2.6"
   }
 }
--- a/components/x-article-save-button/readme.md
+++ b/components/x-article-save-button/readme.md
@@ -1,6 +1,6 @@
 # Article Save Button
 
-This module provides a template for myFT save article buttons, and is intended to replace and is intended to replace the legacy handlebars component in [n-myft-ui](https://github.com/Financial-Times/n-myft-ui/tree/master/components/save-for-later).
+This module provides a template for myFT save article buttons, and is intended to replace the legacy handlebars component in [n-myft-ui](https://github.com/Financial-Times/n-myft-ui/tree/master/components/save-for-later).
 
 ## Installation
 
--- a/components/x-article-save-button/readme.md
+++ b/components/x-article-save-button/readme.md
@@ -37,5 +37,13 @@ Property           | Value
 
 It is up to the consumer of this component to listen for the `x-article-save-button` event, and use this data, along with the user's ID, and carry out the appropriate action.
 
-For example, if using `next-myft-client` to carry out the save/unsave action, n-myft-ui provides a x-button-interaction component for this:
-https://github.com/Financial-Times/n-myft-ui/blob/master/components/x-button-integration/index.js
+For example, if using `next-myft-client`:
+```
+import nextMyftClient from 'next-myft-client';
+
+document.body.addEventListener('x-article-save-button', ({ detail }) => {
+	nextMyftClient[detail.action](detail.actorType, null, detail.relationshipName, detail.subjectType, detail.subjectId, {
+		token: detail.token
+	});
+});
+```
--- a/components/x-article-save-button/readme.md
+++ b/components/x-article-save-button/readme.md
@@ -1,1 +1,41 @@
 # Article Save Button
+
+This module provides a template for myFT save article buttons, and is intended to replace and is intended to replace the legacy handlebars component in [n-myft-ui](https://github.com/Financial-Times/n-myft-ui/tree/master/components/save-for-later).
+
+## Installation
+
+```bash
+npm install --save @financial-times/x-article-save-button
+```
+
+## Props
+
+(Some of the properties don't influence the way button looks or acts, but can be used for e.g. client-side Javascript in the apps).
+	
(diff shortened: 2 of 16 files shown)


tools/x-workbench/register-components.js
```diff
@@ -1,4 +1,5 @@
 const components = [
+	require('../../components/x-article-save-button/stories'),
```
quarterto: this should be `@financial-times/x-article-save-button` as with the rest of the components. adding it as a `file:` reference in `x-workbench`'s `package.json` will work even before the component is released

components/x-article-save-button/src/ArticleSaveButton.scss
```diff
@@ -0,0 +1,36 @@
...
+
+.article-save-button__button {
+	@include background-icon(bookmark-outline, #000000);
+	@include oTypographySans($scale: -2);
+	border: 0;
+	padding: ($icon-size - 5px) 0 0 0;
+	width: $icon-size;
+	color: getColor('black');
+
+	&:focus {
```
i-like-robots: Do we need `:active` as well?

components/x-article-save-button/src/ArticleSaveButton.scss
```diff
@@ -0,0 +1,36 @@
...
+		outline: 3px solid getColor('teal-40');
+		outline-offset: 3px;
+	}
+
+	&:hover,
+	&:focus {
+		@include background-icon(bookmark, oColorsGetPaletteColor('black-50'));
+	}
+
+	.article-save-button--saved &:not(:hover):not(:focus) {
```
i-like-robots: If this class name is moved onto the button element instead of the container then the selector wouldn't need to negate all of the pseudo-classes which would be neater IMO

components/x-article-save-button/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "build": "node rollup.js",
+    "start": "node rollup.js --watch",
+    "postinstall": "bower install"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "devDependencies": {
+    "@financial-times/x-rollup": "file:../../packages/x-rollup",
+    "node-sass": "^4.9.2",
```
i-like-robots: @quarterto should this be a dependency of x-rollup?

components/x-article-save-button/src/ArticleSaveButton.scss
```diff
@@ -0,0 +1,36 @@
...
+
+@mixin background-icon($name, $color) {
+	@include oIconsGetIcon($name, oColorsGetPaletteColor($color), $icon-size, $iconset-version: 1);
+}
+
+.article-save-button {
+	display: inline-block;
+}
+
+.article-save-button__button {
```
quarterto: this isn't documented but because we're using CSS modules, BEM-style class names aren't necessary. i'd recommend using short, valid-JS-identifier classes for readability, e.g. `.articleSave`, `.button`, `.saved`, then you can reference them as `styles.articleSave` etc

components/x-article-save-button/src/ArticleSaveButton.scss
dan-searle: There doesn't appear to a be an active state for the existing save button.

components/x-article-save-button/src/ArticleSaveButton.scss
dan-searle: Yep, sure. I was just allowing for other changes in styling for the saved state - stuff outside of the `button`. There's no immediate use case for that though. Actually I think a separate `--saved` class is not necessary, and the `aria-pressed` attribute could be used instead.

components/x-article-save-button/stories/index.js
```diff
@@ -0,0 +1,11 @@
+const { ArticleSaveButton } = require('../');
+
+exports.component = ArticleSaveButton;
+exports.package = require('../package.json');
+exports.knobs = require('./knobs');
+exports.dependencies = {
+	'o-fonts': '^3.0.0'
```
i-like-robots: If you switch this to [this](https://github.com/Financial-Times/x-dash/blob/master/private/blueprints/component/stories/index.js#L7-L11) then the document should be setup as expected 👌

i-like-robots:
Is there anything left to do on this component or are you waiting for anything from Matt or I?

dan-searle:
I'm not sure how much more there is to do yet. I'm currently trying to get it working in `next-myft-page`. It may be there isn't much more to do. I would like to add some tests though.

dan-searle:
@i-like-robots, @quarterto - update... I'll soon by blocked by needing #170 to be merged.

components/x-article-save-button/stories/index.js
```diff
@@ -0,0 +1,12 @@
+const { ArticleSaveButton } = require('../');
+
+exports.component = ArticleSaveButton;
+exports.package = require('../package.json');
+exports.knobs = require('./knobs');
+exports.dependencies = {
+	'o-normalise': '^1.6.0',
+	'o-typography': '^5.5.0'
```
benbarnett: Do we need to include `o-icons` and the ones listed here? https://github.com/Financial-Times/x-dash/pull/164/files#diff-ef020e752ebe496f72a896942f3824ccR1

components/x-article-save-button/src/ArticleSaveButton.scss
```diff
@@ -0,0 +1,35 @@
...
+@mixin background-icon($name, $color) {
+	@include oIconsGetIcon($name, oColorsGetPaletteColor($color), $icon-size, $iconset-version: 1);
+}
+
+.root {
+	display: inline-block;
+}
+
+.button {
+	@include background-icon(bookmark-outline, #000000);
```
benbarnett: Oh, can you use `oColorsGetPaletteColor(#000000)`? I would expect to see `black` being passed in here.

dan-searle:
@quarterto @i-like-robots are you happy for me to release this as experimental?

quarterto:
prereleases are at your own discretion, go for it!

components/x-article-save-button/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@financial-times/x-article-save-button",
   "version": "0.0.0",
-  "private": true,
+  "private": false,
```
quarterto: `false` is the default, so best to leave this out

benbarnett:
@i-like-robots this one also ready

components/x-article-save-button/readme.md
```diff
@@ -0,0 +1,41 @@
...
+## Props
+
+(Some of the properties don't influence the way button looks or acts, but can be used for e.g. client-side Javascript in the apps).
+	
+Feature                     | Type    | Required | Default value      | Description
+----------------------------|---------|----------|--------------------|----------------------
+`contentId`                 | String  | yes      | none               | UUID of the article
+`contentTitle`              | String  | yes      | none               | Title of the article
+`saved`                     | Boolean | no       | `false`            | Whether the article is saved or not.
+`csrfToken`                 | String  | no       | none               | CSRF token (will be included in a hidden form field).
```
i-like-robots: How can I get one of these?

components/x-article-save-button/readme.md
```diff
@@ -0,0 +1,41 @@
...
+`actorType`        | `user`
+`relationshipName` | `saved`
+`subjectType`      | `content`
+`subjectId`        | the value of the `contentId` prop
+`token`            | the value of the `csrfToken` prop
+
+It is up to the consumer of this component to listen for the `x-article-save-button` event, and use this data, along with the user's ID, and carry out the appropriate action.
+
+For example, if using `next-myft-client` to carry out the save/unsave action, n-myft-ui provides a x-button-interaction component for this:
+https://github.com/Financial-Times/n-myft-ui/blob/master/components/x-button-integration/index.js
```
i-like-robots: Could an example be inlined into this readme as this link is liable to become out of date?

components/x-article-save-button/src/ArticleSaveButton.jsx
```diff
@@ -0,0 +1,53 @@
...
+			className={classNames(articleSaveStyles.root)}
+			action={`/myft/save/${props.contentId}`}
+			method="GET"
+			data-content-id={props.contentId}
+			onSubmit={event => {
+				event.preventDefault();
+				const detail = {
+					action: props.saved ? 'remove' : 'add',
+					actorType: 'user',
+					actorId: null, // myft client sets to user id from session
```
i-like-robots: Does this need specifying at all? From my point of view this doesn't mean anything and is a bit confusing.

components/x-article-save-button/src/ArticleSaveButton.scss
```diff
@@ -0,0 +1,41 @@
...
+
+	&:focus {
+		outline: none;
+	}
+
+	.icon {
+		@include background-icon(bookmark-outline, 'black');
+	}
+
+	body:not(:global(.touch-device)) &:not(:focus):hover .icon {
```
i-like-robots: Worth a comment mentioning what this is?

components/x-article-save-button/stories/basic.js
```diff
@@ -0,0 +1,17 @@
+exports.title = 'Basic ArticleSaveButton';
```
i-like-robots: Default?

components/x-article-save-button/src/ArticleSaveButton.jsx
dan-searle: No it doesn't. As the code that actually performs the save action lives outside of this component, there's no need to even mention this here. I'll remove it.

components/x-article-save-button/readme.md
```diff
@@ -0,0 +1,49 @@
+# Article Save Button
+
+This module provides a template for myFT save article buttons, and is intended to replace and is intended to replace the legacy handlebars component in [n-myft-ui](https://github.com/Financial-Times/n-myft-ui/tree/master/components/save-for-later).
```
i-like-robots: " and is intended to replace and is intended to replace "

components/x-article-save-button/readme.md
dan-searle: I took the view that this component should be agnostic about that. Where the CSRF token comes from (if one is needed at all) would vary depending on the application that consumes the button, and could change in the future.

components/x-article-save-button/readme.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Installation
+
+```bash
+npm install --save @financial-times/x-article-save-button
+```
+
+## Props
```
i-like-robots: I think this needs a follow up "usage" section, might even be able to copy and paste from here!

https://github.com/Financial-Times/x-dash/tree/master/components/x-teaser#usage

dan-searle: I thought this was sufficiently covered in a generic way [here](https://financial-times.github.io/x-dash/docs/integration/using-components#rendering), but I'm happy to add a specific section in here too.

components/x-article-save-button/src/ArticleSaveButton.scss
quarterto: Is `.touch-device` an `n-ui` thing? I'd either document that the component expects this class on touchscreens, or consider `@media (hover: hover)` which is [well-supported](https://caniuse.com/#feat=css-media-interaction) now

components/x-article-save-button/src/ArticleSaveButton.jsx
```diff
@@ -0,0 +1,52 @@
...
+				const detail = {
+					action: props.saved ? 'remove' : 'add',
+					actorType: 'user',
+					relationshipName: 'saved',
+					subjectType: 'content',
+					subjectId: props.contentId,
+					token: props.csrfToken
+				};
+
+				event.target.dispatchEvent(new CustomEvent('x-article-save-button', { bubbles: true, detail }));
```
quarterto: It would be simpler for apps consuming this component within React if, instead of dispatching an event, this called a prop on `ArticleSaveButton` (e.g. `onSaveClick` or similar). If the prop isn't there, it could fall back to dispatching the event.

```suggestion
				if(typeof props.onSaveClick === 'function') {
					props.onSaveClick(detail);
				} else {
					event.target.dispatchEvent(new CustomEvent('x-article-save-button', { bubbles: true, detail }));
				}
```

This way, React consumers won't have to worry about lifecycles and attaching event handlers.

edds:

apaleslimghost (approved):

Issue #163, opened 2018-09-12
external actions in x-interaction


quarterto:
## the problem

it would be useful for certain components to be able to expose actions that can be called from other parts of an app consuming the component. some use cases for this:

- triggering loading data on page load, or component visibility
- showing a component when an event is triggered on a non-x-dash component, e.g. a dialog or menu

there's currently no way to access the actions from an x-interaction wrapped component outside of the component, and this is the only way to alter a component once rendered.

## concerns

there's a few things we need to be careful of, which are mostly things the App is concerned with, as a long-lived web page:

- event listeners must be cleaned up properly when a component is unmounted
- any external actions must be triggered on the right instance, in case multiple instances of one component are rendered at the same time
- needs to work cross-runtime; there might be some nuances in how react & preact handle `ref`s that could cause problems, but idk
- needs to work with `x-interaction` server-side rendering, where the code rendering the component isn't under the app's control, so they don't have access to the instance

## proposals

for a component that uses `x-interaction`, such as `x-increment`:

```jsx
const withIncrementActions = withActions(() => ({
	increment(event) {
		return ({count}) => ({
			count: count + event.detail.amount || 1
		});
	},
}));

const BaseIncrement = ({count, actions: {increment}}) => <div>
	{count}
	<button onClick={increment}>Increment</button>
</div>;

const Increment = withIncrementActions(BaseIncrement);
```

here's a couple of ideas for APIs for triggering these actions externally (these ideas aren't alternatives, they're complementary):

### `actionRef`

allow an app that's rendering it directly to access actions using `actionRef`:

```jsx
import {Increment} from '@financial-times/x-increment';
import React, {Component} from 'react';

class Foo extends Component {
	doAThing() {
		if(this.incrementActions) {
			this.incrementActions.increment({detail: {amount: 5}});
		}
	}

	render() {
		return <Increment actionRef={actions => this.incrementActions = actions} />;
	}
}
```

| pros | cons |
|-|-|
| scoped to the rendering of the parent component | restricted to within other components |
| can pass in any args | possible cross-runtime incompatibility? |

### `CustomEvent`

allow an app rendering via x-interaction SSR to trigger a custom event on a DOM element which fires the action:

#### server-side

```handlebars
{{{x package="x-increment" component="Increment" id="x-increment-1"}}}
```

#### client-side

```js
import {hydrate} from '@financial-times/x-interaction';

document.addEventListener('DOMContentLoaded', hydrate);

function doAThing() {
[...]

PR #165, opened 2018-09-13
enzyme tests for x-increment & x-interaction; document test strategy

quarterto:
as a simple x-interaction-based component that actually uses all of its features, x-increment is a good testing surface area. start testing it with enzyme before rewriting it and doing #163 

todo:

- [x] more in-depth tests for x-interaction
- [x] document


babel-jest.js
```diff
@@ -0,0 +1,8 @@
+const getBabelConfig = require('./packages/x-rollup/src/babel-config');
+const babelJest = require('babel-jest');
+
+const config = getBabelConfig([], false);
+
+delete config.include;
+
+module.exports = babelJest.createTransformer(config);
```
quarterto: don't like doing this here, i'll move it somewhere better

packages/x-rollup/src/babel-config.js
```diff
@@ -18,7 +18,7 @@ module.exports = (targets = []) => ({
 			}
 		],
 		// Instruct Babel to not include any internal helper declarations in the output
-		require.resolve('babel-plugin-external-helpers'),
+		externalHelpers && require.resolve('babel-plugin-external-helpers'),
```
quarterto: ugh sorry

components/x-increment/__tests__/x-increment.test.js
```diff
@@ -0,0 +1,17 @@
...
+Enzyme.configure({ adapter: new Adapter() });
+
+const { mount } = Enzyme;
+
+describe('x-increment', () => {
+	it('should increment when clicked', () => {
+		const subject = mount(<Increment count={1} />);
+		subject.find('button').simulate('click');
+
+		expect(subject.find('span').text()).toEqual('2');
```
quarterto: y r u broken

components/x-interaction/src/InteractionRender.jsx
```diff
@@ -1,21 +1,13 @@
...
 import { loading } from './concerns/symbols';
-import { addSerialisationData } from './concerns/serialiser';
-import shortId from '@quarterto/short-id';
 
 export const InteractionRender = ({
+	id,
 	actions,
 	state,
 	initialState,
+	[loading]: isLoading,
```
i-like-robots: Does this need to be a symbol? Or to put it another way, do you expect consumers of this class to access this property?

components/x-interaction/src/InteractionRender.jsx
```diff
@@ -1,21 +1,13 @@
...
-}) => {
-	const { [loading]: isLoading, ...props } = state || initialState;
-	addSerialisationData({ id, Component, props });
-
-	const rendered = <Component {...props} actions={actions} isLoading={isLoading} />;
-
-	return hydrating ? rendered : <div data-x-dash-id={id}>{rendered}</div>;
-};
+}) => (
+	<Component {...initialState} {...state} {...{ id, actions, isLoading }} />
```
i-like-robots: I think this can be simply `{...state, id, actions, isLoading }`

packages/x-rollup/src/babel-config.js
```diff
@@ -1,43 +1,14 @@
...
-				targets,
-				modules: false,
-				exclude: ['transform-regenerator', 'transform-async-to-generator']
-			}
-		]
-	]
-});
+	);
+
+	base.include = '**/*.{js,jsx}';
```
i-like-robots: Worth a commend why this isn't pulled into the shared config?

tools/x-docs/src/docs/guides/components/cookbook/testing.md
```diff
@@ -0,0 +1,43 @@
+# Testing components
+
+Defining [stories](../setup/#writing-tests--demos) from your components, as well as including them in Storybook, will generate Jest snapshot tests for each story. Your stories should include as many of your component's use cases as possible, so that they're covered by the tests.
```
i-like-robots: "the tests" => "these tests"?

tools/x-docs/src/docs/guides/components/cookbook/testing.md
```diff
@@ -0,0 +1,43 @@
...
+
+Snapshot tests are useful for ensuring that you don't accidentally change the markup that your component outputs, which may inadvertently break the apps consuming your component. They don't cover any kind of interactive behaviour or changing state. For components using `x-interaction`, you should consider adding test cases for the interactivity.
+
+## Interactive components
+
+There's no requirement to use any particular testing framework. However, x-dash comes with Jest and Enzyme pre-configured, which should help save boilerplate and setup when testing your component.
```
i-like-robots: Maybe remove the first sentence and encourage the one way of doing things?

packages/x-rollup/src/babel-config.js
quarterto: it's because it's rollup-specific

components/x-interaction/src/InteractionRender.jsx
quarterto: a) there's no way to access this externally, it's just so anyone using react devtools or whatever can see this is the internal one, b) #170 overhauls this logic anyway?

components/x-interaction/src/InteractionRender.jsx
quarterto: `initialState` needs to be in there because it's not actually included in the `state`; doing `{...{...initialState, ...state, id, actions, isLoading}}` would mean extra object cloning (i think spread into component props doesn't clone, but spread into an object definitely does) plus it looks weird idk

tools/x-docs/src/docs/guides/components/cookbook/testing.md
quarterto: i was going for `the tests` in abstract, not any particular tests, but x-dash's entire test suite

components/x-interaction/src/InteractionRender.jsx
i-like-robots: Ahh, yes, JSX spread is not the spread you were looking for. Does this look better?

```jsx
<Component {...initialState} {...state} id={id} actions={actions} isLoading={isLoading} />
```

components/x-increment/__tests__/x-increment.test.jsx
```diff
@@ -0,0 +1,23 @@
+const { h } = require('@financial-times/x-engine');
+const { mount } = require('@financial-times/x-test-utils/enzyme');
+
+const { Increment } = require('../');
+
+describe('x-increment', () => {
+	it('should increment when clicked', async () => {
+		const subject = mount(<Increment count={1} />);
+		await subject.find('button').prop('onClick')();
```
dan-searle: Can we use Enzyme's `.simulate('click')` here?

quarterto: there would be no way using that to wait for the state change. calling the `onClick` directly we can get hold of the promise that the action returns

dan-searle: ah, of course 😊 

I think I need coffee.

components/x-interaction/src/InteractionRender.jsx
quarterto: it used to be that, i prefer the spread version?

components/x-increment/__tests__/x-increment.test.jsx
i-like-robots: The number of times I've written this rubbish https://github.com/Financial-Times/o-lazy-load/blob/d18cb98615a3915b08bc0abfde6e1c920f252512/test/helpers/wait-until.js 😆

components/x-interaction/src/Interaction.jsx
```diff
@@ -1,16 +1,15 @@
 import { h } from '@financial-times/x-engine';
 import { InteractionClass } from './InteractionClass';
-import { InteractionRender } from './InteractionRender';
+import { InteractionSSR } from './InteractionSSR';
 import wrapComponentName from './concerns/wrap-component-name';
 import { registerComponent } from './concerns/register-component';
 
 // use the class version for clientside and the static version for server
-const Interaction = typeof window !== 'undefined' ? InteractionClass : InteractionRender;
+const Interaction = typeof window !== 'undefined' ? InteractionClass : InteractionSSR;
```
i-like-robots: Good change 👍

packages/x-rollup/src/babel-config.js
```diff
@@ -1,43 +1,15 @@
...
-				modules: false,
-				exclude: ['transform-regenerator', 'transform-async-to-generator']
-			}
-		]
-	]
-});
+	);
+
+	// rollup-specific option not included in base config
+	base.include = '**/*.{js,jsx}';
```
i-like-robots: 👌

PR #168, opened 2018-09-14
Added new update-snapshots command to makefile

### LINKED ISSUE #167: Update snapshots incorrectly documented
The [docs](https://financial-times.github.io/x-dash/guides/components/setup#updating-snapshots) say to regenerate snapshots, run `npm test -- --updateSnapshot`, but that isn't working.

The following work:
* `npm test -- -- --updateSnapshot` (an additional `--`)
* `npm run jest -- --updateSnapshot`

Might be worth adding a new command in the `makefile`?

dan-searle:
Updated docs accordingly.

Fixes #167

### DIFF
--- a/makefile
+++ b/makefile
@@ -15,3 +15,6 @@ build:
 
 test:
 	npm run test
+
+update-snapshots:
+	npm run jest -- --updateSnapshot
--- a/tools/x-docs/src/docs/guides/components/setup.md
+++ b/tools/x-docs/src/docs/guides/components/setup.md
@@ -66,7 +66,7 @@ git commit -m "create snapshots for new story foo"
 If you make a change to a component that means its snapshots are out of date, you'll need to regenerate the snapshots. Run:
 
 ```bash
-npm test -- --updateSnapshot
+make update-snapshots
 ```
 
 and review the changes to the file `__tests__/__snapshots__/snapshots.test.js.snap`. Make sure the only changes are those you expected! When you're satisfied the snapshots are up to date, commit the file.


i-like-robots:
What about going with a "proper" make target e.g. `make snapshots.test.js`?

dan-searle:
That _makes_ a lot sense to me (see what I did there?).

makefile
```diff
@@ -15,3 +15,6 @@ build:
 
 test:
 	npm run test
+
+snapshots.test.js:
```
quarterto: this should be the full path i.e. `__tests__/__snapshots__/snapshots.test.js.snap`, and the documentation should reflect this

quarterto: (or, add an `update-snapshots: __tests__/__snapshots__/snapshots.test.js.snap` alias task)

quarterto: hmm, second (third?) thoughts, doing the actual file path with no prerequisites means the snapshots will never actually update, make will see the file is there and tell you it's up to date. so we need to either:

- mark it as `.PHONY`
- add prerequisites (probably something like `story-files = $(wildcard components/{stories,src}/{**,}/*.{js,jsx}` and `__tests__/__snapshots__/snapshots.test.js.snap: $(story-files)`)
- just use a generic task name like before

dan-searle: OK, I'm guessing I'm not the only one who is not very familiar with makefiles.

My vote would be to go back to the generic `update-snapshots` task name, as that's consistent with how we're using `make` already in this repo and elsewhere, and it's easy to understand. I don't feel strongly though.

quarterto: i'd love to get `n-gage` to a point where it can _make_ (a-hyuck) this kind of thing easy, turn it into more of a framework for writing makefiles instead of the giant blob of whatever it is currently

PR #169, opened 2018-09-17
Storyboard knobs not working for x-interactive wrapped component

dan-searle:
Demonstrating a problem I've run into.

When a component with knobs is made interactive, editing the knobs in Storybook no longer causes the component to be updated 'live'.

I ran into this with `x-article-save-button`, but it's more simply demonstrated with `x-increment`


quarterto:
i've merged this branch into #165, thanks @dan-searle

PR #170, opened 2018-09-17
triggering x-interaction actions externally, fixes #163

### LINKED ISSUE #163: external actions in x-interaction

quarterto:
- implements `actionsRef` and `CustomEvent` pretty much as described in #163 
- reinstates `x-ssr-demo`. i think we should keep it around, it's useful as a testbed
-  fix a potential source of race conditions in x-interaction `isLoading`

TODO:

- [x] document
- [x] ~is `HydrationWrapper` testable?~ would need a full dom environment for that
- [x] rebase once #165 is merged
- [x] refactor the serialiser to be not terrifying


dan-searle:
Nice!

I think it would be good to consider (and include a working example of) how we would pass arguments to actions triggered externally. In this case, perhaps as well as the `increment` action, have an action for setting the count.

I notice that currently, the whole event object is passed to the component's action handler. I think it may be better to limit it to just part of the event - either just `event.detail`, or perhaps just arguments that are explicitly intended for that action. Something like this:

```es6
increment.dispatchEvent(new CustomEvent('x-interaction.trigger-action', {
  detail: {
    action: 'set-count',
    arguments: { // only this object gets passed to the set-count action handler
      count: 10
    }
  },
}));
```

quarterto:
i went for the whole event as i'm guessing it would be common for internal actions to be used as an event handler (e.g. `onClick` in `x-increment`), and then if there's an event used both internally and externally the argument is an event each time.

dan-searle:
Ah, I see, yeah that does make sense, but if the actions are the component's external API, then that should be independent of any particular internal implementation. The internals should be able to be changed without changing the external API (and vice versa)?

components/x-interaction/readme.md
```diff
@@ -108,3 +108,114 @@ export const BaseGreeting = ({greeting, actions}) => <div>
...
+
+### Hydrating server-rendered markup
+
+When you have an `x-interaction` component rendered by the server, and you want to attach the client-side version of the component to handle the actions, rather than rendering the component manually (which might become unwieldy, especially if you have many components & instances on the page), you can have `x-interaction` manage it for you.
+
+There are two parts to this: serialising and hydrating.
+
+#### Serialising
+
+To ensure components are rendered with the same initial data on the client side, and keep track of all the instances that are rendered and their identifiers, you must create a serialiser on the server side. `x-interaction` exports a function `getInteractionSerialiser`. You *must* call this function *exactly once* at the start every HTTP request that responds with any number of `x-interaction` components.
```
i-like-robots: Not quite sure about the final sentence here, if this function returns a new instance then perhaps that could be mentioned?

> You must create a new instance of the serialiser for each request which responds with any `x-interaction` components.

components/x-interaction/readme.md
```diff
@@ -108,3 +108,114 @@ export const BaseGreeting = ({greeting, actions}) => <div>
...
+
+It returns a function that, when called, resets the serialisation state, and returns all serialisation data collected from any components that were rendered in the meantime, in the format of a `<script>` tag that should be included in the rendered page. This function *must* be called once at the end of every request when `getInteractionSerialiser`, _after_ all `x-interaction` components have been rendered. Any component rendered after this function is called won't have its data included in the output.
+
+If you call `getInteractionSerialiser` once for multiple requests, it will collect data for components that aren't rendered in the current request, causing inflated output size and losing that data for the pages that need it. Calling it multiple times per request is a no-op.
+
+If you don't call the function returned by `getInteractionSerialiser` at the end of the request, the data collected during this request will be lost, and possibly returned in a future request. When calling it multiple times, all calls after the first will return empty data, and including these in the response will potentially erase the global variable it outputs on the client side.
```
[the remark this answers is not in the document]
i-like-robots: Sounds like this data could be attached to the request object so that the server (of any flavour) can handle this lifecycle?

components/x-interaction/readme.md
```diff
@@ -108,3 +108,114 @@ export const BaseGreeting = ({greeting, actions}) => <div>
...
+
+#### Hydrating
+
+When rendered on the server side, components output an extra wrapper element, with a `data-x-dash-id` attribute. This is used by the `x-interaction` runtime to identify the component in serialisation data, and attach the correct instance to it on the client side. You can pass this in as the `id` property when rendering the component; otherwise, an identifier will be randomly generated.
+
+`x-interaction` exports a function `hydrate`. This should be called on the client side. It inspects the global serialisation data on the page, uses the identifiers to find the wrapper elements, and calls `render` from your chosen `x-engine` client-side runtime to render component instances into the wrappers.
+
+Before calling `hydrate`, you must first `import` any `x-interaction` components that will be rendered on the page. The components register themselves with the `x-interaction` runtime when imported; you don't need to do anything with the imported component. This will also ensure the component is included in your client-side bundle.
+
+Because `hydrate` expects the wrappers to be present in the DOM when called, it should be called in an event such as `DOMContentLoaded`.
```
i-like-robots: Maybe just say "The `hydrate` function should be called after the `DOMContentLoaded` event" and link off to MDN for info about that?

components/x-interaction/src/Hydrate.jsx
```diff
@@ -1,27 +1,80 @@
...
+
+export function hydrate() {
 	if (typeof window === 'undefined') {
 		throw new Error('x-interaction hydrate should only be called in the browser');
 	}
 
 	if (!('_xDashInteractionHydrationData' in window)) {
 		throw new Error(
-			"x-interaction hydrate was called without hydration data available. this could happen if you called hydrate before the hydration data was defined, or if you're not ouptutting the hydration data in your server-rendered markup."
+			"x-interaction hydrate was called without hydration data available. this could happen if you called hydrate above where the hydration data was defined on the page, or if you're not ouptutting the hydration data in your server-rendered markup."
```
i-like-robots: ouptutting => outputting


How about this?
> this can happen if you call hydrate before the serialised data is available or if you're not including the serialised data in your server-rendered markup.

components/x-interaction/src/Hydrate.jsx
```diff
@@ -1,27 +1,80 @@
...
+	document.querySelectorAll('[data-x-dash-id]').forEach(element => {
+		const {xDashId} = element.dataset;
+
+		const hasData = !window._xDashInteractionHydrationData.find(
+			({ id }) => id === xDashId
+		);
+
+		if(!hasData) {
+			throw new Error(
+				`found component markup for ${xDashId} without any hydration data. it's likely that its hydration data has been output in another request, or that the component was rendered after the serialisation data was output. ${serialiserOrdering}`
```
i-like-robots: Should "hydration data" be "serialised data" or "dehydrated data" ?

quarterto: here's my reasoning: it's the component that's dehydrated, not the data, and the data is used to hydrate, so it's the hydration data? any old data can be serialised but this is for a specific purpose

i-like-robots: Got it. Maybe near the start of the doc it could lead into it, e.g.

> The serialised data used for hydrating a component (_hydration data_)

components/x-interaction/readme.md
quarterto: it's not exactly a new instance because it's pretty much a global singleton, but i want to pretend it's a new instance. i don't particularly like the api we currently have for this but i can't think of a way of not having it as a global (we can't rely on `context` or similar)

components/x-interaction/readme.md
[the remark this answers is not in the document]
quarterto: i don't like the pattern of "just stick it on `req` or `res` because they're there", it's just a function that returns a string, there's nothing special about it, and i don't want to assume people are using express, or even using `http`, so i don't see any way of hooking it into the lifecycle?

components/x-interaction/__tests__/x-interaction.test.jsx
```diff
@@ -163,6 +201,25 @@ describe('x-interaction', () => {
...
+				foo() {},
+			})(Base);
+
+			const target = mount(<Wrapped actionsRef={actionsRef} />);
+
+			expect(actionsRef).toHaveBeenCalled();
+			expect(actionsRef.mock.calls[0][0]).toHaveProperty('foo');
+
+			// TODO: target.unmount() doesn't actually unmount the child for some reason?
+			target.find('InteractionClass').instance().componentWillUnount();
```
dan-searle: `componentWillUnount` -> `componentWillUnmount`

quarterto: welp, that would explain why it wasn't being called on `unmount`

quarterto:
quick update, i'm working on a refactor of the serialiser because the current one uses globals and that makes me uneasy. since it would be a breaking change to the behaviour i've documented in this PR i'll cram it into here

quarterto:
@dan-searle @benbarnett care to give this another quick 👀, not much has changed apart from the [entire serialiser api](https://github.com/Financial-Times/x-dash/pull/170/files#diff-03cdc88541e133c9e56b4b5ca8ec9217) 😬

Issue #172, opened 2018-09-18
Release docs site on merge to master


i-like-robots:
Currently the Github pages site is only deployed when tagging a release. This may not be helpful as new documentation or fixes to existing documentation will not be deployed until the next release.

i-like-robots:
Fixed in #175

PR #175, opened 2018-09-19
Deploy docs to github pages on merge to master

ddigulescu:
Addressing #172

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,6 +92,11 @@ references:
     branches:
       ignore: /.*/
 
+  filters_master_branch: &filters_master_branch
+    branches:
+      only:
+        - master
+
 jobs:
 
   build:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -226,6 +231,12 @@ workflows:
           requires:
             - test
 
+  deploy-docs:
+    jobs:
+      - deploy:
+          filters:
+            <<: *filters_master_branch
```
i-like-robots: I suspect this will deploy a blank directory as the workflow is incomplete. My understanding is that this won't inherit the existing `build-test` workflow used for all branch builds.

This deploy step could be added to the existing `build-test` workflow with the addition of the master branch filter.

ddigulescu: You are right, tested in a dummy repo 👍

i-like-robots:
Awesome, looks good - were you able to figure out a way to test it at all?

ddigulescu:
Only the circle-ci workflows bit not the actual push to `gh-pages` branch

PR #180, opened 2018-09-20
fix x-interaction state resetting

quarterto:


quarterto:
matt h seal of approval

PR #181, opened 2018-09-21
Write more documentation

i-like-robots:
This PR fleshes out and ties together some of the documentation we have so far accumulated. I have gone through all of the items included in the documentation navigation.

PR #184, opened 2018-09-26
Include documentation for mocking http requests

benbarnett:
Simple documentation to describe using fetch-mock for stories.

Following https://github.com/Financial-Times/x-dash/pull/181#discussion_r220573438

### DIFF
--- a/docs/components/stories.md
+++ b/docs/components/stories.md
@@ -47,11 +47,12 @@ exports.knobs = require('./knobs');
 
 Story modules export the configuration required for each component demo.
 
-Property | Description                                                 | Required
----------|-------------------------------------------------------------|----------
-`title`  | The title of the story                                      | Yes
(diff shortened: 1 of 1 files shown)


docs/components/stories.md
```diff
@@ -40,8 +41,15 @@ exports.stories = [
 
 // A module generating dynamically editable knobs
 exports.knobs = require('./knobs');
+
+// Stories will be provided a clean instance of fetch-mock to allow mocking of http requests
+exports.fetchMock = fetchMock => {
+    fetchMock.mock('/api/data', { some: 'data' });
+}
```
i-like-robots: Would it make sense to add something to the example in the story configuration section below?

benbarnett: Oops, it should be there instead of here!

PR #197, opened 2018-10-10
x-gift-article fetch enhancement

dan-searle:
* [x] Allow api protocol and domain to be set via props.
* [x] `fetch`es include credentials

Best to ignore whitespace on the diff: https://github.com/Financial-Times/x-dash/pull/197/files?w=1


fenglish:
There is props section in Readme. Could you add them there please?
https://github.com/Financial-Times/x-dash/blob/x-gift-article/components/x-gift-article/readme.md#properties

PR #202, opened 2018-10-12
x-gift-article for The App

benbarnett:
A WIP branch for any changes required to `x-gift-article` for the App integration.

We are using a separate branch here to keep `x-gift-article` clean as it is in production on next-article.


components/x-gift-article/src/lib/api.js
```diff
@@ -1,67 +1,76 @@
...
 		try {
-			const response = await fetch('/article-email/credits', { credentials: 'same-origin' });
-			const json = await response.json();
+			const json = await this.fetchJson('/article/gift-credits');
 
 			return {
-				monthlyAllowance: json.credits.allowance,
-				giftCredits: json.credits.remainingCredits,
-				nextRenewalDate: json.credits.renewalDate
+				monthlyAllowance: json.allowance,
```
benbarnett: Does still work? The server responses with this data in a `credits` object:

```json
{  
   "credits":{  
      "allowance":20,
      "consumedCredits":2,
      "remainingCredits":18,
      "renewalDate":"2018-12-01T00:00:00Z"
   }
}
```

quarterto:
@benbarnett what's the status of this and #78? this is currently in production in the App, and #78 in `next-article`. i'm a little unclear what's going on with the API `credits` thing

benbarnett:
@quarterto I'm not much clearer I'm afraid... you're right about the different versions in App and next-article. Ideally those would be brought inline and we can get this merged. That requires changes to how next-article renders x-gift-article I think.

The `credits` api can still call the next-article endpoints to work around CORS issues. Its a proxy to membership.

quarterto:
i meant your comment https://github.com/Financial-Times/x-dash/pull/202#discussion_r237556950 about it being wrapped in an extra object? this branch isn't expecting that format, and it's working in the App 🤷‍♀️

benbarnett:
Ohhh yeah. That still blows my mind a bit. Hit https://www.ft.com/article-email/credits in a logged-in browser and you'll see the expected the response. 

Am I missing something wildly obvious?

quarterto:
i've worked it out. it's calling `/article/gift-credits` not `/article-email/credits` now. the former calls the latter, and unwraps the `credits` object https://github.com/Financial-Times/next-article/blob/96dae6713b52d16e31f93db622a1ee258c8ee5dd/server/share/gift.js#L23. i have no idea why former endpoint exists, any thoughts?

quarterto:
as discussed, since #78 isn't being actively developed going to merge this PR into that and review it as a whole

Issue #177, opened 2018-09-19
Proposal for project structure changes


i-like-robots:
We have made a few decisions about the structure of the project that may be causing people working with x-dash a little more friction than we may have realised so this issue is to collect those small gripes and propose solutions to help smooth them out.

## Storybook

Storybook is intended (whether document or not) to run at the root of the component code that it is intended to load and display. This seems to mean the repository root (see examples [a], [b], [c], and [d]) but I don't think this is required.

[a]: https://github.com/IBM/carbon-components-react
[b]: https://github.com/lonelyplanet/backpack-ui
[c]: https://github.com/algolia/react-instantsearch/
[d]: https://github.com/reach/reach-ui

We decided to keep our storybook configuration and dependencies in a separate part of the repository `tools/x-storybook`. This works OK but we are having to work around a few limitations imposed by this decision and we cannot advise following the Storybook documentation to set up components.

To ensure Storybook works correctly we must declare our components as dependencies to link them in. This is because Storybook can only watch files which are descendants of the working directory (my understanding is that this is not a limitation of the tool itself but is a commonly applied rule).

I propose that it may be worth making storybook a top-level dependency of the project and moving the configuration to the top-level as well so that it matches other projects using Storybook and Storybook documentation itself.

## Docs

_This is a small suggestion should the proposed change to our Storybook setup happen..._

Moving Storybook would leave the `tools` directory with only one item, `x-docs`. For consistency with other projects I think it would be good to also to move the `x-docs` folder to the top level as a `www` or `website` directory (see examples [e], [f], and [g]).

[e]: https://github.com/gatsbyjs/gatsby
[f]: https://github.com/facebook/jest/
[g]: https://github.com/reach/reach-ui

## Move private packages

It is not obvious that our private packages `x-rollup`, `x-babel-config`, `x-logo`, and `x-test-utils` are private when looking at the directory structure. I think it could be made clear by keeping only public packages in the `packages` directory and moving the private packages into the (potentially vacated) `tools` directory.

Thoughts?

PR #203, opened 2018-10-16
move storybook to top level, find stories automatically

### LINKED ISSUE #173: Autoload all components with stories into Storybook
Currently adding a component to story book is a bit of a chore. Firstly they must be added as a dependency of `x-workbench` then the stories registered in a configuration file.

Storybook still works when switching from named package reference to a relative path reference, e.g.

```diff
const components = [
-  require('@financial-times/x-teaser/stories')
+  require('../../components/x-teaser/stories')
];
```

This would indicate that it should be possible to use a glob match to load all of the stories without any configuration 🤞

quarterto:
does part of #177, fixes #173 (which we closed as impossible turns out)

adding the stories automatically drastically slows down storybook's build however, because `require.context` adds all files in the components directory then filters out the ones that don't match the regex which seems inefficient. can we do something with a glob?

- [ ] get build working
- [ ] can we improve the speed of `require.context`?


docs/components/styling.md
```diff
@@ -74,10 +74,9 @@ const Button = ({ large, danger }) => {
 };
 ```
 
-_Please note: referencing classes and toggling them based on properties can become unwieldy so the [classnames] or [classcat] packages can help avoid some of the formatting hassle._
+_Please note: referencing classes and toggling them based on properties can become unwieldy so the [classcat] package can help avoid some of the formatting hassle._
```
i-like-robots: What did it do to you?

.storybook/build-story.js
```diff
@@ -65,9 +65,6 @@ function buildStory ({package: pkg, dependencies, component: Component, knobs, s
 		return (
 			<div className="story-container">
 				{dependencies && <BuildService dependencies={dependencies} />}
-				{pkg.style && <Helmet>
```
i-like-robots: Is the intention here that Storybook will transpile any Sass source files too?

.storybook/config.js
```diff
@@ -0,0 +1,12 @@
+import { configure } from '@storybook/react';
+import buildStory from './build-story';
+
+const requireStories = require.context('../', true, /components\/[^\/]+\/stories\/index.js$/);
```
i-like-robots: Would switching the subdirectories flag to `false` speed this up since this regexp is quite specific?

i-like-robots:
> can we improve the speed of `require.context`

You may be able use a glob to provide a list of files to this method using this doo daa: https://webpack.js.org/guides/dependency-management/#context-module-api

quarterto:
nah, because `require.context` is magic. it happens at compile time

.storybook/config.js
i-like-robots: Answer from IRL: "No because it won't ever traverse the directories"

i-like-robots:
To cover our chat: 

- We could use a macro and some Webpack tomfoolery to compile a list of matching files but that complexity is not worth the slight slowdown.
- The context can be slimmed down to only the components directory.

.storybook/build-story.js
quarterto: yeah, it kind of has to if we're importing component source files. i've set up `css-loader` and `style-loader` in storybook's webpack config and imo they work far better than including compiled styles (`x-styling-demo` loads styles instantly rather than in 2 seconds)

docs/components/styling.md
quarterto: oh haha i could not get it working with webpack ESM _at all_. looking at the source code it's a wonder how it ever worked in an ESM context and the maintainers seem quite hostile to writing it in a sensible way

quarterto:
i'm going to close this; i think it needs revisiting

PR #204, opened 2018-10-17
x-teaser-timeline

dan-searle:
x-dash version of myFT article timeline feed.

Checklist:

* [x] Take array of articles and split into groups by day, earlier, latest etc
* [x] Unit tests of article grouping
* [x] Render each article using x-teaser
* [x] styling of timeline in different viewports, heading styles etc
* [x] Documentation


components/x-timeline-feed/readme.md
```diff
@@ -0,0 +1,55 @@
...
+All `x-` components are designed to be compatible with a variety of runtimes, not just React. Check out the [`x-engine`][engine] documentation for a list of recommended libraries and frameworks.
+
+[jsx-wtf]: https://jasonformat.com/wtf-is-jsx/
+
+### Properties
+
+Feature              | Type   | Notes
+---------------------|--------|----------------------------
+`articles`           | Array  | (Mandatory) Array of objects, in Teaser format, representating articles to render. The articles should be in newest-first order.
+`timezoneOffset`     | Number | (Default = 0) Minutes to offset article publish times in order to display in user's timezone
```
benbarnett: As just discussed, perhaps this could default to the runtime clock as `localTodayDate` does. Makes it simpler for client-side usage.

components/x-timeline-feed/src/TimelineFeed.scss
```diff
@@ -0,0 +1,51 @@
...
+		grid-template: "heading articles" min-content / 1fr 3fr;
+	}
+}
+
+.articleGroup__heading {
+	@include oTypographySansBold($scale: 1);
+	@include oTypographyMargin($top: 2, $bottom: 2);
+
+	@include oGridRespondTo($from: M) {
+		grid-area: heading;
```
benbarnett: Minor thing, but I wonder if the grid template itself should change at the various viewports, rather than the elements within opting in & out of it?

dan-searle: I think it's better to not use CSS grid unless its necessary. I don't feel strongly about it though.

components/x-timeline-feed/src/TimelineFeed.jsx
```diff
@@ -0,0 +1,40 @@
...
+
+							return (
+								<li key={article.id} className={styles.article}>
+									<Teaser
+										{...article}
+										{...presets.SmallHeavy}
+									/>
+									{articleActions && <div
+										className={classNames(styles.articleActions)}
+										dangerouslySetInnerHTML={{__html: articleActions }}
```
quarterto: this should allow passing in a JSX child as well, like with `x-teaser`'s [`CustomSlot`](https://github.com/Financial-Times/x-dash/blob/b196dc20a09b861fce69e4e50fc77f3105f948b7/components/x-teaser/src/CustomSlot.jsx)

quarterto: actually, should this just pass `articleActions` into the teaser as the custom slot?

dan-searle: I've changed it to allow passing a JSX child or a function now.

I didn't use the teaser custom slot, because there didn't seem to be any advantage to doing so, and it made the CSS simpler to just treat the teaser and actions slot as separate grid areas.

components/x-timeline-feed/src/TimelineFeed.jsx
```diff
@@ -0,0 +1,40 @@
...
+			{articleGroups.map(group => (
+				<section key={group.date} className={classNames(styles.articleGroup)}>
+					<h2 className={classNames(styles.articleGroup__heading)}>{group.title}</h2>
+					<ul className={classNames(styles.articleGroup__articles)}>
+						{group.articles.map(article => {
+							const slotContent = typeof articleCustomSlot === 'function' ? articleCustomSlot(article): articleCustomSlot;
+
+							return (
+								<li key={article.id} className={styles.article}>
+									<Teaser
```
i-like-robots: As mentioned in #207 I think it may be better to not couple these components at all and provide the component to be rendered in this slot as a prop instead.

If it is better to keep the two components tightly coupled then perhaps this component should be called `x-teaser-list` instead to make that explicit.

components/x-timeline-feed/src/TimelineFeed.scss
```diff
@@ -0,0 +1,51 @@
...
+	display: flex;
+	justify-content: space-between;
+	border-bottom: 1px solid oColorsGetPaletteColor('black-20');
+	@include oTypographyMargin($bottom: 2);
+
+	:global {
+		.o-teaser {
+			border-bottom: 0;
+			@include oTypographyPadding($bottom: 0);
+		}
```
i-like-robots: If possible I think we should try to avoid components "reaching into" other components. This sort of thing lacks visibility and will never get cleaned up.

components/x-timeline-feed/src/lib/date.js
```diff
@@ -0,0 +1,51 @@
+import { format, isAfter, differenceInCalendarDays, subMinutes } from 'date-fns';
```
i-like-robots: In x-teaser we're using the `dateformat` module, would be good to see if we can standardise on one or the other.

components/x-timeline-feed/src/lib/date.js
```diff
@@ -0,0 +1,51 @@
...
+
+	articles.forEach(article => {
+		if (isAfter(article.localisedLastUpdated, splitDate)) {
+			latestArticles.push(article);
+		} else {
+			earlierArticles.push(article);
+		}
+	});
+
+	return { latestArticles, earlierArticles };
```
i-like-robots: Minor thing - these aren't necessarily articles but could be any type of content

components/x-timeline-feed/src/lib/transform.js
```diff
@@ -0,0 +1,83 @@
...
+	});
+
+	return Object.entries(articlesByLocalisedDate).map(([localDate, articles]) => ({
+		date: localDate,
+		articles
+	}));
+};
+
+export const splitTodaysArticles = (articleGroups, localTodayDate, latestArticlesTime) => {
+	const firstGroupIsToday = articleGroups[0].date === localTodayDate;
```
i-like-robots: Do you need to check if `articleGroups[0] != null`?

components/x-timeline-feed/src/lib/date.js
dan-searle: Should we adopt `content` rather than `articles`, everywhere?

components/x-timeline-feed/src/lib/date.js
dan-searle: Makes sense. I'd favour standardising on [date-fns](https://date-fns.org/) as it provides many date functions, including formatting.

components/x-timeline-feed/src/lib/date.js
i-like-robots: Up to you! They're usually referred to as "content items" so either `content` or `items` works for me personally.

components/x-timeline-feed/src/TimelineFeed.scss
dan-searle: I agree, though I couldn't see a way around it in this case.

components/x-timeline-feed/src/lib/date.js
dan-searle: OK cool, I'll go with `items` & `item`, as `content` is a bit more ambiguous regarding whether it's single or multiple etc.

components/x-timeline-feed/src/TimelineFeed.scss
quarterto: have this override as a separate class that your component passes to `x-teaser`?

dan-searle: Ah, I didn't realise x-teaser supported this. I'll do that. Thanks

components/x-timeline-feed/src/TimelineFeed.jsx
dan-searle: Talking to @benbarnett about this, the myFT feed timeline view is the only use case we have for this - we're x-dashing it so we can use it both the web and the app.

We also have a need for a simple list of teasers for use on the myFT Save Articles page - that won't have the time/date localisation and grouping that makes up the bulk of this component.

So our conclusion was it would be better to rename this component to `x-teaser-timeline`, and have a separate component (`x-teaser-list` probably) that may well have more use cases in future.

dan-searle: PR to rename this to **x-teaser-timeline**: https://github.com/Financial-Times/x-dash/pull/219

dan-searle:
Closing as this now exists on a new branch: https://github.com/Financial-Times/x-dash/pull/223

PR #209, opened 2018-10-24
Add audio to content type

debugwand:
Add 'audio' to ContentType

https://github.com/Financial-Times/next-article/blob/38c5c2c54f313ea128ff0aadf4eb76c85ed610ac/server/model/audio.js#L30


robsquires:
👍

PR #211, opened 2018-10-24
Add isPodcast indicator

debugwand:
ES Interface isPodcast indicator:
https://github.com/Financial-Times/next-es-interface/blob/8982c96c901dd9e95df5846e27297defc7211395/server/transformers/plugins/audio/teaser.js#L15

### DIFF
--- a/components/x-teaser/Props.d.ts
+++ b/components/x-teaser/Props.d.ts
@@ -133,6 +133,7 @@ export interface Indicators {
 	accessLevel: 'premium' | 'subscribed' | 'registered' | 'free';
 	isOpinion?: boolean;
 	isColumn?: boolean;
+	isPodcast?: boolean;
 	/** Methode packaging options */
 	isEditorsChoice?: boolean;
 	isExclusive?: boolean;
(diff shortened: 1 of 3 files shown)


components/x-teaser/Props.d.ts
```diff
@@ -133,6 +133,7 @@ export interface Indicators {
 	accessLevel: 'premium' | 'subscribed' | 'registered' | 'free';
 	isOpinion?: boolean;
 	isColumn?: boolean;
+	isPodcast?: boolean;
```
i-like-robots: Does this mean the top level `type` property will be `audio` in future rather than `podcast`?

debugwand: yes. type audio, subtype podcast, isPodcast in the teaser indicators.

i-like-robots: Good to know (though I'm sure you've told me this already!)

robsquires:
LGTM 👍  Am I right in thinking we'll be able to add a story for x-teaser once we finalise the design?

PR #213, opened 2018-10-25
Update changelog.md

debugwand:
update the changelog for beta 4 and 5


i-like-robots:
Could you add some notes to the PR description about the Acorn thing so we have some history attached to it?

PR #214, opened 2018-11-02
new x-interaction features

quarterto:
- single-nesting for rewrapped `withActions`
- default state as second argument to `withActions`
- updaters returned from actions can be `async`


components/x-interaction/src/InteractionClass.jsx
```diff
@@ -11,32 +11,32 @@ export class InteractionClass extends Component {
...
+			const nextState = typeof stateUpdate === 'function'
+				? Object.assign(
+					this.state.state,
+					await Promise.resolve(stateUpdate(Object.assign(
+						{},
+						props.initialState,
+						this.state.state
+					)))
+				)
+				: Object.assign(this.state.state, stateUpdate);
```
i-like-robots: Do you reckon it is worth micro-optimising these `Object.assign()` calls by using a faster merge function?

quarterto: do you have a preferred merge? i'm not terribly concerned about performance here as a) they're async, so they won't be blocking UI, and b) if somebody's triggering actions in a tight loop how about instead of doing that they don't

PR #217, opened 2018-11-12
x-node-jsx packages

i-like-robots:
This adds a new package which provides a `require()` hook to support the usage of `.jsx` files at runtime. It uses [Sucrase] to parse and transform the source code to plain ES2017 and [Pirates] to safely install (and remove) the hook. Adding this feature to Node's built-in functionality means we can take advantage of the default module resolution and caching mechanisms.

This works by adding a new file extension to `require.extensions`, copying the existing JS handler, then overwriting the compile method. Simple!

The reason for this module is so that `.jsx` files may be used transparently in an app without a build step. This is similar to many template libraries - including Handlebars - which compile templates on-the-fly on app startup. This functionality is convenient and avoids applications re-implementing build step tooling.

This is not intended to provide a wide range of functionality (as per tools like `@babel/register`), it is only intended to target JSX and ES6 import syntax. Transformation with Sucrase is super fast and keeps the dependency tree small (http://npm.broofa.com/?q=sucrase).

Usage:

```js
// Install the hook
require('@financial-times/x-node-jsx/register');

// Transparently require .jsx files
const App = require('./components/app.jsx');
```

_Note:_ I attempted to write tests for this package but Jest provides it's own `require` function (https://github.com/facebook/jest/issues/6034) making this impossible.

[Sucrase]: https://github.com/alangpierce/sucrase
[Pirates]: https://github.com/ariporad/pirates


maxcbc:
Out of curiosity do we have a support matrix for Node.js version, 6+, 8+, LTS?

packages/x-node-jsx/readme.md
```diff
@@ -0,0 +1,58 @@
...
+const installHook = require('@financial-times/x-node-jsx');
+
+const removeHook = installHook();
+
+removeHook();
+```
+
+## Usage
+
+This package provides a single helper function to. An options object may be provided to the function when called. The options and their defaults are shown below:
```
dan-searle: "...provides a single helper function to" - to what? :neutral_face:

benbarnett: "This package _even_ provides a single helper function, too!"

i-like-robots:
> Out of curiosity do we have a support matrix for Node.js version, 6+, 8+, LTS?

We do not but I've been working on the basis of latest LTS releases. So that was 6 and 8 when we started and it's 8 and 10 now.

quarterto:
it would be useful to include a `x-node-jsx/register` module to automatically register, so the hook can be included with `node -r @financial-times/x-node-jsx/register`

i-like-robots:
@quarterto updated to do just this 👌

PR #218, opened 2018-11-12
X topic search

fenglish:
This is for x-dashing topic search bar on myFT page.

screen shot 2018-11-12 at 14 41 36


i-like-robots:
More components \0/

If you need a steer with this particular component I can help, I've built typeahead components a few times now!

components/x-topic-search/readme.md
```diff
@@ -0,0 +1,44 @@
...
+[jsx-wtf]: https://jasonformat.com/wtf-is-jsx/
+
+### Properties
+
+Property                | Type   | Required | Note
+------------------------|--------|----------|------------------
+`minSearchLength`       | Number | No       | Minimum chars to start search. Default is 2
+`maxSuggestions`        | Number | No       | Maximum number to display suggestions. Default is 5
+`apiUrl`                | String | Yes      | The url to use when making requests to get topics
+`currentlyFollowingTopics` | Array | Yes    | Each item should have `name` and `uuid` properties
```
dan-searle: I'd go for just `followedTopics`

components/x-topic-search/src/ResultContainer.jsx
```diff
@@ -0,0 +1,46 @@
...
+export default ({ result, searchTerm }) => (
+	<div className={ resultContainerClassNames } data-component="topic-search">
+		<ul className={ classNames(styles["suggestions"]) } aria-live="polite">
+
+			{ result.status === 'suggestions' && <SuggestionList suggestions={ result.suggestions } searchTerm={ searchTerm }/> }
+
+			{ result.status === 'no-suggestions' && <NoSuggestions searchTerm={ searchTerm }/> }
+
+			{ result.status === 'all-followed' &&
+				<li className={ classNames(styles["no-suggestions"]) }>You already follow { transformFollowingTopics(result.followingTopicsIncludeSearchTerm) }</li> }
```
dan-searle: I realise this is how the existing markup is, but I think it would be better to only render  the `<ul class="suggestions ">` list, if there are suggestions, and in the case of `no-suggestions` or `all-followed`, to render a `<div>` with a different css class name.

So perhaps `ResultContainer` is only used in case of `suggestions`, and then the top level `TopicSearch` is responsible for choosing which template to render based on `result.status`.

components/x-topic-search/src/TopicSearch.jsx
```diff
@@ -0,0 +1,77 @@
...
+				className={ classNames(styles["input"]) }
+				data-trackable="topic-search"
+				onChange={ props.actions.checkInput }
+				onClick={ props.actions.selectInput }
+				onFocus={ props.actions.selectInput }
+				onBlur={ props.actions.hideResult }
+			/>
+		</div>
+
+		{ props.showResult && !props.isLoading && <ResultContainer {...props}/> }
```
dan-searle: As mentioned elsewhere, I think the `result.status` logic should go in here, with the `all-followed` JSX being its own JSX component too.

Also, as a general point I would suggest avoiding prop spreading, as it makes it hard to know which components are relying on which props. Being more explicit also makes it clearer when components are relying on props that they probably shouldn't rely on.

components/x-topic-search/readme.md
fenglish: 👍

components/x-topic-search/src/ResultContainer.jsx
fenglish: 👍

components/x-topic-search/src/TopicSearch.jsx
fenglish: 👍
Thank you for your reviews!!

components/x-topic-search/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@financial-times/x-topic-search",
+  "version": "0.0.0",
+  "description": "",
+  "main": "dist/TopicSearch.cjs.js",
+  "module": "dist/TopicSearch.esm.js",
+  "browser": "dist/TopicSearch.es5.js",
+  "style": "dist/TopicSearch.css",
+  "scripts": {
+    "prepare": "npm run build",
```
dan-searle: Doing this should fix the CI build (and mean other devs don't have to manually run `bower install`). 

```suggestion
    "prepare": "bower install && npm run build",
```

(I only figured this out this week 😀)

components/x-topic-search/src/AllFollowed.jsx
```diff
@@ -0,0 +1,29 @@
+import { h } from '@financial-times/x-engine';
+import styles from './TopicSearch.scss';
+import classNames from 'classnames';
+
+// transform like this => topic1, topic2 and topic3
+const transformFollowedTopics = followedTopicsIncludeSearchTerm => {
```
dan-searle: I think this function can be named more better.

Something like `arrayToSentence` would be more descriptive and agnostic about what the subject matter was?

components/x-topic-search/src/AllFollowed.jsx
```diff
@@ -0,0 +1,29 @@
...
+					// one before the last topic
+					return <span key={ index }><b>{ topic.name }</b> </span>;
+				} else {
+					return <span key={ index }><b>{ topic.name }</b>, </span>;
+				}
+			}
+		})
+};
+
+export default ({ followedTopicsIncludeSearchTerm }) => (
```
dan-searle: Would something like `matchingFollowedTopics` be better than `followedTopicsIncludeSearchTerm`? I'm trying to think of something a bit shorter...

dan-searle:
I think we need to consider if `followedTopics` needs to be updated during the component's lifecycle. E.g. as user follows or unfollows topics.

dan-searle:
@fenglish have a look at the x-teaser-timeline unit tests here: https://github.com/Financial-Times/x-dash/pull/204/files#diff-573aecfd17ec8013aa4d6804eda8694bR22

These are separate to the snapshot tests that happen automatically against stories. You should be able to mock the responses from the search topics service to create the `no-suggestions` and `all-followed` scenarios.

dan-searle:
Looking good.

Are we going to update #228 and merge that into here?

Also, it would be good to update this to a class component #238 before we release this on the web or app.

fenglish:
Do we need to apply #238 before x-topic-search releases on FT.com?
I would prefer to release x-topic-search before the change and merge the PR https://github.com/Financial-Times/next-myft-page/pull/1949 now before apply #238.

edds:

PR #221, opened 2018-11-13
x-teaser-list

dan-searle:
Renders a simple list of content teasers, with optional save buttons.

screenshot 2018-11-20 at 12 00 25

The initial use case is for the myFT Saved Articles page on the app (and web), so the only presentation option is what you see above, but in future we expect that other presentation options will be supported.


components/x-teaser-list/src/TeaserList.jsx
```diff
@@ -0,0 +1,39 @@
...
+							theme="teaser-list"
+						/>
+					</div>
+					{showSaveButtons && <div className={classNames(styles.listItem__actions)}>
+						<ArticleSaveButton
+							id={`${item.id}-save-button`}
+							contentId={item.id}
+							contentTitle={item.title}
+							csrfToken="dummy-token"
+							saved={true}
```
benbarnett: Is this a sensible default for this?

dan-searle: Ah, no it isn't!

components/x-teaser-list/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "build": "node rollup.js",
+    "start": "node rollup.js --watch"
+  },
+  "keywords": [
+    "x-dash"
+  ],
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "@financial-times/x-engine": "file:../../packages/x-engine",
```
i-like-robots: Move this to `peerDependencies`

components/x-teaser-list/readme.md
```diff
@@ -0,0 +1,53 @@
...
+
+[o-teaser](https://registry.origami.ft.com/components/o-teaser) styles will need to be imported by the consumer of this component.
+
+If selectively importing o-teaser's styles via scss, then you will need the following:
+
+```scss
+$o-teaser-is-silent: true;
+@import 'o-teaser/main';
+@include oTeaser(('default'), ('small'));
+```
```
i-like-robots: I think this section should link to the x-teaser docs as there is a risk the instructions will fall out of sync.

components/x-teaser-list/readme.md
```diff
@@ -0,0 +1,53 @@
...
+
+[jsx-wtf]: https://jasonformat.com/wtf-is-jsx/
+
+### Props
+
+Feature           | Type    | Notes
+------------------|---------|----------------------------
+`items`           | Array   | Array of objects representing content items in Teaser format.
+`showSaveButtons` | Boolean | Default = true. Whether to show the save buttons.
+`csrfToken`       | String  | Cross-site Request Forgery token. Required if save buttons are shown.
```
i-like-robots: How does this get generated?

components/x-teaser-list/src/TeaserList.scss
```diff
@@ -0,0 +1,34 @@
+@import 'o-colors/main';
+
+:global {
+	@import "~@financial-times/x-article-save-button/dist/ArticleSaveButton";
+}
```
i-like-robots: Is there a risk this chunk get duplicated if somebody includes the save button component directly?

(I'm not suggesting any fix for this problem, but it's probably worth a comment)

edds:

apaleslimghost (approved):

PR #222, opened 2018-11-14
Make `x-rollup`, `x-babel-config`, `x-test-utils` publishable

maxcbc:
# Make `x-rollup`, `x-babel-config`, `x-test-utils` publishable

## What
Removes private from `x-rollup`, `x-babel-config` and `x-test-utils` so that these can be used in non-x projects.

## Why
Will allow the publishing of these tooling libraries for use by x-like (jsx) projects elsewhere in the ft-ecosystem that might not be stable enough to be promoted to x-dash.

### DIFF
--- a/packages/x-babel-config/package.json
+++ b/packages/x-babel-config/package.json
@@ -1,6 +1,5 @@
 {
   "name": "@financial-times/x-babel-config",
-  "private": true,
   "version": "1.0.0",
   "description": "",
   "main": "index.js",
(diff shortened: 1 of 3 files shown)


maxcbc:
Just trying to get discussion going on this. Basically I think that it'd be nice if `x-dash` set the benchmark for `x-like` projects to follow, so they can aspire to be promoted to `x-dash` and be cross-platform, stable e.t.c

PR #223, opened 2018-11-20
x-teaser-timeline

dan-searle:
Takes an array of content items, and presents them as `x-teasers`, in time/date-based groups, according to the user's _local_ time:

screenshot 2018-11-27 at 14 51 57

The content is grouped as follows:

- Date (always)
- Latest/Earlier today (only if a `latestItemsTime` is provided).

A consumer of this component is expected to provide the user's timezone offset and today's date.

The initial use case is for the myFT followed topics feed.

This component previously existed on a different branch: https://github.com/Financial-Times/x-dash/pull/204


i-like-robots:
It would be really useful if you could add some images of the component and its suggested use case to the PR description.

components/x-teaser-timeline/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "keywords": [
+    "x-dash"
+  ],
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "@financial-times/x-article-save-button": "0.0.11",
+    "@financial-times/x-engine": "file:../../packages/x-engine",
+    "@financial-times/x-teaser": "file:../x-teaser",
+    "classnames": "^2.2.6",
```
i-like-robots: `classnames` or `classcat` @quarterto ? I think we might prefer the latter.

components/x-teaser-timeline/readme.md
```diff
@@ -0,0 +1,69 @@
...
+
+[o-teaser](https://registry.origami.ft.com/components/o-teaser) styles will need to be imported by the consumer of this component.
+
+If selectively importing o-teaser's styles via scss, then you will need the following:
+
+```scss
+$o-teaser-is-silent: true;
+@import 'o-teaser/main';
+@include oTeaser(('default', 'images', 'timestamp'), ('small'));
+```
```
i-like-robots: As per #221 I think this should link out to the x-teaser docs as they will always be up-to-date.

components/x-teaser-timeline/src/TeaserTimeline.jsx
```diff
@@ -0,0 +1,65 @@
+import { h } from '@financial-times/x-engine';
+import { ArticleSaveButton } from '@financial-times/x-article-save-button';
+import { Teaser, presets } from '@financial-times/x-teaser';
+import { getGroupAndIndex, getItemGroups } from './lib/transform';
+import styles from './TeaserTimeline.scss';
+import classNames from 'classnames';
+
+const TeaserTimeline = props => {
+	const {
+		csrfToken = null,
```
i-like-robots: `csrfToken` isn't listed in the readme

components/x-teaser-timeline/src/TeaserTimeline.jsx
```diff
@@ -0,0 +1,65 @@
...
+					<h2 className={classNames(styles.itemGroup__heading)}>{group.title}</h2>
+					<ul className={classNames(styles.itemGroup__items)}>
+						{group.items.map(item => {
+							if (item.id) {
+								return (
+									<li key={item.id} className={styles.item}>
+										<Teaser
+											{...item}
+											{...presets.SmallHeavy}
+											theme="timeline-teaser"
```
i-like-robots: Is `theme` chosen over `modifiers` so that it always "wins"? If so I think a comment is required, otherwise probably better to use `modifiers` as this will override opinion styling at the very least.

components/x-teaser-timeline/src/TeaserTimeline.scss
```diff
@@ -0,0 +1,61 @@
+@import 'o-colors/main';
+@import 'o-grid/main';
+@import 'o-typography/main';
+
+:global {
```
i-like-robots: As per #221 it might be worth a comment stating that this could cause duplicate output.

(to avoid this perhaps the docs could instruct implementers to include the save button styles manually?)

package.json
```diff
@@ -19,7 +19,7 @@
     "eslint-config-prettier": "^2.9.0",
     "eslint-plugin-jest": "^21.17.0",
     "eslint-plugin-jsx-a11y": "^6.1.1",
-    "eslint-plugin-react": "^7.10.0",
+    "eslint-plugin-react": "7.10.0",
```
i-like-robots: ?

components/x-teaser-timeline/package.json
debugwand: we are already using classnames in x-styling-demo

classcat is about 70b smaller
https://bundlephobia.com/result?p=classnames@2.2.6
https://bundlephobia.com/result?p=classcat@3.2.5

quarterto: i had run into bundling problems with `classnames` but this might have been fixed by updates to it or our tooling 🤷‍♀️

components/x-teaser-timeline/src/TeaserTimeline.jsx
dan-searle: I'm pretty sure there isn't a good reason to use `theme`, and `modifiers` seems more appropriate. I'll change it.

package.json
dan-searle: Ah, there was an obscure issue in CI (if I recall correctly) with a patch release of this module. I suspect it's since been fixed.

components/x-teaser-timeline/src/TeaserTimeline.jsx
```diff
@@ -0,0 +1,65 @@
...
+		items,
+		showSaveButtons = true
+	} = props;
+	const itemGroups = getItemGroups(props);
+
+	if (customSlotContent) {
+		const insertPosition = Math.min(customSlotPosition, items.length);
+		const insert = getGroupAndIndex(itemGroups, insertPosition);
+
+		itemGroups[insert.group].items.splice(insert.index, 0, customSlotContent);
```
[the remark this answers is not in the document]
dan-searle: It's safe, but yeah, it would be safer to clone it first.

edds:

PR #225, opened 2018-11-26
Fix layout issue now we have 1 extra item

debugwand:
Fix a layout issue in the documentation https://financial-times.github.io/x-dash/components/x-teaser

WAS
screen shot 2018-11-26 at 15 59 01

NOW
screen shot 2018-11-26 at 15 58 01

### DIFF
--- a/tools/x-docs/src/components/story-viewer/index.jsx
+++ b/tools/x-docs/src/components/story-viewer/index.jsx
@@ -25,18 +25,20 @@ class StoryViewer extends React.Component {
 		return (
 			<div id="component-demos" className="story-viewer">
 				<h2 className="story-viewer__heading">Component demos</h2>
-				<div className="story-viewer__list" role="tablist">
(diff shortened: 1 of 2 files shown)


tools/x-docs/src/components/story-viewer/index.jsx
```diff
@@ -25,18 +25,20 @@ class StoryViewer extends React.Component {
...
-							key={`story-${i}`}
-							role="tab"
-							className="story-viewer__item"
-							aria-selected={this.state.selected === i}
-							onClick={this.onClick.bind(this, i)}>
-							{story}
-						</button>
+						<li key={`story-${i}`} className="story-viewer__item">
+							<button
+								key={`story-${i}`}
```
i-like-robots: Small thing: a `key` is not required here, it's only needed on the parent element

i-like-robots:
Cool, I remember building the first version of the companies header right back when I started... 👴

PR #227, opened 2018-12-05
Add a 'className' prop so consumers can add additional classes

benbarnett:
__Please note this PR is only to merge into the component branch, not master__

This prop has been added because, for example, the app needs an 'online-only' class name for these components.


i-like-robots:
Side note: the x-teaser component calls these "modifiers" as the component remains in control of the base class name prefix: https://github.com/Financial-Times/x-dash/blob/master/components/x-teaser/src/Container.jsx

Is the intention with this to provide the full class name or just an `--online-only` suffix?

benbarnett:
So the intention here is to provide a full class name rather than a modifier. The `online-only` class is specific to the app (it is applied by the app, to the `<body>` when it thinks it is offline).

dan-searle:
This wouldn't be a modifier class, as it's not something that the x-dash component is responsible for implementing the styles for. 

This is a tricky one, because we wouldn't want this to abused by a consumer using it make its own custom variants of an x-dash component.

benbarnett:
As just discussed with @dan-searle we can just wrap the component with the class name we need in this case. Given we only have one use-case at the moment, on balance it seems better to avoid this as an option at the moment.

PR #228, opened 2019-01-06
X topic search: modify tests

fenglish:
Use unit testing style (not snapshots)

There are a lot of linting errors. I can't figure out the reason...


quarterto:
other components use enzyme (via the internal package `x-test-utils`) instead of react-test-utils, can you refactor to use that please?

the lint error seem to be from storybook's build output. i don't know why that's being included in the linting now. adding `x-docs/static/**` to https://github.com/Financial-Times/x-dash/blob/master/.eslintignore should silence these.

__tests__/__snapshots__/snapshots.test.js.snap
```diff
@@ -5555,9 +5555,9 @@ exports[`@financial-times/x-topic-search renders a default Topic Search Bar x-to
       data-trackable="topic-search"
       id="topic-search-input"
       onBlur={[Function]}
+      onChange={[Function]}
```
fenglish: @dan-searle I'm not sure with the change, `onChange` instead of `onInput`. I heard form Ben and Matt, we should use `onInput` for React instead of `onChange`. It works differently, I heard. I lost the conversation log because of forced git push... 😞 
^^ cc @benbarnett

fenglish: Yeah, I know Enzyme doesn't do for `mount`. It is only available for `shallow` rendering.
https://airbnb.io/enzyme/docs/future.html#improved-event-simulation-and-propagation-support
 That's the reason I used `react-test-utils`.

dan-searle: Enzyme's `simulate` does work with `mount`, the problem was specifically simulating an event that would trigger the `oninput` event handler.

I found this article about `onchange` vs `oninput`, but I can't find the official React advice it alludes to:

> the React team recommends using onChange over onInput. The bottom line is that React’s onChange works differently than onChange in HTML, in that it’s more consistent (and more like HTML’s onInput). onChange is triggered on every change and not on the loss of focus.

https://medium.com/capital-one-tech/how-to-work-with-forms-inputs-and-events-in-react-c337171b923b

benbarnett: Just caught up with Matt about this as I couldn't remember either:

`onInput` should fire on every key change, `onChange` should fire only when the text field loses focus

However, in React, `onChange` fires like `onInput` does. So behaviourally, this is ok, but really what we mean here is `onInput` because we do actually care about key changes as they happen, not just when the element loses focus. So its just being slightly more explicit.

dan-searle: I think in a React/JSX world `onChange` is the idiomatic norm, and using `onInput` could suggest there was some non-obvious technical reason for doing it differently, which in this case at least, would be misleading.

benbarnett: Sounds reasonable to me. I can't remember if there was more context to this. @i-like-robots do you agree `onChange` is alright in this case?

dan-searle: Actually, I've just read that Preact's `onChange` behaviour is the same as the native browser one, so we should use `onInput`:

https://github.com/developit/preact/issues/1034

PR #232, opened 2019-01-17
Add topicFollowed/topicUnfollowed externally triggered actions

fenglish:
To update `followedTopics` after clicking myFT follow buttons, add topicFollowed/topicUnfollowed actions which are triggered externally.
https://financial-times.github.io/x-dash/components/x-interaction#triggering-actions-externally

### DIFF
--- a/components/x-topic-search/src/TopicSearch.jsx
+++ b/components/x-topic-search/src/TopicSearch.jsx
@@ -49,7 +49,7 @@ const topicSearchActions = withActions(({ minSearchLength = 2, maxSuggestions = 
 	topicUnfollowed (subjectId) {
 		const unfollowedTopicIndex = getFollowedTopicIndex(followedTopics, subjectId);
 
-		if (unfollowedTopicIndex > 0) {
+		if (unfollowedTopicIndex > -1) {
 			followedTopics.splice(unfollowedTopicIndex, 1);
 		}
 
--- a/components/x-topic-search/src/TopicSearch.jsx
+++ b/components/x-topic-search/src/TopicSearch.jsx
@@ -11,7 +11,7 @@ import AllFollowed from './AllFollowed';
 
 const debounceGetSuggestions = debounce(getSuggestions, 150);
 const getFollowedTopicIndex = (followedTopics, targetTopicId) => (
-	followedTopics.findIndex(topic => topic && topic.conceptId === targetTopicId)
+	followedTopics.findIndex(topic => topic && topic.conceptId || topic.uuid === targetTopicId)
 );
 
 let resultExist = false;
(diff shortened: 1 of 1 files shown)


components/x-topic-search/src/TopicSearch.jsx
```diff
@@ -43,8 +43,26 @@ const topicSearchActions = withActions(({ minSearchLength = 2, maxSuggestions =
 	}
 }));
 
-const TopicSearch = topicSearchActions(({ searchTerm, showResult, result, actions, isLoading, csrfToken }) => (
-	<div className={ classNames(styles['container']) }>
+const TopicSearch = topicSearchActions(({ searchTerm, showResult, result, actions, isLoading, csrfToken, followedTopics }) => {
+
+	document.body.addEventListener('x-follow-button', event => {
```
dan-searle: I think the action function this lives in will get called multiple times, meaning multiple event listeners will be added.

I don't think there's a better place to put this, so I think the solution is to extract the event handler function, and then remove the listener before adding it. Then only one such listener will ever exist. Something like:

```javascript
const handleFollowEvent = event => {
   // unchanged...
};

document.body.removeEventListener('x-follow-button', handleFollowEvent);
document.body.addEventListener('x-follow-button', handleFollowEvent);
```

dan-searle: The `x-follow-button` event just means a button has been clicked on, it doesn't mean the follow/unfollow action has been successfully performed. Also, if a topic is followed via some other means, this won't know about it.

I think it would be better to listen for the events from myft-client, like here: https://github.com/Financial-Times/n-myft-ui/blob/master/components/x-button-integration/index.js#L27

components/x-topic-search/src/TopicSearch.jsx
```diff
@@ -33,6 +36,22 @@ const topicSearchActions = withActions(({ minSearchLength = 2, maxSuggestions =
...
+		if (followedTopicIndex === -1) {
+			followedTopics.push({ name: subjectName, uuid: subjectId })
+		}
+	},
+
+	topicUnfollowed (subjectId) {
+		const unfollowedTopicIndex = getFollowedTopicIndex(followedTopics, subjectId);
+
+		if (unfollowedTopicIndex > 0) {
+			delete followedTopics[ unfollowedTopicIndex ];
```
dan-searle: `delete` on an array will leave an empty element - it won't reindex the array or affect the array `.length`

`splice` does the job properly:

```suggestion
			followedTopics.splice(unfollowedTopicIndex, 1);
```

components/x-topic-search/src/TopicSearch.jsx
```diff
@@ -33,6 +36,22 @@ const topicSearchActions = withActions(({ minSearchLength = 2, maxSuggestions =
 		}
 	},
 
+	topicFollowed (subjectId, subjectName) {
+		const followedTopicIndex = getFollowedTopicIndex(followedTopics, subjectId);
+
+		if (followedTopicIndex === -1) {
+			followedTopics.push({ name: subjectName, uuid: subjectId })
+		}
```
dan-searle: do you need to `return { followedTopics };` at the end of this function and `topicUnfollowed`, so the component will update?

components/x-topic-search/src/TopicSearch.jsx
fenglish: I see!! Thank you!

fenglish:
Thank you, @dan-searle!

PR #234, opened 2019-01-22
Fetch suggestions include followed topics

fenglish:
This change is to update followedTopics properly when users click an external follow button. x-topic-search executes the topicFollowed action (which is triggered in the consumer side) when that happens. x-topic-search gets only the topic id from the event, but it needs the whole topic data to render correct info.

Changing from **Fetch suggestions except followedTopics** to **Fetch all suggestions and then pick followedTopic up by the ids** will make easier to update followedTopics with the whole topic data.

### DIFF
--- a/components/x-topic-search/readme.md
+++ b/components/x-topic-search/readme.md
@@ -36,10 +36,10 @@ All `x-` components are designed to be compatible with a variety of runtimes, no
 
 ### Properties
 
-Property         | Type   | Required | Note
------------------|--------|----------|------------------
-`minSearchLength`| Number | No       | Minimum chars to start search. Default is 2
-`maxSuggestions` | Number | No       | Maximum number to display suggestions. Default is 5
-`apiUrl`         | String | Yes      | The url to use when making requests to get topics
-`followedTopics` | Array  | Yes      | Each item should have `name` and `uuid` properties. Default is `[]`
-`csrfToken`      | String | Yes      | Value included in a hidden form field for x-follow-button
+Property          | Type   | Required | Note
+------------------|--------|----------|------------------
+`minSearchLength` | Number | No       | Minimum chars to start search. Default is 2
+`maxSuggestions`  | Number | No       | Maximum number to display suggestions. Default is 5
+`apiUrl`          | String | Yes      | The url to use when making requests to get topics
+`followedTopicIds`| Array  | Yes      | Array of followed topic `id`s. Default is `[]`
+`csrfToken`       | String | Yes      | Value included in a hidden form field for x-follow-button
--- a/components/x-topic-search/readme.md
+++ b/components/x-topic-search/readme.md
@@ -41,5 +41,5 @@ Property          | Type   | Required | Note
 `minSearchLength` | Number | No       | Minimum chars to start search. Default is 2
 `maxSuggestions`  | Number | No       | Maximum number to display suggestions. Default is 5
 `apiUrl`          | String | Yes      | The url to use when making requests to get topics
-`followedTopicIds`| Array  | Yes      | Array of followed topic `id`s. Default is `[]`
+`followedTopicIds`| Array  | Yes      | Array of followed topic `id`s.
 `csrfToken`       | String | Yes      | Value included in a hidden form field for x-follow-button
--- a/components/x-topic-search/src/AllFollowed.jsx
+++ b/components/x-topic-search/src/AllFollowed.jsx
@@ -10,19 +10,17 @@ const arrayToSentence = followedSuggestions => {
 		return <b>{ followedSuggestions[0].prefLabel }</b>;
 	} else {
 		return followedSuggestions
-		.map((topic, index) => {
-			if (index + 1 === topicsLength) {
-				// the last topic
-				return  <span key={ topic.id }>and <b>{ topic.prefLabel }</b></span>
-			} else {
-				if ((topicsLength - 2) === index) {
+			.map((topic, index) => {
+				if (index === topicsLength - 1) {
+					// the last topic
+					return  <span key={ topic.id }>and <b>{ topic.prefLabel }</b></span>
+				} else if (index === topicsLength - 2) {
 					// one before the last topic
 					return <span key={ topic.id }><b>{ topic.prefLabel }</b> </span>;
 				} else {
 					return <span key={ topic.id }><b>{ topic.prefLabel }</b>, </span>;
 				}
-			}
-		})
+			})
 	}
 
 };
--- a/components/x-topic-search/src/AllFollowed.jsx
+++ b/components/x-topic-search/src/AllFollowed.jsx
@@ -13,13 +13,13 @@ const arrayToSentence = followedSuggestions => {
 		.map((topic, index) => {
 			if (index + 1 === topicsLength) {
 				// the last topic
-				return  <span key={ index }>and <b>{ topic.prefLabel }</b></span>
+				return  <span key={ topic.id }>and <b>{ topic.prefLabel }</b></span>
 			} else {
 				if ((topicsLength - 2) === index) {
 					// one before the last topic
-					return <span key={ index }><b>{ topic.prefLabel }</b> </span>;
+					return <span key={ topic.id }><b>{ topic.prefLabel }</b> </span>;
 				} else {
-					return <span key={ index }><b>{ topic.prefLabel }</b>, </span>;
+					return <span key={ topic.id }><b>{ topic.prefLabel }</b>, </span>;
 				}
 			}
 		})
--- a/components/x-topic-search/src/AllFollowed.jsx
+++ b/components/x-topic-search/src/AllFollowed.jsx
@@ -3,13 +3,13 @@ import styles from './TopicSearch.scss';
 import classNames from 'classnames';
 
 // transform like this => topic1, topic2 and topic3
-const arrayToSentence = matchingFollowedTopics => {
-	const topicsLength = matchingFollowedTopics.length;
+const arrayToSentence = followedSuggestions => {
+	const topicsLength = followedSuggestions.length;
 
 	if (topicsLength === 1) {
-		return <b>{ matchingFollowedTopics[0].prefLabel }</b>;
+		return <b>{ followedSuggestions[0].prefLabel }</b>;
 	} else {
-		return matchingFollowedTopics
+		return followedSuggestions
 		.map((topic, index) => {
 			if (index + 1 === topicsLength) {
 				// the last topic
@@ -27,8 +27,8 @@ const arrayToSentence = matchingFollowedTopics => {
 
 };
 
-export default ({ matchingFollowedTopics }) => (
+export default ({ followedSuggestions }) => (
 	<div className={ classNames(styles["all-followed"]) } aria-live="polite">
-		You already follow { arrayToSentence(matchingFollowedTopics) }
+		You already follow { arrayToSentence(followedSuggestions) }
 	</div>
 );
--- a/components/x-topic-search/src/AllFollowed.jsx
+++ b/components/x-topic-search/src/AllFollowed.jsx
@@ -7,19 +7,19 @@ const arrayToSentence = matchingFollowedTopics => {
 	const topicsLength = matchingFollowedTopics.length;
 
 	if (topicsLength === 1) {
-		return <b>{ matchingFollowedTopics[0].name }</b>;
+		return <b>{ matchingFollowedTopics[0].prefLabel }</b>;
 	} else {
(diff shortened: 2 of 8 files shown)


components/x-topic-search/src/lib/get-suggestions.js
```diff
@@ -30,13 +50,6 @@ export default (searchTerm, maxSuggestions, apiUrl, followedTopics) => {
 	const dataSrc = addQueryParamToUrl('count', maxSuggestions, apiUrl, false);
 	let url = addQueryParamToUrl('partial', searchTerm.replace(' ', '+'), dataSrc);
```
dan-searle: Can use `const` instead of `let`, now `url` isn't being changed.

components/x-topic-search/src/lib/get-suggestions.js
```diff
@@ -3,23 +3,43 @@ const addQueryParamToUrl = (name, value, url, append = true) => {
 	return append === true ? `${url}&${queryParam}` : `${url}?${queryParam}`;
 };
 
+const separateFollowedAndNot = (suggestions, followedTopics) => {
+	const matchingFollowedTopics = [];
+	const matchingNotFollowedTopics = [];
```
dan-searle: `matchingUnfollowedTopics` might be more consistent with what's used elsewhere.

fenglish: @dan-searle I thought the word `unfollow` is an action to stop following something. For this case, the topics are not followed by the user. I thought `matchingNotFollowedTopics ` is more appropriate. However I may have used `unfollow` in the same case in somewhere :sweat_smile:

components/x-topic-search/src/lib/get-suggestions.js
fenglish: Oh! Thank you!

components/x-topic-search/src/lib/get-suggestions.js
fenglish: I found the wrong use of `unfollowed` in only tests, However the wrong use will be removed by this PR.
https://github.com/Financial-Times/x-dash/pull/228

dan-searle: Yeah, "unfollow" is the action, and "unfollowed" would describe that it's not currently followed.

fenglish: :open_mouth: In that case, I'm going to modify the name immediately :dash:

components/x-topic-search/src/lib/get-suggestions.js
```diff
@@ -3,39 +3,47 @@ const addQueryParamToUrl = (name, value, url, append = true) => {
 	return append === true ? `${url}&${queryParam}` : `${url}?${queryParam}`;
 };
 
-const suggest = function (suggestions, followedTopics, searchTerm) {
+const separateFollowedAndNot = (suggestions, followedTopics) => {
+	const matchingFollowedTopics = [];
+	const matchingUnfollowedTopics = [];
+
+	if (!followedTopic.length) {
```
dan-searle: I don't think this is necessary. If `followedTopics` is an empty array, then the `.forEach` below will effectively do nothing, and then the function can just `return { matchingFollowedTopics, matchingUnfollowedTopics }` (which will be two empty arrays).

components/x-topic-search/src/lib/get-suggestions.js
```diff
@@ -3,39 +3,47 @@ const addQueryParamToUrl = (name, value, url, append = true) => {
 	return append === true ? `${url}&${queryParam}` : `${url}?${queryParam}`;
 };
 
-const suggest = function (suggestions, followedTopics, searchTerm) {
+const separateFollowedAndNot = (suggestions, followedTopics) => {
+	const matchingFollowedTopics = [];
+	const matchingUnfollowedTopics = [];
```
dan-searle: I think some of the conditions can be removed by taking a slightly different approach to this:

```javascript
const separateFollowedAndUnfollowed = (suggestions, followedTopics) => {
	const followedSuggestions = suggestions.filter(suggestion => followedTopics.some(topic => topic.uuid === suggestion.id));
	const unfollowedSuggestions = suggestions.filter(suggestion => !followedTopics.some(topic => topic.uuid === suggestion.id));

	return { followedSuggestions, unfollowedSuggestions };
};

const suggest = function (suggestions, followedTopics) {
	suggestions.forEach((suggestion) => {
		if (suggestion && !suggestion.url){
			// TODO App needs different url?
			suggestion.url = '/stream/' + suggestion.id;
		}
	});

	return separateFollowedAndUnfollowed(suggestions, followedTopics);
};
```

components/x-topic-search/src/lib/get-suggestions.js
```diff
@@ -3,39 +3,47 @@ const addQueryParamToUrl = (name, value, url, append = true) => {
...
+	return { matchingFollowedTopics, matchingUnfollowedTopics };
+}
+
+const suggest = function (suggestions, followedTopics) {
+
 	if (suggestions.length) {
 		suggestions.forEach((suggestion) => {
 			if (suggestion && !suggestion.url){
 				// TODO App needs different url?
 				suggestion.url = '/stream/' + suggestion.id;
```
dan-searle: This only needs to be done to suggestions that will be rendered? If so, then perhaps this is something that should be done at the point of render in `SuggestionList.jsx`?

```jsx
				<a
					data-trackable="topic-link
					className={classNames(styles["suggestion__name"])}
					href={`/stream/${suggestion.url}`}
				>
					{ suggestion.prefLabel }
				</a>
```

components/x-topic-search/src/lib/get-suggestions.js
```diff
@@ -3,39 +3,47 @@ const addQueryParamToUrl = (name, value, url, append = true) => {
...
+	}));
+
+	return { matchingFollowedTopics, matchingUnfollowedTopics };
+}
+
+const suggest = function (suggestions, followedTopics) {
+
 	if (suggestions.length) {
 		suggestions.forEach((suggestion) => {
 			if (suggestion && !suggestion.url){
```
dan-searle: Will some but not all suggestions have a `url`?

components/x-topic-search/src/AllFollowed.jsx
```diff
@@ -3,32 +3,32 @@ import styles from './TopicSearch.scss';
...
-		return <b>{ matchingFollowedTopics[0].name }</b>;
+		return <b>{ followedSuggestions[0].prefLabel }</b>;
 	} else {
-		return matchingFollowedTopics
+		return followedSuggestions
 		.map((topic, index) => {
 			if (index + 1 === topicsLength) {
 				// the last topic
-				return  <span key={ index }>and <b>{ topic.name }</b></span>
+				return  <span key={ index }>and <b>{ topic.prefLabel }</b></span>
```
dan-searle: Practically, it probably won't make much difference in this case, but it's best to avoid using array indexes for `key`, and instead use something that identifies the element value itself.

```suggestion
				return  <span key={ topic.id }>and <b>{ topic.prefLabel }</b></span>
```

This helps p/react know which DOM elements have been added/removed and can make rendering more efficient.

fenglish: 👍

components/x-topic-search/src/AllFollowed.jsx
```diff
@@ -3,32 +3,32 @@ import styles from './TopicSearch.scss';
...
+		return <b>{ followedSuggestions[0].prefLabel }</b>;
 	} else {
-		return matchingFollowedTopics
+		return followedSuggestions
 		.map((topic, index) => {
 			if (index + 1 === topicsLength) {
 				// the last topic
-				return  <span key={ index }>and <b>{ topic.name }</b></span>
+				return  <span key={ topic.id }>and <b>{ topic.prefLabel }</b></span>
 			} else {
```
benbarnett: Probably don't need the `else` block here as you're `returning` in the `if` statement above. Might make this block slightly cleaner?

components/x-topic-search/src/AllFollowed.jsx
```diff
@@ -3,32 +3,32 @@ import styles from './TopicSearch.scss';
...
+	const topicsLength = followedSuggestions.length;
 
 	if (topicsLength === 1) {
-		return <b>{ matchingFollowedTopics[0].name }</b>;
+		return <b>{ followedSuggestions[0].prefLabel }</b>;
 	} else {
-		return matchingFollowedTopics
+		return followedSuggestions
 		.map((topic, index) => {
 			if (index + 1 === topicsLength) {
```
benbarnett: As per line 18, for some reason `if (index === topicsLength - 1)` is (to me) more readable. What do you think?

components/x-topic-search/src/TopicSearch.jsx
```diff
@@ -10,18 +10,15 @@ import NoSuggestions from './NoSuggestions';
...
-const topicSearchActions = withActions(({ minSearchLength = 2, maxSuggestions = 5, apiUrl, followedTopics = [] }) => ({
+const topicSearchActions = withActions(({ minSearchLength = 2, maxSuggestions = 5, apiUrl, followedTopicIds = [] }) => ({
 	async checkInput(event) {
 		const searchTerm = event.target.value && event.target.value.trim();
 
 		if (searchTerm.length >= minSearchLength) {
-			return debounceGetSuggestions(searchTerm, maxSuggestions, apiUrl, followedTopics)
+			return debounceGetSuggestions(searchTerm, maxSuggestions, apiUrl, followedTopicIds)
 				.then(result => {
 					resultExist = true;
```
benbarnett: `resultExists` would probably be more accurate here

components/x-topic-search/readme.md
```diff
@@ -36,10 +36,10 @@ All `x-` components are designed to be compatible with a variety of runtimes, no
...
-`maxSuggestions` | Number | No       | Maximum number to display suggestions. Default is 5
-`apiUrl`         | String | Yes      | The url to use when making requests to get topics
-`followedTopics` | Array  | Yes      | Each item should have `name` and `uuid` properties. Default is `[]`
-`csrfToken`      | String | Yes      | Value included in a hidden form field for x-follow-button
+Property          | Type   | Required | Note
+------------------|--------|----------|------------------
+`minSearchLength` | Number | No       | Minimum chars to start search. Default is 2
+`maxSuggestions`  | Number | No       | Maximum number to display suggestions. Default is 5
+`apiUrl`          | String | Yes      | The url to use when making requests to get topics
+`followedTopicIds`| Array  | Yes      | Array of followed topic `id`s. Default is `[]`
```
benbarnett: Is this required if the default is `[]`?

components/x-topic-search/src/AllFollowed.jsx
```diff
@@ -3,32 +3,32 @@ import styles from './TopicSearch.scss';
...
 				// the last topic
-				return  <span key={ index }>and <b>{ topic.name }</b></span>
+				return  <span key={ topic.id }>and <b>{ topic.prefLabel }</b></span>
 			} else {
 				if ((topicsLength - 2) === index) {
```
benbarnett: Similar to above, I would swap these conditions around, so this becomes `if (index === topicsLength - 2)`

components/x-topic-search/src/AllFollowed.jsx
fenglish: I don't have a strong opinion with this. This part is come from the original I guess 🤔 But I think yours is more clean! I'm going to change it. Thank you!!

components/x-topic-search/src/AllFollowed.jsx
fenglish: 👍

components/x-topic-search/src/TopicSearch.jsx
fenglish: 👍

components/x-topic-search/readme.md
dan-searle: Yeah, I think just remove the mention of a default from the readme, but keep it in the code?

benbarnett: Yep sounds good

components/x-topic-search/src/AllFollowed.jsx
fenglish: :open_mouth: No, don't need the `else` block! Thank you!

components/x-topic-search/readme.md
fenglish: Ok! I will do that.

fenglish:
@dan-searle @benbarnett Thanks both of you!

PR #235, opened 2019-01-23
Remove onBlur which hides result

fenglish:
Can't click follow button in topic search result because the result disappears (this happens because of `onBlur` on `<input>`). Hiding the result container need to be controled by external triggers.


dan-searle:
I'm wondering if this event listener could be set by `x-topic-search` after all? Although it's attaching the listener to an element outside of the component, it won't cause any side-effects to anything else, so it should be safe.

PR #238, opened 2019-01-24
Extend Component class instead of using x-interaction

dan-searle:
**Work in progress... do not merge**


quarterto:
what's not working for you about x-interaction?

dan-searle:
@quarterto might be quickest/easiest to talk through in person?

dan-searle:
This branch now includes cherry-pick of #241

components/x-topic-search/src/TopicSearch.jsx
```diff
@@ -1,95 +1,125 @@
...
-		return { followedTopicIds };
-	},
-
-	topicUnfollowed (subjectId) {
-		const targetIdIndex = followedTopicIds.indexOf(subjectId);
-
-		if (targetIdIndex > -1) {
-			followedTopicIds.splice(targetIdIndex, 1);
+	handleInputClickOrFocus() {
+		if (this.state.searchTerm.length >= this.minSearchLength) {
```
fenglish: @dan-searle This line will prevent the bug which it displays results even if there is no searchTerm?

components/x-topic-search/src/lib/get-suggestions.js
```diff
@@ -20,13 +13,15 @@ export default (searchTerm, maxSuggestions, apiUrl, followedTopicIds) => {
...
-		})
-		.catch(() => {
-			throw new Error();
+		.then(suggestions => ({
+			resultsForTerm: searchTerm,
+			followedSuggestions: suggestions.filter(suggestion => followedTopicIds.includes(suggestion.id)),
+			unfollowedSuggestions: suggestions.filter(suggestion => !followedTopicIds.includes(suggestion.id))
+		}))
+		.catch((err) => {
+			throw new Error(err);
```
benbarnett: Do we need this if its just going to throw anyway?

components/x-topic-search/stories/knobs.js
```diff
@@ -0,0 +1,19 @@
+module.exports = (data, { select }) => {
+	return {
+		followedTopicIds() {
+			return select(
+				'Followed Topics',
+				{
+					None: [],
+					'World Elephant Water Polo': ['f95d1e16-2307-4feb-b3ff-6f224798aa49'],
```
benbarnett: 👍

components/x-topic-search/src/TopicSearch.jsx
dan-searle: Unfortunately not. This only happens when you click into the input. The bug happens when the input already has focus, a request has been made for suggestions, and the response arrives after the results container has been closed (because the input length < 2)

components/x-topic-search/src/TopicSearch.jsx
```diff
@@ -1,95 +1,125 @@
...
+					<input
+						type="search"
+						id="topic-search-input"
+						placeholder="Search and add topics"
+						className={ classNames(styles["input"]) }
+						data-trackable="topic-search"
+						autoComplete="off"
+						onInput={this.handleInputChange}
+						onClick={this.handleInputClickOrFocus}
+						onFocus={this.handleInputClickOrFocus}
```
fenglish: Space consistency for these three brackets? Although I know you prefer without spaces... I love spaces...
```
onInput={ this.handleInputChange }
onClick={ this.handleInputClickOrFocus }
onFocus={ this.handleInputClickOrFocus }
```

components/x-topic-search/src/ResultContainer.jsx
```diff
@@ -9,49 +9,37 @@ import NoSuggestions from './NoSuggestions';
...
-					return <span key={ topic.id }><b>{ topic.prefLabel }</b>, </span>;
-				}
-			})
-	}
-
+	return followedSuggestions.map((topic, index) => (
+		<span key={ topic.id }>
+			{topicsLength > 1 && index === topicsLength - 1 && ' and '}
+			<b>{topic.prefLabel}</b>
+			{index < topicsLength - 2 && ', '}
```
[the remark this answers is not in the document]
fenglish: ... I wonder what this means. Dinner with an old woman ... 😑

components/x-topic-search/src/TopicSearch.jsx
dan-searle: I don't feel strongly either way, but I tend to follow airbnb react lint rules, as (I believe) it's fairly widely adopted:

http://airbnb.io/javascript/react/#spacing

I agree it should be consistent though, but without an agreed standard (for all x-dash devs), this could go back and forward endlessly. We could raise a PR which would enable the lint rule for this and then let people argue in the PR until someone clicks merge 😈

components/x-topic-search/src/ResultContainer.jsx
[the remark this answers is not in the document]
dan-searle: It's a reference to this: https://github.com/Financial-Times/ft-app/pull/93#pullrequestreview-198153413

components/x-topic-search/src/TopicSearch.jsx
dan-searle: Actually, I had to revert that way of fixing it, as it didn't work very well. There's a different way of fixing it now.

PR #241, opened 2019-01-29
Support class components in story snapshot tests

### LINKED ISSUE #240: x-dash test fails when component is a class
When we extend Preact's `Component` component class instead of using `x-interaction` for components that have state, and run `npm run jest`, we get this error:

```console
  ● @financial-times/x-topic-search › renders a default Topic Search Bar x-topic-search

    TypeError: Cannot call a class as a function
```

dan-searle:
Fixes #240

### DIFF
--- a/__tests__/snapshots.test.js
+++ b/__tests__/snapshots.test.js
@@ -17,7 +17,7 @@ for(const pkg of packageDirs) {
 	const storiesDir = path.resolve(pkgDir, 'stories');
 
 	if(fs.existsSync(storiesDir)) {
-		const { package: pkg, stories, component: Component } = require(storiesDir);
+		const { package: pkg, stories, component } = require(storiesDir);
 		const { presets = { default: {} } } = require(pkgDir);
 		const name = path.basename(pkg.name);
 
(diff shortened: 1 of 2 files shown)


__tests__/snapshots.test.js
```diff
@@ -1,7 +1,9 @@
+/* eslint-disable no-unused-vars, react/react-in-jsx-scope */
```
dan-searle: Is there a better alternative to this?

__tests__/snapshots.test.js
```diff
@@ -16,16 +18,16 @@ for(const pkg of packageDirs) {
...
 		const name = path.basename(pkg.name);
 
 		describe(pkg.name, () => {
 			for (const { title, data } of stories) {
-				for (const [ preset, options ] of Object.entries(presets)) {
+				for (const [preset, options] of Object.entries(presets)) {
 					it(`renders a ${preset} ${title} ${name}`, () => {
 						const props = { ...data, ...options };
-						const tree = renderer.create(component(props)).toJSON();
+						const tree = renderer.create(<Component {...props} />).toJSON();
```
i-like-robots: Could switch this to `h(Component, props)` to avoid the lint rule?

dan-searle:
@i-like-robots thanks for the tip, I've made that change now.

__tests__/snapshots.test.js
```diff
@@ -16,16 +17,16 @@ for(const pkg of packageDirs) {
 	const storiesDir = path.resolve(pkgDir, 'stories');
 
 	if(fs.existsSync(storiesDir)) {
-		const { package: pkg, stories, component } = require(storiesDir);
-		const { presets = {default: {}} } = require(pkgDir);
+		const { package: pkg, stories, component: Component } = require(storiesDir);
```
i-like-robots: I'm guessing the casing is no longer important without the JSX syntax?

dan-searle: Oh yes, of course.

PR #242, opened 2019-01-29
Apply custom styles to search cancel button

fenglish:
The search cancel button is supported only in WebKit and Blink, hence the vendor prefix `::-webkit-` is needed.

**BEFORE**
screen shot 2019-01-29 at 16 08 31

**AFTER**
screen shot 2019-01-29 at 16 06 29

### DIFF
--- a/components/x-topic-search/src/TopicSearch.scss
+++ b/components/x-topic-search/src/TopicSearch.scss
@@ -42,6 +42,11 @@
 	&::placeholder {
 		color: oColorsGetPaletteColor('white');
 	}
+
+	&::-webkit-search-cancel-button {
+		@include oIconsGetIcon('cross', oColorsGetPaletteColor('white'), 26);
+		-webkit-appearance: none;
+	}
 }
 
 .result-container {


components/x-topic-search/src/TopicSearch.scss
```diff
@@ -42,6 +42,11 @@
 	&::placeholder {
 		color: oColorsGetPaletteColor('white');
 	}
+
+	&::-webkit-search-cancel-button {
+	  -webkit-appearance: none;
+		@include oIconsGetIcon('cross', oColorsGetPaletteColor('white'), 26);
```
dan-searle: Looks like the indentation isn't quite right here?

Also, I believe it's SASS best practice to `@include` mixins before property declarations. This is just in case the mixin sets the same property you're also trying to set directly - CSS will always take the second value of a duplicate property.

fenglish: @dan-searle ` -webkit-appearance: none;` need to declare before `@include` to remove the default styles applied by Webkit. I'm going to add a comment about that.

dan-searle: The `oIconsGetIcon` mixin doesn't output the `-webkit-appearance` property though? 

https://github.com/Financial-Times/o-icons/blob/master/scss/_mixins.scss#L17

fenglish: !?

fenglish: I'm confused.

PR #245, opened 2019-02-01
Set input height to keep result container's position consistently

fenglish:
**The bug**
-----
Only on Safari, the input box's height is set differently.
Chrome, Firefox, IE11 => 40px
Safari => 46px
wrong-top-position

**The reason**
-----

It seems Safari includes the height of search cancel button as the input height and Chrome doesn't.

css-difference

```css
input {
    min-height: 40px;
    padding: 9px 9px 9px;
    font-size: 16px;
    line-height: 20px;
    margin: 0;
    border-bottom: 2px solid #ffffff;
    padding-left: 24px;

    &::-webkit-search-cancel-button {
	@include oIconsGetIcon('cross', oColorsGetPaletteColor('white'), 26);
	-webkit-appearance: none;
    }
}
```

**The solution**
-----
Adding `position: absolute` or `z-index` doesn't solve the problem. Overwrite padding top/bottom 0 and keep the input height the min-height(40px) no matter whether the browser includes the cancel button height as the input height or not.

### DIFF
--- a/components/x-topic-search/src/TopicSearch.scss
+++ b/components/x-topic-search/src/TopicSearch.scss
@@ -38,6 +38,7 @@
 	max-width: none;
 	color: oColorsGetPaletteColor('white');
 	background: transparent;
+	height: 40px;
 
 	&::placeholder {
 		color: oColorsGetPaletteColor('white');


components/x-topic-search/src/TopicSearch.scss
```diff
@@ -38,6 +38,7 @@
 	max-width: none;
 	color: oColorsGetPaletteColor('white');
 	background: transparent;
+	height: 40px;
```
benbarnett: Is it possible to set this instead with `line-height` (and possibly `padding`)? For accessibility, if users change their font size, it could mean that his hardcoded value may not be enough for larger fonts.

Using `line-height` should scale with the font changing.

PR #246, opened 2019-02-04
Fix wrong input height on Safari

fenglish:
I accidentally deleted the branch and closed this PR... https://github.com/Financial-Times/x-dash/pull/245 🙈 

**The bug**
-----
Only on Safari, the input box's height is set differently.
Chrome, Firefox, IE11 => 40px
Safari => 46px
wrong-top-position

**The reason**
-----

It seems Safari includes the height of search cancel button as the input height and Chrome doesn't.

css-difference

```css
input {
    min-height: 40px;
    padding: 9px 9px 9px;
    font-size: 16px;
    line-height: 20px;
    margin: 0;
    border-bottom: 2px solid #ffffff;
    padding-left: 24px;

    &::-webkit-search-cancel-button {
	@include oIconsGetIcon('cross', oColorsGetPaletteColor('white'), 26);
	-webkit-appearance: none;
    }
}
```

**The solution**
-----
Adding `position: absolute` or `z-index` doesn't solve the problem. Overwrite padding top/bottom 0 and keep the input height the min-height(40px) no matter whether the browser includes the cancel button height as the input height or not.

### DIFF
--- a/components/x-topic-search/src/TopicSearch.scss
+++ b/components/x-topic-search/src/TopicSearch.scss
@@ -34,7 +34,7 @@
 	margin: 0;
 	border: none;
 	border-bottom: 2px solid oColorsGetPaletteColor('white');
-	padding-left: 24px;
+	padding: 0 9px 0 24px;
 	max-width: none;
 	color: oColorsGetPaletteColor('white');
 	background: transparent;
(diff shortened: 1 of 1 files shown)


i-like-robots:
Just a note on something I noticed, I don't expect it to be addressed in this PR... I see that this component is using some styles from o-forms, but I can't see that the design has much commonality with o-forms and that many of the styles are being overwritten. I'd be tempted to try `.input { font: inherit; }` as the base and go from there.

fenglish:
@i-like-robots Thank you for your comment! I will try to get rid of oForm mixins by new PR.

PR #249, opened 2019-02-05
Select the text when input box is changed from blur to focus

fenglish:
**on Blur**
screenshot 2019-02-05 at 14 11 14

⏬ ⏬ ⏬ 

**on Focus**
screenshot 2019-02-05 at 14 11 24

### DIFF
--- a/components/x-topic-search/__tests__/x-topic-search.test.jsx
+++ b/components/x-topic-search/__tests__/x-topic-search.test.jsx
@@ -43,6 +43,18 @@ describe('x-topic-search', () => {
 		});
 	});
 
+	describe('when input receives focus', () => {
+		it('selects the text in the input', () => {
+			const selectMock = jest.fn();
+			const inputBox = target.find('input');
+
+			inputBox.simulate('blur');
+			inputBox.simulate('focus', { target: { select: selectMock }});
+
+			expect(selectMock).toHaveBeenCalledTimes(1);
+		});
+	});
+
 	describe('given inputted text is shorter than minSearchLength', () => {
 		const apiUrlWithResults = buildSearchUrl('a');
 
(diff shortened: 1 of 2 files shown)


components/x-topic-search/__tests__/x-topic-search.test.jsx
```diff
@@ -43,6 +44,18 @@ describe('x-topic-search', () => {
 		});
 	});
 
+	describe('changed from on blur to on focus', () => {
+		it('should call select on event.target', () => {
```
dan-searle: Maybe the following are better descriptions for `describe` and `it` respectively:

* "input receives focus when it already contains a search term"
* "selects the text in the input"

This describes what should be achieved, rather than the technical steps involved.

components/x-topic-search/__tests__/x-topic-search.test.jsx
```diff
@@ -43,6 +44,18 @@ describe('x-topic-search', () => {
...
+			const selectMock = sinon.spy();
```
dan-searle: Can you use Jest for the spy, so we're not introducing another dependency?

https://jestjs.io/docs/en/jest-object

components/x-topic-search/src/TopicSearch.jsx
```diff
@@ -71,14 +72,23 @@ class TopicSearch extends Component {
 		}
 	}
 
-	handleInputClickOrFocus() {
+	checkSearchTermLength() {
```
dan-searle: I would say this method is more about showing the result than checking the searchTerm length, but in any case, I think the following would be simpler, and avoids an additional class method:

```es6
	handleInputClick() {
		if (this.state.searchTerm.length >= this.minSearchLength) {
			this.setState({
				showResult: true
			});
		}
	}

	handleInputFocus(event) {
		event.target.select();
		this.handleInputClick();
	}
```

PR #254, opened 2019-02-06
Tidy up for majour release

fenglish:


components/x-topic-search/src/SuggestionList.jsx
```diff
@@ -4,26 +4,26 @@ import styles from './TopicSearch.scss';
...
 export default ({ suggestions, searchTerm, csrfToken, followedTopicIds = [] }) => (
-	<ul className={ classNames(styles["suggestions"]) } aria-live="polite">
+	<ul className={classNames(styles["suggestions"])} aria-live="polite">
 
-		{ suggestions.map((suggestion, index) => (
-			<li className={ classNames(styles["suggestion"]) }
-				key={ index }
+		{ suggestions.map(suggestion => (
+			<li className={classNames(styles["suggestion"])}
+				key={suggestion.id}
```
dan-searle: Could spot! 👍

components/x-topic-search/src/SuggestionList.jsx
```diff
@@ -4,26 +4,26 @@ import styles from './TopicSearch.scss';
 import classNames from 'classnames';
 
 export default ({ suggestions, searchTerm, csrfToken, followedTopicIds = [] }) => (
-	<ul className={ classNames(styles["suggestions"]) } aria-live="polite">
+	<ul className={classNames(styles["suggestions"])} aria-live="polite">
 
-		{ suggestions.map((suggestion, index) => (
-			<li className={ classNames(styles["suggestion"]) }
-				key={ index }
+		{ suggestions.map(suggestion => (
```
dan-searle: Maybe remove the space after `{` here too.

fenglish: oh! I missed them! Thank you!

components/x-topic-search/src/SuggestionList.jsx
fenglish: 😊

components/x-topic-search/readme.md
```diff
@@ -1,6 +1,6 @@
 # x-topic-search
 
-This module provides a topic search bar.
+This module provides a topic search bar. It displays topic suggestions which are fetched from the api, passed as a property, based on user input(search term). x-follow-button is displayed with the topic suggested.
```
benbarnett: Could we link/reference the API it uses?

benbarnett: Also might be worth rewording the last sentence to suggest there are multiple topics:

```suggestion
This module provides a topic search bar. It displays topic suggestions which are fetched from the api, passed as a property, based on user input(search term). `x-follow-button`s are used for each of the suggestions to allow users to follow directly
```

dan-searle: Maybe this should be more focussed on the user and less on the technical detail of how it works?

>This module allows a user to search for topics by name, and follow them. If the an already-followed topic is return in the search results, then those topics are indicated as such.

benbarnett: Agree with this actually. Might still be worth mentioning where it gets its suggestions from though.

fenglish: Thank you, @benbarnett and @dan-searle!

PR #255, opened 2019-02-15
Update data for podcast teaser

debugwand:
This reflects the change we made to next-es-interface to use a different image for podcasts when in a teaser


i-like-robots:
Thanks @debugwand 🙇

PR #256, opened 2019-02-21
Update o-labels dependency and HTML template

liamkeaton:
To update `o-teaser`, `o-labels` has to be updated. There was a breaking change where label class names were changed so the premium label in `x-teaser` was changed from `.premium` to `.content-premium`

### DIFF
--- a/components/x-teaser/src/Title.jsx
+++ b/components/x-teaser/src/Title.jsx
@@ -4,6 +4,8 @@ import Link from './Link';
 export default ({ title, altTitle, headlineTesting, relativeUrl, url, indicators, ...props }) => {
 	const displayTitle = headlineTesting && altTitle ? altTitle : title;
 	const displayUrl = relativeUrl || url;
+	// o-labels--premium left for backwards compatibility for o-labels v3
+	const premiumClass = 'o-labels o-labels--premium o-labels--content-premium';
 
 	return (
 		<div className="o-teaser__heading">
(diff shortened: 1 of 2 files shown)


components/x-teaser/src/Title.jsx
```diff
@@ -15,7 +15,7 @@ export default ({ title, altTitle, headlineTesting, relativeUrl, url, indicators
 				{' '}
 			</Link>
 			{indicators && indicators.accessLevel === 'premium' ? (
-				<span className="o-labels o-labels--premium" aria-label="Premium content">
+				<span className="o-labels o-labels--content-premium" aria-label="Premium content">
```
i-like-robots: Do you think we should keep the old class name as well to smooth over the transition for apps which haven't had o-labels bumped, or not do that and bump apps as and when we see display issues?

liamkeaton: Good idea, I think having both in won't cause issues apart from slightly messy HTML. Otherwise we could make this a breaking change, which might be a bit extreme.

i-like-robots: We haven't released v1 yet so breaking slightly is OK IMO.

liamkeaton: Ok, shall I remove the messy HTML then?

i-like-robots: I think this is good and it's commented so 👍

liamkeaton:
Any ETA on a release, want to bump the `next-stream-page` and close the ops cops issue?

i-like-robots:
Apologies @liamkeaton, I've released v1.0.0-beta.8 now!

PR #259, opened 2019-02-28
Added onClick prop to X-Follow-Button

pavelpichrt:
Adds an `onClick` prop to X-Follow-Button, looks like we don't have `prop-types`, so `if` will have to do

### DIFF
--- a/components/x-follow-button/src/FollowButton.jsx
+++ b/components/x-follow-button/src/FollowButton.jsx
@@ -10,7 +10,7 @@ export const FollowButton = (props) => {
 		isFollowed,
 		csrfToken,
 		followPlusDigestEmail,
-		onClick,
+		onSubmit,
 		variant
 	} = props;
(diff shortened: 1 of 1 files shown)


components/x-follow-button/src/FollowButton.jsx
```diff
@@ -51,6 +52,10 @@ export const FollowButton = (props) => {
 					token: csrfToken
 				};
 
+				if (typeof onClick === 'function') {
+					onClick(detail);
```
dan-searle: I do think it's a little odd that this is an `onClick` prop but it's called `onSubmit`. It would be more logical to either move it onto the actual button, or keep it where it is but rename it to `onSubmit`?

Having said that, in this case it probably doesn't make much difference practically.

pavelpichrt: updated

Issue #263, opened 2019-03-05
Rebranding for 2019


i-like-robots:
We are still using the original description of the project from when it was first started:

> x-dash is an experiment in building new shared front-end components for FT.com and the FT Apps. A detailed introduction is available in Google Slides.

Now that x-dash is being used in production and is no longer flagged as experimental (#261) we should revisit this using the information and perspective we now have about how to use x-dash effectively.

I think an updated introduction to x-dash should answer the question "what is this?" as well as "why should I use this?" and be crafted to suit both a technical and non-technical audience.

PR #269, opened 2019-04-17
x-audio

fenglish:
🚧 WIP 🚧 
Audio player

audio-player-2


components/x-audio/stories/knobs.js
```diff
@@ -1,6 +1,10 @@
 module.exports = (data, { boolean }) => ({
 	playing: boolean('Playing', data.playing),
 	loading: boolean('Loading', data.loading),
-	onPlay: () => console.log('Pressed play'),
-	onPause: () => console.log('Pressed pause')
```
benbarnett: Could you see what eslint rule this was? Not sure what's wrong with it!

benbarnett: Oh, `console` maybe?

components/x-audio/package-lock.json
```diff
@@ -0,0 +1,6798 @@
+{
```
quarterto: `package-lock.json` shouldn't be committed in components, see #158. I think we should be able to prevent this using a top-level `npmrc` file

quarterto: Oh, this should already be happening #200 🤦‍♀

components/x-audio/bower.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "x-audio",
+  "main": "dist/Audio.es5.js",
```
i-like-robots: This could be the ES6 bundle as we always transform the contents of Bower components (Origami components are ES6)

components/x-audio/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "x-audio"
+  ],
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "@financial-times/o-audio": "1.0.0-beta.11",
+    "@financial-times/x-engine": "file:../../packages/x-engine",
+    "classnames": "^2.2.6",
+    "hammerjs": "^2.0.8",
+    "prop-types": "^15.7.2",
```
i-like-robots: Should this be a peer dependency?

i-like-robots: I think this may also require that we add some extra configuration to our build step to ensure they are removed in production?

components/x-audio/readme.md
```diff
@@ -0,0 +1,101 @@
+# x-audio
+
+This module can be used to play an audio file via pluggable engines/players.
+
+It comes bundled with a web audio player (via [HTMLAudioElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLAudioElement)), but can be used with another engine such as in the native layer. The bundled engine uses a [Redux](https://redux.js.org/) to manage state.
```
i-like-robots: > use a Redux to manage state

-> "uses Redux to manage state"

i-like-robots:
This is too big for me to do an in-depth review of, but it looks good! It appears to be documented and tested well and I trust that you had a good feedback loop in the team.

However, this is not on FT.com and I'm not sure if there are any plans to make it so which means it is not technically an `x-` component.

As discussed IRL I don't think that necessarily has to be a problem (cc #263) but some reassurance that the component has been tested and _would work_ on FT.com in its current form would be a bonus - has it been tried?

debugwand:
we've not as there would be work to get it all set up on dot com, so whoever comes back to it would need to do that. for example, the slider for the timeline is set up for touch devices and would need mouse support added in.

edds:

rowanbeentje:
This component was never added to ft.com, and because we needed tighter integration within the app*, we [have imported and adopted it as a player component within ft-app](https://github.com/Financial-Times/ft-app/tree/main/lib/components/audio/player). So I'm going to close this PR :)

(* Note this would have been perfectly possibly with x-interaction etc, but we could simplify the implementation if we knew it was app-only!)

PR #275, opened 2019-04-24
Move Storybook to the repository root

i-like-robots:
This PR moves Storybook from a subdirectory (`tools/x-storybook`) up to the repository root.

This enables several useful features as discussed in #177:

- Components no longer need to be specified as dependencies of Storybook
- Storybook will be able to watch components for changes

_Please note_ that this PR does not make it possible to directly include a component's source code into a story as this will require changes to JSX transpilation. I will do this in another PR.

_Please also note_ that this implements Storybook v5 rather than the previous v4. Other than changes to configuration the "options" add-on has been deprecated.

To rebase onto this branch you will need to:

- Update `.storybook/register-components.js`, please note that you should now use a path specifier rather than a bare module reference.
- Remove any changes made to `x-storybook/package.json`


quarterto:
I've removed the `x-storybook` project from Snyk to get it passing

Issue #274, opened 2019-04-24
Features to speed development


debugwand:
Right now, if I make changes to a component I have to stop storybook, rebuild with athloi and then restart storybook (`athloi run build --filter 'keywords:"x-audio"' && npm run start-storybook`)

This could be speeded up by adding watch and hot reloading

quarterto:
Yeah, it's a good experience right now. It used to be better, and I don't know what happened (I think something we lost in the move to Athloi?).

FWIW, Storybook does hot reload, but the components don't build. Some components (e.g. x-teaser) have a `watch` or `start` script, but you have to run it separately, manually, and it's not done in a standardised way.

i-like-robots:
Unfortunately this is caused by Storybook not being installed at the repository root, details in #203.

We "fixed" this in Anvil and plan to port what we learned to x-dash in the near future.

quarterto:
See also #198. At one point we had watching/hot reloading working without toplevel Storybook by having a single command to start Storybook and the watch builds of every component; I'm not sure when we lost that. I tried toplevel Storybook in #203 and it worked, but slowly, using `require.context`. I see you've taken a different approach in Anvil.

Now that the docs are using Storybook's generated `iframe.html` files (#155) I think we can step away from the abstract story format which was only there so Gatsby could generically consume stories without using Storybook's API.

i-like-robots:
Yes, I definitely agree. The details are lost to me now, but as you say we are no longer attempting to do what we originally tried to do with Storybook.

Moving away from the custom config sounds like a good plan.

i-like-robots:
I've made some further changes in #336 which enable component stories to be authored in the regular manner documented by Storybook and import components directly from source.

PR #277, opened 2019-04-25
make snapshot tests explicit in each component

### LINKED ISSUE #152: Create separate snapshot files for each component
Because all component shapshots are currently pushed into a single file I am worried that any changes to components not currently under test may not be caught as I suspect people may only look at the part of the diff related to their component.

### LINKED ISSUE #196: switch jest snapshots to use storyshots
https://github.com/storybooks/storybook/tree/release/3.4/addons/storyshots

pros:

- less custom x-dash magic is good

cons:

- i do not trust storybook

quarterto:
closes #196, fixes #152. we'll need to port this to the new components at some point; i think it'll be easier to wait until they're all merged before merging this to save on churn


quarterto:
i'm actually going to do #196 instead. this will allow us to directly use `storiesOf` and hopefully enable #274

PR #278, opened 2019-04-25
Hook up Play + Pause actions to the app

robsquires:
Wire up x-audio so it can be manipulated by its consumers.

In this case the playing state will be managed by ft-app redux: https://github.com/Financial-Times/ft-app/pull/248


robsquires:
LGTM

PR #279, opened 2019-04-26
X audio grid layout & basic styles

fenglish:
**Expanded**
Screenshot 2019-04-26 at 16 01 18

**Minimised**
Screenshot 2019-04-26 at 16 01 27

**Dummy**
- minimise button
- play icon 
- range bar for timeline


components/x-audio/src/components/styles.scss
```diff
@@ -1,31 +1,220 @@
+@import 'o-icons/main';
```
benbarnett: Is this file uses spaces over tabs?

PR #280, opened 2019-04-26
X-audio component

debugwand:
x-audio load component

* shows under the various loading events, waiting and stalled
* removed when canplay event fires
* dark when expanded mode, light when minimised
* expecting to inherit colours and background colours from containers

also:
* bower added so we can use typography mixins
* `x-audio` added as a keyword in package json so we can build locally with `athloi run build --filter 'keywords:"x-audio"'` instead of building everything
* `x-test-utils` and our first test file

### DIFF
--- a/components/x-audio/.bowerrc
+++ b/components/x-audio/.bowerrc
@@ -1,0 +1,8 @@
+{
+	"registry": {
+		"search": [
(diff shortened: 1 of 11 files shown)


benbarnett:
Closing as this has been rebased since https://github.com/Financial-Times/x-dash/pull/279

PR #286, opened 2019-05-08
x-audio loading component

benbarnett:
***this PR supersedes https://github.com/Financial-Times/x-dash/pull/280 (I have done on a separate branch as it was written by @debugwand, I wanted to make sure the original changes were still in tact)***

x-audio load component

* shows under the various loading events, waiting and stalled
* removed when canplay event fires
* dark when expanded mode, light when minimised
* expecting to inherit colours and background colours from containers

also:
* bower added so we can use typography mixins
* `x-audio` added as a keyword in package json so we can build locally with `athloi run build --filter 'keywords:"x-audio"'` instead of building everything
* `x-test-utils` and our first test file

### DIFF
--- a/components/x-audio/src/components/styles.scss
+++ b/components/x-audio/src/components/styles.scss
@@ -1,6 +1,9 @@
 @import 'o-icons/main';
 @import 'o-colors/main';
+
+$o-typography-load-fonts: false;
 @import 'o-typography/main';
+@include oTypography();
 
 $height--minimised: 48px;
 $info-height--expanded: 72px;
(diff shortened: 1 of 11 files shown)


components/x-audio/src/components/styles.scss
```diff
@@ -218,3 +223,20 @@ $info-height--expanded: 72px;
 	align-self: center;
 	margin-left: 8px;
 }
+
+.audio-player__status {
+  grid-area: status;
```
fenglish: @benbarnett this added part using spaces like my PR earlier.
I think the original scss file was using spaces instead of tabs.

benbarnett: Well spotted, thanks, will update

PR #288, opened 2019-05-08
expose source files in packages

quarterto:
from conversation with @pavelpichrt: Apps Team would like to be able to use their own webpack configuration for x-dash components, to interoperate with their CSS modules configuration and other tooling. Previously, x-teaser'ds source files were being excluded by the npmignore, preventing this.

This PR also adds the `source` field to every component, allowing tooling to pick up on it and do e.g. for webpack:

```js
{
  resolve: {mainFields: ['source', 'browser', 'module', 'main']}}
}
```

allowing `import {Teaser} from '@financial-times/x-teaser'` to load the source JSX

### DIFF
--- a/components/x-increment/package.json
+++ b/components/x-increment/package.json
@@ -3,6 +3,7 @@
   "version": "0.0.0",
   "private": true,
   "description": "",
+  "source": "src/Increment.jsx",
   "main": "dist/Increment.cjs.js",
   "module": "dist/Increment.esm.js",
   "browser": "dist/Increment.es5.js",
--- a/components/x-interaction/package.json
+++ b/components/x-interaction/package.json
@@ -2,6 +2,7 @@
   "name": "@financial-times/x-interaction",
   "version": "0.0.0",
   "description": "This module enables you to write x-dash components that respond to events and change their own data.",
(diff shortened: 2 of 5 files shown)


i-like-robots:
To clarify... "untranspiled" may refer to anything, including compile-to-JS code but what we're talking about is something where the JS is directly consumable but the CSS has not yet been processed and the original link between them remains, is that along the right lines?

quarterto:
hmm, good point about CSS, i don't know how we'd do that (would it be possible to mark those as externals? even then, it'd be exposing some internal module structure as part of the external API)

(editor's note: there's some timey wimey stuff going on in the replies in this thread)

pavelpichrt:
I did a bit of research and testing around rollup and it looks like a "raw" bundle is a really bad idea. Rollup is clearly intended to deliver a ready to use bundle in a format which a client can directly use. The problem is that even the existing ES bundles of our components generated by rollup will almost always be rebundled by the client app (and transpiled again) creating redundant code in the process.

I'm not a rollup expert though, so let's pick it up tomorrow morning in person because I may be missing something. @i-like-robots

PR #289, opened 2019-05-08
X audio remaining time

fenglish:
This PR includes
- Remove css grid for info section
- Display Loading state in correct place for minimised and expanded player
- Display Time remaining for minimised and expanded player
- Display Current time for expanded player
- Add updateCurrentTime logic in player logic

loadig-and-remaining-expanded

loadig-and-remaining-minimised


components/x-audio/src/components/format-seconds-to-hhmmss.js
```diff
@@ -0,0 +1,16 @@
...
+  const displayHour = hours < 10 ? `0${hours}` : hours;
+  const displayMin = minutes < 10 ? `0${minutes}` : minutes;
+  const displaySec = seconds < 10 ? `0${seconds}` : seconds;
+
+  if (hours > 0) {
+    return `${displayHour}:${displayMin}:${displaySec}`;
+  } else {
+    return `${displayMin}:${displaySec}`;
+  }
+}
```
benbarnett: I think indentation is spaces here

components/x-audio/src/components/TimeRemaining.jsx
```diff
@@ -0,0 +1,30 @@
+import { h } from '@financial-times/x-engine';
+import * as PropTypes from 'prop-types';
+import classNameMap from './classnames-helper';
+import formatToHHMMSS from './format-seconds-to-hhmmss';
+
+const formatToMinsRemaining = (targetSeconds) => {
+	const minutes = Math.floor((targetSeconds / 60))
```
benbarnett: missing `;`

benbarnett: Just thinking about `Math.floor` here... 

I guess for the last 60 seconds it will always say "0 minutes remaining", which could feel like a long time? I.e. when I see "0 minutes remaining" I would expect it to end very soon.

Would `Math.round` work better? So it would only flip to 0 when less than 30 seconds?

components/x-audio/src/components/index.jsx
```diff
@@ -15,25 +17,24 @@ export const Audio = ({
...
+	duration
 }) => (
 	<div className={classNameMap('audio-player', `audio-player--${expanded ? 'expanded' : 'minimised'}`)}>
 		{expanded && <button className={classNameMap('audio-player__minimise-button')} title='minimize player'/>}
 		{expanded && <div className={classNameMap('audio-player__control-timeline')}><input style={{width: '100%'}} type='range'/></div>}
 		{expanded && <button className={classNameMap('audio-player__rewind')} title='rewind 30 seconds'/>}
 		{expanded && <button className={classNameMap('audio-player__forward')} title='forward 30 seconds'/>}
 		{expanded && <button className={classNameMap('audio-player__control-speed')} title='change play speed'>x1</button>}
-
+		{expanded && !loading && <div className={classNameMap('audio-player__info__current-time')}>{formatTime(currentTime)}</div>}
```
benbarnett: Perhaps not for this PR but as the source order is being controlled by CSS Grid, we could arrange these lines into groups to potentially make them more readable.

Might it be easier to read by grouping the elements within the `expanded` block? e.g.

```jsx
{expanded && (
  <button className={classNameMap('audio-player__minimise-button')} title='minimize player'/>
  <div className={classNameMap('audio-player__control-timeline')}><input style={{width: '100%'}} type='range'/></div>
)
```

components/x-audio/src/redux/__tests__/player-logic.test.js
```diff
@@ -0,0 +1,139 @@
+import { middleware, actions, initialState, reducer } from '../player-logic';
+const runActions = (initialState, ...actions) => actions.reduce(reducer, initialState);
+
+describe('actions and reducer', () => {
+
+  test('Actions not relevant to audio have no effect on state', () => {
```
benbarnett: Think this file is also using spaces not tabs

components/x-audio/src/redux/player-logic.js
```diff
@@ -11,15 +12,14 @@ export function reducer (state = initialState, action) {
...
-		case 'LOADING': {
-			return {...state, loading: true}
-		}
-		case 'LOADED': {
-			return {...state, loading: false}
-		}
+		case 'LOADING':
+			return {...state, loading: true }
+		case 'LOADED':
+			return {...state, loading: false }
```
benbarnett: Missing semicolons after all of these cases

fenglish:
@benbarnett Thank you for reviewing this PR! I have added a commit after your review.
Could you have a look at it as well please?
The commit is about 
 - Change hh:mm:ss format to h:mm:ss
 - Add tests for format seconds

https://github.com/Financial-Times/x-dash/pull/289/commits/8a0d5806cae93d9429710c5a33c448851851a8f1

fenglish:
@benbarnett Thank you for reviewing this PR!
I have added a commit after your review(I haven't reflected your reviews yet).
Could you have a look as well please?

The commit https://github.com/Financial-Times/x-dash/pull/289/commits/e0460fc6ec8873d5d66f38fec6dc0a5ecc096671 is about 
 - Change hh:mm:ss format to h:mm:ss
 - Add tests for format seconds

components/x-audio/src/components/index.jsx
fenglish: @benbarnett Is this possible to do?
I got this error when I try to do this way. Am I missing something?

```
Adjacent JSX elements must be wrapped in an enclosing tag. Did you want a JSX fragment <>...</>?
```

components/x-audio/src/components/TimeRemaining.jsx
fenglish: I will tackle this by another PR after asking designer Luke whether it need to be change into seconds when the remaining is less than minute or not.

PR #291, opened 2019-05-10
Modify babel config to allow to module config for tests

fenglish:
When running jest against files in the src directory, it was failing because modules import syntax wasn't being transpiled to commonJS format. This change allows the modules property to be overwritten and does for the Jest config.


packages/x-babel-config/index.js
```diff
@@ -1,4 +1,4 @@
-module.exports = (targets = []) => ({
+module.exports = (targets = [], modules = false) => ({
```
quarterto: when calling this function it isn't clear what the arguments are for. i'd suggest using an object:

```suggestion
module.exports = ({ targets = [], modules = false } = {}) => ({
```

packages/x-babel-config/jest.js
```diff
@@ -1,4 +1,6 @@
 const getBabelConfig = require('./');
 const babelJest = require('babel-jest');
 
-module.exports = babelJest.createTransformer(getBabelConfig());
+const base = getBabelConfig([{ node: 6 }], 'commonjs');
```
quarterto: with the change to an object, this would be:

```suggestion
const base = getBabelConfig({
	targets: [{ node: 'current' }],
	modules: 'commonjs'
});
```

you can also use `current` as the target node version, so `babel-preset-env` only transforms what it needs for the version of node it's running in.

fenglish:
@quarterto Thank you for your suggestion!

quarterto:
@fenglish it will need updating in https://github.com/Financial-Times/x-dash/blob/modify-babel-config/packages/x-rollup/src/rollup-config.js as well 🙂

packages/x-rollup/src/rollup-config.js
```diff
@@ -25,7 +25,12 @@ module.exports = ({ input, pkg }) => {
 			{
 				input,
 				external,
-				plugins: [babel(babelConfig({ node: 6 })), ...plugins]
+				plugins: [
+					babel(babelConfig({
+						target: [{ node: 6 }]
```
quarterto:
```suggestion
						targets: [{ node: 6 }]
```

packages/x-rollup/src/rollup-config.js
```diff
@@ -36,7 +41,12 @@ module.exports = ({ input, pkg }) => {
...
```
quarterto:
```suggestion
						targets: [{ node: 6 }]
```

packages/x-rollup/src/rollup-config.js
```diff
@@ -47,7 +57,12 @@ module.exports = ({ input, pkg }) => {
 			{
 				input,
 				external,
-				plugins: [babel(babelConfig({ browsers: ['ie 11'] })), ...plugins]
+				plugins: [
+					babel(babelConfig({
+						target: [{ browsers: ['ie 11'] }]
```
quarterto:
```suggestion
						targets: [{ browsers: ['ie 11'] }]
```

fenglish:
@quarterto Thank you! I'm sorry for my careless mistakes 😞

quarterto:
don't worry about it 🙂 mistakes happen to everyone

i-like-robots:
As this was not merged in...

Would it be preferable to add Babel to our Jest setup and include our source files in the tests? This would skip the build step for running specs and enable simpler use of Jest watch.

debugwand:
@i-like-robots from what I can gather we're doing a fairly standard way of applying babel to jest (https://jestjs.io/docs/en/getting-started).

PR #293, opened 2019-05-13
Add an `ended` notifier so consumers can respond to this event

benbarnett:
Adds an `ended` notifier that can be used by the consumer.

For example, the app will use this to close the player when audio has finished playing in the minimised state.

### DIFF
--- a/components/x-audio/src/redux/__tests__/player-logic.test.js
+++ b/components/x-audio/src/redux/__tests__/player-logic.test.js
@@ -33,6 +33,24 @@ describe('actions and reducer', () => {
 		expect(updatedState).toMatchSnapshot();
 	});
 
+	test('Ended action sets ended to false', () => {
+		const updatedState = runActions(initialState, actions.play(), actions.ended());
+
+		expect(updatedState).toMatchSnapshot();
+	});
+
+	test('Playing audio after previous has ended resets ended state', () => {
+		const updatedState = runActions(initialState, actions.ended(), actions.play());
+
+		expect(updatedState).toMatchSnapshot();
+	});
+
+	test('Requesting new audio play resets ended state', () => {
+		const updatedState = runActions(initialState, actions.ended(), actions.requestPlay());
+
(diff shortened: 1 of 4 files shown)


benbarnett:
Going to re-work this so it follows the notifier model

components/x-audio/src/redux/__tests__/player-logic.test.js
```diff
@@ -33,6 +33,24 @@ describe('actions and reducer', () => {
 		expect(updatedState).toMatchSnapshot();
 	});
 
+	test('Ended action sets ended to false', () => {
```
fenglish: @benbarnett Sorry for the late review... I know this PR is already merged. This text is incorrect?
'Ended action sets ended to false' => 'Ended action sets ended to **true**'

components/x-audio/src/redux/__tests__/player-logic.test.js
```diff
@@ -33,6 +33,24 @@ describe('actions and reducer', () => {
...
+		const updatedState = runActions(initialState, actions.play(), actions.ended());
+
+		expect(updatedState).toMatchSnapshot();
+	});
+
+	test('Playing audio after previous has ended resets ended state', () => {
```
fenglish: @benbarnett I'm not sure which part this should be done. I'm sorry if this is the wrong place to mention.
We need to reset `currentTime` as well?

benbarnett: I think we're ok for the `currentTime` as that bit is controlled by the audio element itself.... i.e as soon as the audio starts playing again, it will fire a new event, which will take control of the `currentTime`.

I couldn't see any side effects in testing, but this is worth double checking. Thanks.

components/x-audio/src/redux/__tests__/player-logic.test.js
benbarnett: Whoops! Yes you're right, thanks. Fixed in e436af48

PR #295, opened 2019-05-13
Add aria-labels to buttons

debugwand:
* switched `title` for `aria-label` because `aria-label` doesn't have the issues `title` has
* added `aria-label` to the close button


benbarnett:
hey @debugwand do you have any more details on what this means? Just curious!
 
> because aria-label doesn't have the issues title has

benbarnett:
Ah, seems to avoid potential duplication according to this: https://www.deque.com/blog/text-links-practices-screen-readers/

PR #298, opened 2019-05-15
loadMedia action

robsquires:
- Required to configure tracking, so that when new audio is played, tracking can be reconfigured
- Required for storybook (+ likely .com) so that when the page loads the URL can be set BUT the audio won't start playing

### DIFF
--- a/components/x-audio/src/redux/__tests__/player-logic.test.js
+++ b/components/x-audio/src/redux/__tests__/player-logic.test.js
@@ -59,22 +59,16 @@ describe('middleware', () => {
 		return { store, next, invoke, audio };
 	}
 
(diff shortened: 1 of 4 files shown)


components/x-audio/src/redux/__tests__/player-logic.test.js
```diff
@@ -59,22 +59,16 @@ describe('middleware', () => {
 		return { store, next, invoke, audio };
 	}
 
-	test('requestPlay with url sets src and plays', () => {
+	test('loadMedia set the URL and configures tracking', () => {
```
benbarnett: Pedant comment but worth just changing the name of this test in the tracking work?

PR #299, opened 2019-05-15
Get duration by durationchange event

fenglish:


components/x-audio/src/redux/player-logic.js
```diff
@@ -95,6 +102,11 @@ export const middleware = (store, audio = new Audio()) => {
 	audio.addEventListener('loadeddata', () => store.dispatch(actions.loading()));
 	audio.addEventListener('canplay', () => store.dispatch(actions.loaded()));
 
+	audio.addEventListener('durationchange', () => {
+		const duration = Math.floor(audio.duration);
```
benbarnett: Just trying to think if `Math.floor` is the right one here.... 🤔 

So if an audio file is 13.7 seconds long, we would then internally recognise that as 13 seconds. I suppose that technically means you couldn't scrub to the absolute end of the audio. But that's probably not a problem.

What do you think?

fenglish: @benbarnett Yeah, you are right! the `Math.floor` doesn't need there. I'm going to delete it. Thank you!

PR #300, opened 2019-05-15
O audio integration

robsquires:
So we maintain the same audio tracking events as the current audio player has, we need to hook up the o-audio library.

This adds a new prop `trackingContext` which looks like:
```
trackingContext: PropTypes.shape({
  contentId: PropTypes.string,
  playerType: PropTypes.string,
  audioSubtype: PropTypes.string
}).isRequired
```

The [Tracking](https://github.com/Financial-Times/o-audio/blob/master/src/js/tracking.js) class within o-audio is being pulled in explicitly.

### DIFF
--- a/components/x-audio/src/redux/index.jsx
+++ b/components/x-audio/src/redux/index.jsx
@@ -16,13 +16,15 @@ export default function connectPlayer (Player) {
 	const playerActions = wrapWithDispatch(store, {
 		onPlayClick: actions.requestPlay,
 		onPauseClick: actions.requestPause,
-		loadMedia: actions.loadMedia
+		loadMedia: actions.loadMedia,
+		willClose: actions.willClose
 	});
 
 	class ConnectedPlayer extends Component {
 		constructor(props) {
 			super(props);
 			this.unsubscribe = store.subscribe(this.storeUpdated.bind(this));
+			this.onCloseClick = this.onCloseClick.bind(this);
 			this.state = initialState;
 		}
 
@@ -37,8 +39,12 @@ export default function connectPlayer (Player) {
 		}
 
 		componentWillUnmount() {
-			playerActions.onPauseClick();
 			this.unsubscribe();
+		}
+
+		onCloseClick() {
+			playerActions.willClose();
+			this.props.onCloseClick();
 		}
 
 		storeUpdated() {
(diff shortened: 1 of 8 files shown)


components/x-audio/src/redux/__tests__/__snapshots__/player-logic.test.js.snap
```diff
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`actions and reducer Ended action sets ended to false 1`] = `
+exports[`actions and reducer Ended action sets ended to true 1`] = `
```
benbarnett: 🤦‍♂  sorry this was me

components/x-audio/src/redux/index.jsx
```diff
@@ -16,31 +16,32 @@ export default function connectPlayer (Player) {
...
+			playerActions.loadMedia({ url, trackingContext, autoplay: playing });
 		}
 
 		componentWillUnmount() {
-			playerActions.onPauseClick();
 			this.unsubscribe();
 		}
 
+		onCloseClick() {
+			playerActions.willClose();
```
benbarnett: Could this go in `componentWillUnmount` maybe? Then no need to overload `onCloseClick`?

components/x-audio/src/redux/player-logic.js
```diff
@@ -91,7 +96,10 @@ export const middleware = (store, audio = new Audio()) => {
 	audio.addEventListener('waiting', () => store.dispatch(actions.loading()));
 	audio.addEventListener('stalled', () => store.dispatch(actions.loading()));
 	audio.addEventListener('loadstart', () => store.dispatch(actions.loading()));
-	audio.addEventListener('loadedmetadata', () => store.dispatch(actions.loading()));
+	audio.addEventListener('loadedmetadata', () => {
```
benbarnett: Will the tracking be set up in time for autoplays?

i.e. `play()` is run straight away but then presumably `loadedmetadata` could fire after that? Or does it always need metadata before it can play? 

Soz if this makes no sense

robsquires: yeah I had a look at that, `loadedmetadata` is dispatched first. I haven't check what happens if the mp3 has been cached - ie the user _plays_, _closes_ and _plays_ the same . podcast again. Will check now

PR #303, opened 2019-05-16
X audio error

debugwand:
* add handling for error event that will show a message to users
* allow users to recover from a network failure resulting in an error that reloads the audio and starts it from where they left off
* also reset all the state when load media occurs

### DIFF
--- a/components/x-audio/src/components/index.jsx
+++ b/components/x-audio/src/components/index.jsx
@@ -2,6 +2,7 @@ import { h } from '@financial-times/x-engine';
 import * as PropTypes from 'prop-types';
 import classNameMap from './classnames-helper';
 import Loading from './Loading';
+import ErrorMessage from './ErrorMessage';
 import {
 	Close,
 	PlayPause
@@ -11,6 +12,7 @@ import formatTime from './format-seconds-to-hmmss';
 
 export const Audio = ({
 	loading,
+	error,
 	expanded,
 	playing,
 	onPlayClick,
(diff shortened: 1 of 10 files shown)


components/x-audio/src/components/Loading.jsx
```diff
@@ -6,7 +6,7 @@ export default ({expanded}) => {
 	return (
 		<div role="status" aria-live="polite" className={classNameMap('audio-player__status')}>
 			<div className={`o-loading o-loading--${foregroundColor} o-loading--mini ${classNameMap('audio-player__loader')}`}></div>
-			Loading
+			LOREM
```
benbarnett: Accidental commit?

components/x-audio/src/components/index.jsx
```diff
@@ -32,8 +34,9 @@ export const Audio = ({
 		{expanded && <div className={classNameMap('audio-player__info__image')}><img alt="dummy"/></div>}
 		<div className={classNameMap('audio-player__info__title')}>{title}</div>
 		<div className={classNameMap('audio-player__info__series-name')}>{expanded ? seriesName : `${seriesName}:`}</div>
-		{!expanded && loading ? null : <TimeRemaining currentTime={currentTime} duration={duration} expanded={expanded}/>}
-		{loading && <Loading expanded={expanded} />}
+		{!expanded && loading || error ? null : <TimeRemaining currentTime={currentTime} duration={duration} expanded={expanded}/>}
+		{!error && loading && <Loading expanded={expanded} />}
+		{error && <ErrorMessage />}
```
benbarnett: This makes me wonder if we should split out the minimised and expanded views into separate components, and just pass them all the props. Perhaps something to consider after the release?

components/x-audio/src/redux/player-logic.js
```diff
@@ -77,18 +85,19 @@ export const middleware = (store, audio = new Audio()) => {
...
+	// 	'loadstart',
+	// 	'progress',
+	// 	'canplay',
+	// 	'canplaythrough',
+	// 	'error'
+	// ].forEach(evtName => {
+	// 	audio.addEventListener(evtName, () => {
+	// 		console.log('audio eventy', evtName);
+	// 	});
+	// });
```
benbarnett: Worth just killing this now?

debugwand: might be useful for when we get onto tracking

components/x-audio/src/components/index.jsx
debugwand: would be easier to understand it

PR #282, opened 2019-05-02
Remove extra space after title text

SamiTriki:
After discussing it with @chee 

An extra space is inserted after the title text on teaser titles, it is apparently serving as a workaround to an issue where the labels weren't displayed correctly. 

This is now causing issue with the "external link" icon on promoted content teasers. Where the icon overlaps the text, and this weird behaviour seems to be linked to this extra space.

I ran x-teaser with premium labels and without the space, it doesn't change anything visually, it seems safe to remove this workaround now

PR #304, opened 2019-05-17
reinstate space just for premium label

Thurst-ft:
Reverts the change made in https://github.com/Financial-Times/x-dash/pull/282 with as slight amend so space is only applied when a premium content label is required.

Fixes part of the bug raised here: https://trello.com/c/CkoIA7pZ/889-padding-issue-with-topic-tracker-label

spacing_on_teaser_title

Following a discussion with Lee about amending this in o-tracker it was recommened that the fix was made in x-teaser: https://financialtimes.slack.com/archives/C02FU5ARJ/p1557917172106700

### DIFF
--- a/components/x-teaser/src/Title.jsx
+++ b/components/x-teaser/src/Title.jsx
@@ -16,8 +16,11 @@ export default ({ title, altTitle, headlineTesting, relativeUrl, url, indicators
 				{displayTitle}
 			</Link>
 			{indicators && indicators.accessLevel === 'premium' ? (
-				<span className={premiumClass} aria-label="Premium content">
-					Premium
+				<span>
+					{' '}
+					<span className={premiumClass} aria-label="Premium content">
+						Premium
+					</span>
 				</span>
 			) : null}
 		</div>


i-like-robots:
As I understand it (I could be wrong!) I think the solution to this is to ensure the app is using an up-to-date version of o-teaser which implements the correct margins between components.

https://github.com/Financial-Times/o-teaser/pull/119

notlee:
Hey, I helped look into this yesterday. Looks like those changes apply to tag elements `o-teaser__tag`, `o-teaser__tag-prefix`, and `o-teaser__tag-suffix` (previously o-teaser__duration) but labels are applied independent of tags.

'o-teaser' demos output [promoted content labels](https://www.ft.com/__origami/service/build/v2/demos/o-teaser@3.5.3/promoted-content?brand=master) in a span and rely on the space. We should probably output premium labels as part of `o-teaser` like we do with promoted content labels `o-teaser__promoted-prefix`, but that's a separate issue.

components/x-teaser/src/Title.jsx
```diff
@@ -16,8 +16,11 @@ export default ({ title, altTitle, headlineTesting, relativeUrl, url, indicators
 				{displayTitle}
 			</Link>
 			{indicators && indicators.accessLevel === 'premium' ? (
-				<span className={premiumClass} aria-label="Premium content">
-					Premium
+				<span>
```
i-like-robots: Is the outer span required?

i-like-robots:
Ah, I'm with you, thanks for clarifying @notlee.

components/x-teaser/src/Title.jsx
Thurst-ft: It seems so, throws an error without, which I think is to do with jsx not allowing adjsent elements.

i-like-robots: Ah, boo.

quarterto: you can in React with a `<Fragment>`, but that's not supported in stable Preact, so we can't use it in x-dash yet

Thurst-ft: @i-like-robots happy to take a recommendation on a better way.

i-like-robots: It's fine the way it is, no need to delay!

PR #307, opened 2019-05-21
X audio click tracking

robsquires:
1. Moves triggering of  notifiers from react lifecycle callbacks in `redux/index.jsx` to redux `notifications` middleware
2. Uses notifications middleware to trigger a tracking notifier, which will be hooked up to otracking app side

Couple points:

1. Are people happy with the `notification` terminology?
2. Regards to `isInternal` on the actions and state..I originally called it `isExternal` which I liked because everything is internal...so flagging things that can be both seemed logical. However, the logic in the notifications middleware looked a bit weird so I flipped it. What do you think?

### DIFF
--- a/components/x-audio/src/redux/__tests__/middleware/notifier.test.js
+++ b/components/x-audio/src/redux/__tests__/middleware/notifier.test.js
@@ -45,14 +45,14 @@ describe('Notifier middleware', () => {
 	});
 
 	test('external PLAY action does not trigger play notification', () => {
-		const { invoke, notifier, next } = create({ isPlayInternal: false });
+		const { invoke, notifier, next } = create({ willPlayNotify: false });
 		const action = actions.play();
 		invoke(action);
 		expect(notifier.play).not.toHaveBeenCalled();
 		expect(next).toBeCalledWith(action);
 	});
 	test('external PAUSE action does not trigger pause notification', () => {
-		const { invoke, notifier, next } = create({ isPauseInternal: false });
+		const { invoke, notifier, next } = create({ willPauseNotify: false });
 		const action = actions.pause();
 		invoke(action);
 		expect(notifier.pause).not.toHaveBeenCalled();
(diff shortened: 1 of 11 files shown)


components/x-audio/stories/redux-player.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
```
quarterto: maybe worth adding https://github.com/storybooks/storybook-addon-console? or is this just for testing

components/x-audio/src/redux/__tests__/middleware/notifier.test.js
```diff
@@ -0,0 +1,119 @@
...
+	});
+
+	describe('tracking', () => {
+		const state = { loading: true };
+
+		[
+			[ actions.requestPlay(), 'play', ['play-click', state ] ],
+			[ actions.requestPause(), 'pause', ['pause-click', state ]],
+			[ actions.willClose(), 'close', ['close-click', state ] ],
+		].forEach(([ action, notifyFn, payload ]) => {
```
benbarnett: Appreciate that this makes the tests DRY and the console output will be neat but I wonder if its worth just breaking these out...

Could potentially have a helper function like `fireAction()`? Then expect the relevant notifier to have called?

benbarnett: In fact `fireAction` would be pointless. But still wonder if its worth just duplicating them for the sake of simplicity in the tests

robsquires: Yeah know what you mean. Main intention of doing it was to make sure `next` is being called each time, cos it's not immediately clear from the code you need. Thought about using `before` and `after` hooks but got a bit messy. If we split it out, then future tests will just need to remember to check next has been called...which might be ok??!

benbarnett: Yeah, I think on reading it again it's fine as is. It was just the whole tests as documentation thing, which this does do!

robsquires: Split it out 👍

components/x-audio/stories/redux-player.js
robsquires: that plugin looks useful, I'll have a go at installing over on master

PR #308, opened 2019-05-22
X audio error text

debugwand:
ensure error text is always removed when we start loading again after an error


debugwand:
closing because pr should point to x-audio not master, and that has been done

PR #313, opened 2019-05-28
Add Snyk to components

jenniferemshepherd:
We are installing [Snyk](https://snyk.io) in all of our repos for security reasons. It will patch or ignore known vulnerabilities. Currently Snyk is enabled on our repos (config and results can be viewed from the Snyk webpage) but we want to install it on each project for increased security when building. See https://github.com/Financial-Times/next/issues/365


quarterto:
how does Snyk work in monorepos? does it need to be installed in every subpackage or is top-level sufficient?

sjparkinson:
> does it need to be installed in every subpackage or is top-level sufficient?

I believe as long as it's available on the command line that is good enough, even a global install should be okay to run the commands we have specified in our CircleCI jobs.

sjparkinson:
The thing to validate is if running `npx snyk monitor --org=customer-products --project-name=Financial-Times/x-dash` will pick up the `package.json` files in sub-directories, I believe it should.

sjparkinson:
There maybe something to consider about running `npx snyk protect`, but that is only a concern for apps. I don't think we actually need to run it for components.

PR #315, opened 2019-05-31
Expanded state

robsquires:


components/x-audio/src/components/ClickableContainer.jsx
```diff
@@ -0,0 +1,28 @@
...
+	...props
+}) => (
+	<div
+		{ ...props }
+		role='button'
+		tabIndex='0'
+		onKeyPress={() => onClick()}
+		onClick={({ target = {} }) => {
+			const tagName = target.tagName || '';
+			if (CLICKABLE_ELEMENTS.indexOf(tagName.toLowerCase()) === -1) {
```
benbarnett: `CLICKABLE_ELEMENTS.includes()`?

components/x-audio/src/components/ClickableContainer.jsx
```diff
@@ -0,0 +1,28 @@
...
+export const ClickableContainer = ({
+	children,
+	onClick,
+	...props
+}) => (
+	<div
+		{ ...props }
+		role='button'
+		tabIndex='0'
+		onKeyPress={() => onClick()}
```
benbarnett: Is this to deal with a particular scenario?

components/x-audio/src/components/ClickableContainer.jsx
robsquires: nice

components/x-audio/src/components/ClickableContainer.jsx
robsquires: eslint caught it: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/click-events-have-key-events.md

Not sure if it's best to restrict it to a particular key? /cc @debugwand

debugwand: the advice at the start of that plugin is wrong. and he doesn't mention anything about focus state.

certainly on desktop, click events are also triggered by pressing enter or space when you are focussed on an element. I believe click events are also triggered by double tapping when you are running a screenreader on a touch device. I am going to double check that so I'm not giving you bad advice.

debugwand: tested with https://jsbin.com/weginoduka/edit?html,js,output on webviewtest app and screenreader and it works as expected.

I don't know how screenreaders will interpret role of button on a div that contains other stuff, so when its merged in we can test with the phone and find out. It might need some hidden instructions to explain you get more controls when you activate it.

robsquires: ah good idea, will have a play when it's merged

debugwand: the one case I've found where it won't is using a keyboard and a non-typical element (as div here, anchors work) but no screenreader is active

In that case the keypresses we want to listen for are ones with keyCode 32 or 13
```
onKeyPress{(e) => {
if e.keyCode ===32 || e.keyCode===13 {
onClick();
}
});
```

debugwand: so: anchor element, button elements and checkbox inputs are the ones that don't need extra keypress handling. other elements like divs do, but only if you are not using a screenreader.

PR #318, opened 2019-05-31
Switch off click events have key events a11y rule

debugwand:
click events are triggered with a) mouse click b) pressing enter when focussed on an element c) double tapping when focussed on an element and using a screenreader

mousedown, mouseover, mouseup events need keyboard alternatives. not clicks.

very simple POC https://jsbin.com/weginoduka/edit?html,js,output


debugwand:
closing, this rule is true in certain situations, but not all.

PR #320, opened 2019-06-03
Refactor teaser snapshots

i-like-robots:
This refactor demonstrates how to separate a component from the automatic snapshot tests and implement component-level snapshot tests. This helps to uncouple us from the configuration based stories as discussed in #274.

- Abstracts data out of the teaser component story configuration into separate files which can be directly consumed by test files.
- Re-names the `stories` folder to `storybook` to prevent it being found by the automated test setup.


components/x-teaser/__tests__/snapshots.test.js
```diff
@@ -0,0 +1,27 @@
...
+		Object.entries(presets).forEach(([name, settings]) => {
+			it(`renders a ${name} teaser with ${type} data`, () => {
+				const props = { ...data, ...settings };
+				const tree = renderer.create(h(Teaser, props)).toJSON();
+
+				expect(tree).toMatchSnapshot();
+			});
+		});
+	});
+});
```
i-like-robots: I'm being quite lazy here by not writing some actual tests but this does preserve the same level of testing as we have with the automated snapshot generation.

PR #321, opened 2019-06-03
Fire `WILL_CLOSE` action with onCloseClick, rather than on unmount

benbarnett:
This is to fix a race condition where the close event is firing as soon as the close button is clicked, which results in the Redux state being cleared.

The tracking notifier is then run, which reports a `time_since_open` property. This is impossible to calculate because the time the player was opened is kept in Redux state, which has since been cleared.

So by moving the action into our own close button control, we can be sure the tracking notifier is called before we clear the state.


robsquires:
Nice

PR #323, opened 2019-06-03
Remove x-styling demo package

i-like-robots:
We have several components now demonstrating CSS modules, including `x-gift-article` in master so I believe this example component is no longer required.


benbarnett:
Worth adding a reference to some example implementations to the readme somewhere?/

i-like-robots:
I'm going to close this as several other pieces of work need to align first.

PR #324, opened 2019-06-04
Add playback speed control to expanded player

benbarnett:
And into the redux store.


fenglish:
@benbarnett I run this on ft-app. It's working!! On `0.5`, they speak really slow 😆

components/x-audio/src/components/PlaybackRate.jsx
```diff
@@ -0,0 +1,27 @@
...
+import classNameMap from './classnames-helper';
+
+import { getNextPlaybackRate } from './playback-rates';
+
+export const PlaybackRate = ({
+	onClick,
+	rate
+}) => (
+	<button
+		aria-label="Change playback speed"
```
debugwand: I found the best solution here was to update the aria-label to inform the user what the speed would change to. `aria-live` didn't make a difference. The user can get the button label to repeat

benbarnett: thanks @debugwand, I've added this now

PR #325, opened 2019-06-05
Add swipe down animation for expanded player

fenglish:
Swipe animation to minimise Expanded player using  [HammerJs](https://hammerjs.github.io/)


components/x-audio/src/redux/swipe-down-to-minimise.js
```diff
@@ -0,0 +1,31 @@
+const EXPANDED_PLAYER_HEIGHT = 360;
```
debugwand: can we get this from the DOM instead of saving it as a magic number?

components/x-audio/src/redux/index.jsx
```diff
@@ -65,6 +75,23 @@ export default function connectPlayer (Player) {
 			if (this.playingStateAndPropsNeedSync()) {
 				this.updatePlayingStateFromProps(prevProps);
 			}
+
+			if (this.expandedPlayerRef && !this.hasSetSwipeAnimation) {
+				this.setSwipeDownAnimation()
```
debugwand: lint: semi-colon at end of line

components/x-audio/src/redux/index.jsx
```diff
@@ -65,6 +75,23 @@ export default function connectPlayer (Player) {
...
+			}
+		}
+
+		setSwipeDownAnimation () {
+			const hammer = new Hammer.Manager(this.expandedPlayerRef);
+			hammer.add(new Hammer.Pan({
+				direction: Hammer.DIRECTION_ALL,
```
debugwand: could this be `DIRECTION_DOWN` or `DIRECTION_VERTICAL` instead?

benbarnett:
@fenglish I think the `package-lock.json` should be removed

components/x-audio/src/redux/swipe-down-to-minimise.js
```diff
@@ -0,0 +1,31 @@
...
+const DISPLAY_EXPANDED_THRESHOLD = EXPANDED_PLAYER_HEIGHT / 2;
+
+function setPositionY (target, posY) {
+	target.style.transform = `translate3d(0, ${posY}px, 0)`;
+}
+
+export default (event, playerActions) => {
+
+	const direction = event.offsetDirection;
```
debugwand: if we only listen to `DIRECTION_DOWN` then we presumably don't need these checks

components/x-audio/src/redux/swipe-down-to-minimise.js
```diff
@@ -0,0 +1,31 @@
...
+	// SWIPE ENDED
+	if (event.isFinal) {
+		if (posY >= DISPLAY_EXPANDED_THRESHOLD) {
+			playerActions.onMinimise();
+		}
+
+		setPositionY(expandedPlayer, 0);
+	}
+
+}
```
benbarnett: Wondering if the `redux` directory is the right location for this file. What do you think?

components/x-audio/src/redux/index.jsx
```diff
@@ -30,12 +32,20 @@ export default function connectPlayer (Player) {
...
+			this.hasSetSwipeAnimation = false;
 			this.state = initialState;
 		}
 
 		componentDidMount() {
 			const { playing, url, trackingContext } = this.props;
 			playerActions.loadMedia({ url, trackingContext, autoplay: playing });
+
+			if (this.expandedPlayerRef && !this.hasSetSwipeAnimation) {
+				this.setSwipeDownAnimation();
```
benbarnett: I wonder if this would be clearer named as `listenForSwipeDown`?

components/x-audio/src/redux/swipe-down-to-minimise.js
```diff
@@ -0,0 +1,31 @@
...
+export default (event, playerActions) => {
+
+	const direction = event.offsetDirection;
+	const directionDown = direction === 16;
+
+	if (!directionDown) {
+		return;
+	}
+
+	const expandedPlayer = event.target.closest("#audio-player-expanded");
```
benbarnett: Should this be using the `ref` from the main component? We could pass it in.

components/x-audio/src/redux/swipe-down-to-minimise.js
```diff
@@ -0,0 +1,31 @@
+const EXPANDED_PLAYER_HEIGHT = 360;
+const DISPLAY_EXPANDED_THRESHOLD = EXPANDED_PLAYER_HEIGHT / 2;
+
+function setPositionY (target, posY) {
+	target.style.transform = `translate3d(0, ${posY}px, 0)`;
+}
+
+export default (event, playerActions) => {
```
robsquires: so this file doesn't need to be aware of redux, could pass in an `onEnd` function?

components/x-audio/src/redux/index.jsx
```diff
@@ -65,6 +75,23 @@ export default function connectPlayer (Player) {
 			if (this.playingStateAndPropsNeedSync()) {
 				this.updatePlayingStateFromProps(prevProps);
 			}
+
+			if (this.expandedPlayerRef && !this.hasSetSwipeAnimation) {
```
robsquires: Is there any cleanup to do when the component unmounts, `hammer.destroy()` perhaps?

components/x-audio/src/redux/swipe-down-to-minimise.js
fenglish: @benbarnett Thank you for this comment!! I felt wrong when I put the file there, however I was not sure where it should be. Could you suggest where is a good place? `x-audio/src/lib`?

components/x-audio/src/redux/swipe-down-to-minimise.js
fenglish: 👍

components/x-audio/src/redux/index.jsx
fenglish: 👍

components/x-audio/src/redux/index.jsx
fenglish: 👍

components/x-audio/src/redux/swipe-down-to-minimise.js
fenglish: 👍

components/x-audio/src/redux/index.jsx
fenglish: 👍

components/x-audio/src/redux/index.jsx
fenglish: 👍

components/x-audio/src/redux/swipe-down-to-minimise.js
fenglish: 📝 I moved this file in components folder where other js file and helper file are. We will clean up when everything finished.

components/x-audio/src/redux/swipe-down-to-minimise.js
fenglish: 👍

components/x-audio/src/redux/swipe-down-to-minimise.js
fenglish: @benbarnett I got expanded player ref `null` when I did 
`minimised` => `expanded` => `minimised` => `expanded(the ref is null)`,
because I didn't know they create new expanded player 🙈 
I solved this by passing ref to `listenForSwipeDown` function.

benbarnett: Ah, well figured out!

components/x-audio/src/components/handle-swipe-down.js
```diff
@@ -0,0 +1,21 @@
...
+
+	const DISPLAY_EXPANDED_THRESHOLD = expandedPlayerRef.offsetHeight / 2;
+	const posY = event.deltaY;
+
+	setPositionY(expandedPlayerRef, posY);
+
+	// SWIPE ENDED
+	if (event.isFinal) {
+		if (posY >= DISPLAY_EXPANDED_THRESHOLD) {
+			onSwipeEnd();
```
benbarnett: One tiny thing, do we want to `return` here? If I understand it correctly there could be a very short delay before the `onMinimise` action does its thing, which means the player could possibly flick back to the top (because of line 18)?

components/x-audio/src/redux/index.jsx
```diff
@@ -65,6 +80,24 @@ export default function connectPlayer (Player) {
...
+			}
+		}
+
+		listenForSwipeDown (expandedPlayerRef) {
+			this.hammer = new Hammer.Manager(expandedPlayerRef);
+			this.hammer.add(new Hammer.Pan({
+				direction: Hammer.DIRECTION_DOWN,
+				threshold: 0
+			}) );
+			this.hammer.on("pan", (ev) => {
```
benbarnett: Pedant comment but these arguments are typically wrapped with single quotes 😄

components/x-audio/src/components/handle-swipe-down.js
fenglish: 📝 There is a reason, I couldn't `return` there. It needs to run `setPositionY(expandedPlayerRef, 0);` after `onSwipeEnd()`. I'm going to add comments to make the situation clear.

components/x-audio/src/redux/index.jsx
fenglish: @benbarnett Thank you!

PR #326, opened 2019-06-05
Build one cli

debugwand:
Convenience command so we can run commands like `make build-x-audio` to build just the x-audio component


docs/integration/local-development.md
```diff
@@ -46,3 +46,20 @@ If you encounter this error, ensure that x-dash is set up correctly in the paren
...
+
+If you don't want to build all the components while you are working on one, there is a convenience `make` command that allows you to limit builds by keyword
+
+* Add a keyword to your `package.json` file
+
+```javascript
+keywords: [
+	'x-name-of-component'
+]
+```
```
i-like-robots: By default `--filter` works on the name property, to target keywords you'd have to do something like this:

```sh
athloi run build --filter 'keywords:keyword'
```
```

debugwand: ah ok, I've been using x-audio but then the keywords and the name are identical. will update the docs to use name

PR #327, opened 2019-06-06
AT-2769: Teaser v2 base implementation

pavelpichrt:
Creating as a draft as it still needs:
- nicer exports of layouts
- unit tests
- documentation
- snapshots
- stories
- a bit of clean up

But code wise is pretty much there, so feel free to add any feedback


components/x-teaser/src/v2/components/Image/index.jsx
```diff
@@ -0,0 +1,24 @@
...
+
+// Transparent PNG with 16:9 dimensions
+const placeholderSrc = 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAKAAAABaCAQAAACVz5XZAAAAhklEQVR42u3QMQ0AAAwDoNW/6Yno1QQkkKMSBQIFChSIQIECBSJQoECBCBQoUCACBQoUiECBAgUiUKBAgQgUKFAgAgUKFIhAgQIFIlCgQIECEShQoEAEChQoEIECBQpEoECBAhEoUKBABAoUKBCBAgUKRKBAgQIRKFCgQAQKFChQoAKBApc9gMUAW29OZusAAAAASUVORK5CYII=';
+
+const Image = ({ src, isLazyLoaded, lazyLoadClassName, className }) => {
+	const fullClassName = c(className, lazyLoadClassName);
+	const actualSrc = src || placeholderSrc;
+	const commonProps = {
+		className: fullClassName,
+		alt: 'Article image',
```
notlee: Is this a default alt attribute for when one hasn't been provided? If so better to default to an empty string than have something with no meaning (but omitting the alt attribute makes most screen readers read out the url).

components/x-teaser/src/v2/components/Meta/styles.scss
```diff
@@ -0,0 +1,69 @@
...
+
+		&:after {
+			border-bottom-color: themed('metaColorPrimary');
+		}
+	}
+}
+
+.hasBorder:after {
+	border-bottom-width: oTypographySpacingSize(1);
+	margin-top: oTypographySpacingSize(2);
```
notlee: This is all good, but [o-spacing](https://github.com/Financial-Times/o-spacing) is available now 🎉 
`oTypographySpacingSize(2)` -> `oSpacingByName('s2')`

components/x-teaser/src/v2/components/Title/styles.scss
```diff
@@ -0,0 +1,56 @@
...
+
+		&:hover,
+		&:focus {
+			color: themed('titleTextColorHover');
+		}
+	}
+
+	text-decoration: none;
+	vertical-align: middle;
+	margin-right: 5px;
```
notlee: Is the `5px` arbitrary? Instead of set `px` values it would be great to use [o-spacing](https://github.com/Financial-Times/o-spacing) or a relative unit like `rem`, `ch`, `ex` -- makes things a little more consistent and amenable to changes. If it makes sense here

components/x-teaser/src/v2/components/Image/index.jsx
pavelpichrt: fixed

components/x-teaser/src/v2/components/Meta/styles.scss
pavelpichrt: updated

components/x-teaser/src/v2/components/Title/styles.scss
pavelpichrt: Checked with Mark and 4px is fine, so switched to `s1` spacing

components/x-teaser/src/v2/components/Timestamp/index.jsx
```diff
@@ -0,0 +1,26 @@
...
+import ftDateFormat from '@financial-times/ft-date-format';
+import c from 'classnames';
+
+import styles from './styles.scss';
+import { withThemes } from '../../themes/withThemes';
+
+const oneMonthMs = 2628000000;
+
+export const Timestamp = ({ publishedDate, className }) => {
+	const displayCutOffDate = Date.now() - oneMonthMs;
```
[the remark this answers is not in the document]
rowanbeentje: (Answer: we don't know for sure, was an eyeballed value)

PR #333, opened 2019-06-07
Fix expanded player not set swipe down

fenglish:
`this.hasSwipeDownListener` was not working as what it should be. It's never reset the value once it was set. 

I missed this bug because expanded player behaves differently between ft-app and storybook. On ft-app it works fine(the detail below). However the ft-app behaviour is incorrect and also the `hasSwipeDownListener` value is incorrect anyway.

------
 :confused: Weird behaviour on ft-app

When I change the player like the below, the player works differently between storybook and ft-app.
`minimised player` => `expanded player`1️⃣  => `minimised player` => `expanded player`2️⃣ 

**Storybook**
The expanded player 1️⃣  do swipe action.
The expanded player 2️⃣  doesn't do swipe action.

**ft-app**
The expanded player 1️⃣  do swipe action.
The expanded player 2️⃣  do swipe action.

I'm not sure the reason but it looks like the player on ft-app opens the cached expanded player every time once it was created.

### DIFF
--- a/components/x-audio/src/redux/index.jsx
+++ b/components/x-audio/src/redux/index.jsx
@@ -6,6 +6,8 @@ import createStore from './store';
 import { NotifiersProxy } from './middleware/notifier';
 import handleSwipeDown from '../components/handle-swipe-down';
 
+const swipeDownClass = 'added-swipe-down';
+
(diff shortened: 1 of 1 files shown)


fenglish:
@benbarnett Thank you for your advice! It works beautifully.

PR #335, opened 2019-06-10
Fix tabable image issue

fenglish:
This is a part of fixing DAC issues. Teasers from x-teaser provides tabable image links which leads repetitive navigation for the screen reader users.

The DAC trello ticket => https://trello.com/c/JCgvCagH/79-dacredundantlinksissue1

-----

Screenshot 2019-06-10 at 14 48 18

On selecting image link(For ex: 'Extinction Rebellion...') and links(for ex: 'Extinction Rebellion: inside the new climate resistance') the sane page gets opened which leads to repetitive navigation for the screen reader user.
Similar issue has been observed on 'News article' page.


i-like-robots:
Thanks! This appears to differ by the rendering engine used but this should make it consistent 🤞

PR #336, opened 2019-06-10
Enable regular story format

i-like-robots:
This demonstrates how to define a Storybook demo for a component using [regular Storybook configuration](https://storybook.js.org/docs/basics/writing-stories/) rather than our previous custom format.

To do this I did three things:

1. Removed the stories configuration from the `.storybook/register-components.js` module, this skips Story generation and removes the component from the automated snapshot tests.
2. Rewrote the stories using regular `storiesOf()` format
3. Added this module import to `.storybook/config.js`

I have refactored the `x-increment` component stories to use this format.


components/x-increment/storybook/index.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { Increment } from '../src/Increment';
```
i-like-robots: Look mom, direct from source code!

i-like-robots:
As discussed IRL, this is strangely how it is documented: https://storybook.js.org/docs/guides/guide-react/#step-3-create-the-config-file

The reasons for not going full ESM are:

1. hoisting the story imports up to the top scope breaks Storybook
2. using an async `import()` also breaks Storybook

But we could refactor this to be fully CJS, although that would span several files to avoid wrangling `.default`.

robsquires:
Works nicely. Not sure if it's intended, but I've used `buildStory` as opposed to `storiedOf`. Takes care of all the dependency stuff, but still allows me to override the component + knobs for a specific story 👍

i-like-robots:
That's fine for now, if it's a good pattern then we don't need to get rid of it entirely.

PR #339, opened 2019-06-11
Add package-lock.json files to .gitignore configuration

### LINKED ISSUE #239: x-topic-search and x-interaction challenges
In order for [x-topic-search](https://github.com/Financial-Times/x-dash/pull/218)'s results to be closed when you interact with something else on the page, we need do something like this:

```javascript
['focusout', 'focusin', 'click'].forEach(action => {
	document.body.addEventListener(action, event => {
		if(!componentRootNode.contains(event.target)) {
			topicSearchActions.hideResult();
		}
	});
});
```

(We can't just listen for the `blur` of the topic search input, as this will cause the search results to close when you click inside the search results (e.g. on the follow button).)

So, there's a couple of issues here:

1. What is an acceptable way for an x-dash component to add (and remove) DOM event listeners outside of themselves? (we should avoid it if we can, but in this case it seems unavoidable).
2. How to pass a ref to a DOM node from the wrapped component to the action? E.g. the event handler has this condition: `componentRootNode.contains(event.target)`

### LINKED ISSUE #329: use a precommit hook to prevent package-lock.json files being committed
despite our `.npmrc` and `make install` `--no-package-lock` configurations, these files keep cropping up and causing build problems. we should add a pre-commit hook to remove these files from any commit (with a warning).

i-like-robots:
As it says! Fixes #329.


.gitignore
```diff
@@ -1,5 +1,6 @@
-node_modules
-bower_components
-npm-debug.log
+bower_components/
+node_modules/
+dist/
 .DS_Store
```
benbarnett: Always feel this should live in a global config on dev machines rather than being a project concern. Belt and braces though.

quarterto:
"fixes #239"? sure you've got the right issue number?

i-like-robots:
Good point, but the issue referenced specifically references issues caused by committing these files so this should fix that.

In Anvil we have sprinkled `.npmrc` files around the place to disable lockfiles, shall I add one to the repo root?

Example `.npmrc`:

```
package-lock=false
shrinkwrap=false
```

.gitignore
i-like-robots: Agreed! But yes, belt and braces

jenniferemshepherd:
so... we merging this?

edds:

PR #341, opened 2019-06-18
x-audio: Use a marquee for text that spans a wider area than we have

benbarnett:
Including a small change to the grid for the minimised player – we need to concatenate the series name and title into a single element.

Please note that it is far smoother than the gif makes it look!

marquee


components/x-audio/src/components/styles.scss
```diff
@@ -57,8 +57,8 @@ $image-size: 72px;
 }
 
 .audio-player--minimised {
-	grid-template-columns: $height--minimised min-content auto $height--minimised;
-	grid-template-areas: "play-pause series-name title close-button"
+	grid-template-columns: $height--minimised auto auto $height--minimised;
+	grid-template-areas: "play-pause series-name series-name close-button"
 	"play-pause status status close-button";
```
fenglish: This grid can be less `columns` like below because we don't need to separate series-name and title anymore.
```
  grid-template-columns: $height--minimised auto $height--minimised;
  grid-template-areas: "play-pause series-name close-button"
  "play-pause status close-button";
```

benbarnett: Very good point, thanks!

benbarnett:
@fenglish thanks for your suggestion about `draw()` - I've moved it out in this commit: https://github.com/Financial-Times/x-dash/pull/341/commits/9ad7c9c1349469c35c46e4ef74836269fa0a577f

PR #346, opened 2019-06-20
Add adjustments for swiping

debugwand:
Handing over because I am in meetings until 5 today

* stops the user from swiping above the original top position
* stops the user from accidentally swiping while using the scrubber

requires:

*tests
*we might be able to use just the redux state in timeline - however removing this.setState({ scrubbing: true }) stops the handle from moving as you swipe


components/x-audio/src/redux/__tests__/index.test.jsx
```diff
@@ -0,0 +1,91 @@
...
+	}
+};
+
+describe('Connected player', () => {
+
+	beforeEach(() => {
+		NotifiersProxy.mockClear();
+	});
+
+	test('loads media when the player is initiallly rendered', () => {
```
benbarnett: triple 'l' in `initiallly`

components/x-audio/src/redux/player-logic.js
```diff
@@ -32,6 +33,7 @@ export const MINIMISE = 'MINIMISE';
 export const SET_PLAYBACK_RATE = 'SET_PLAYBACK_RATE';
 export const SEEKING = 'SEEKING';
 export const SEEKED = 'SEEKED';
+export const SCRUBBING = 'SCRUBBING';
```
benbarnett: very minor, but would `SET_SCRUBBING` be more consistent naming here?

benbarnett: or even `UPDATE_`?

PR #349, opened 2019-06-24
Display image from props

fenglish:


components/x-audio/src/components/Players.jsx
```diff
@@ -35,7 +37,7 @@ export const ExpandedPlayer = ({
 		<Rewind currentTime={currentTime} updateCurrentTime={updateCurrentTime} />
 		<Forward currentTime={currentTime} updateCurrentTime={updateCurrentTime} duration={duration} />
 		<PlaybackRate rate={playbackRate} onClick={newRate => onPlaybackRateClick({ playbackRate: newRate })} />
-		<div className={classNameMap('audio-player__info__image')}><img alt="dummy"/></div>
+		<Image imageDataSet={imageDataSet} />
```
benbarnett: Should this be `imageSrcSet`?

benbarnett: Suppose it contains the srcSet so isn't strictly that. Makes sense as is - ignore!

fenglish:
some reason (I don't know the reason maybe circle side problem), the circle build doesn't pass.
I have created new PR with the same changes => #350

PR #356, opened 2019-07-10
Call audio.load() on LOAD_MEDIA action to trigger preloading

benbarnett:
This should deal with the delaying in audio being played seen on iOS/Safari devices.

It seems to be that Safari doesn't honour the `audio.preload` attribute until `audio.load()` is called. In other browsers, setting `audio.src` is enough.

Current use case for x-audio is that the audio player is only shown after a user has requested to play audio, so preloading is ok here. In the future we may need to make this an option.


robsquires:
Just tested locally, I'm still seeing the delay between removing the loading spinner and playback starting on ios. I think we also need to change the `LOADED` action to be dispatched by `canplaythrough`.

PR #359, opened 2019-07-12
Modify player design

fenglish:
### Appy new layout
Screenshot 2019-07-12 at 16 56 14

### Optimised the css grid for expanded player

BEFORE --- 3 x 6
Screenshot 2019-07-12 at 15 00 29

AFTER --- 1 x 7
Screenshot 2019-07-12 at 14 57 42

### DIFF
--- a/components/x-audio/src/components/Players.jsx
+++ b/components/x-audio/src/components/Players.jsx
@@ -38,9 +38,6 @@ export const ExpandedPlayer = ({
 		onSwipeEnd={onMinimise}
 		swipeEnabled={!scrubbing}>
 		<button onClick={() => onMinimise()} className={classNameMap('audio-player__minimise-button')} aria-label='minimize player'/>
-		<Rewind currentTime={currentTime} updateCurrentTime={updateCurrentTime} />
-		<Forward currentTime={currentTime} updateCurrentTime={updateCurrentTime} duration={duration} />
-		<PlaybackRate rate={playbackRate} onClick={newRate => onPlaybackRateClick({ playbackRate: newRate })} />
 		<Image imageDataSet={imageDataSet} />
 		<Title text={title} />
 		<SeriesName text={seriesName} />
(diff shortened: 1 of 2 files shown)


components/x-audio/src/components/Players.jsx
```diff
@@ -11,8 +11,12 @@ import { TimeRemaining } from './TimeRemaining';
...
 import { Image } from './Image';
 
-const Title = ({ text }) => <Marquee className={classNameMap('audio-player__info__title')} text={text} />
-const SeriesName = ({ text }) => <Marquee className={classNameMap('audio-player__info__series-name')} text={text} />
+const Title = ({ text }) => (
+	<div className={classNameMap('audio-player__title__marquee')}>
+		<Marquee className={classNameMap('audio-player__info__title')} text={text} />
+	</div>
+);
+const SeriesName = ({ text }) => <div className={classNameMap('audio-player__info__series-name')}><span title={text}>{text}</span></div>;
```
benbarnett: Does the series name not need to be a marquee anymore?

components/x-audio/src/components/Players.jsx
```diff
@@ -11,8 +11,12 @@ import { TimeRemaining } from './TimeRemaining';
 import { Marquee } from './Marquee';
 import { Image } from './Image';
 
-const Title = ({ text }) => <Marquee className={classNameMap('audio-player__info__title')} text={text} />
-const SeriesName = ({ text }) => <Marquee className={classNameMap('audio-player__info__series-name')} text={text} />
+const Title = ({ text }) => (
+	<div className={classNameMap('audio-player__title__marquee')}>
```
benbarnett: Totally up to you, but I would say the original class name (`audio-player__title`) was sufficient here.

fenglish: 👍

components/x-audio/src/components/Players.jsx
fenglish: 📝 The reasons why we remove the marquee from `SeriesName`
- When  both `Title` and `SeriesName`'s scrolling marquee animation happen, it is impossible to read either of them.
- Initially it was not asked by Designer.
- `SeriesName` is tend to be short. It's an edge case that `SeriesName` needs the animation.

We will put back it when requested.

fenglish:
Thank you @benbarnett !

PR #360, opened 2019-07-15
Hide image on mobile landscape

fenglish:
local ft com_8080_vanity-test-article-22cda1acxd2a0x42ecx928axf26f6b5c34b6(iPhone X)

### DIFF
--- a/components/x-audio/src/components/styles.scss
+++ b/components/x-audio/src/components/styles.scss
@@ -1,7 +1,6 @@
 @import 'o-icons/main';
 @import 'o-colors/main';
 
-
 $o-typography-load-fonts: false;
 @import 'o-typography/main';
 @include oTypography();
@@ -14,7 +13,6 @@ $image-size: 160px;
 	width: 64px;
 	height: 64px;
 	background-color: oColorsGetPaletteColor('white');
-	margin: 16px;
 	border-radius: 32px;
 
 	&:focus,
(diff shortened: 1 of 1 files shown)


components/x-audio/src/components/styles.scss
```diff
@@ -359,3 +358,36 @@ $primary-scrubber-color: oColorsGetPaletteColor('teal');
 // 	width: var(--buffered-width);
 // 	background-color: green;
 // }
```
benbarnett: Oh I know this change is nothing to do with this PR but I think this code was accidentally committed. That's right isn't it @robsquires? I think this was an experiment for buffering stuff that we're not doing anymore.

components/x-audio/src/components/styles.scss
```diff
@@ -359,3 +358,36 @@ $primary-scrubber-color: oColorsGetPaletteColor('teal');
...
+
+@media (orientation: landscape) and (max-height: 620px - 1) {
```
benbarnett: Just looking at this, I suppose the actual issue here is just the `max-height`, not necessarily the orientation. Is that right? Like if you shrank your browser down to a small height, you'd probably also want these rules to apply. What do you think?

fenglish: @benbarnett That's true. I'm going to remove the landscape query. Thank you!

components/x-audio/src/components/styles.scss
robsquires: yup totally, good spot!

PR #362, opened 2019-07-17
X audio loaded

debugwand:
Set x-audio state to loaded when we get the playing event from the audio object, as when events fire can happen in a different order to what you might expect when you fast forward

https://docs.google.com/spreadsheets/d/1lYdCJv74oyVhyHyoyZIR_Lh5TnuyNQ_cifOsvAnrGdU/edit#gid=747106634


debugwand:
wrong branch

PR #378, opened 2019-08-01
x-podcast-launchers component

robsquires:
Screenshot 2019-08-20 at 11 45 12

ft.com => https://github.com/Financial-Times/next-article/pull/3430
ft-app => https://github.com/Financial-Times/ft-app/pull/459


components/x-podcast-launchers/src/generate-rss-url.js
```diff
@@ -0,0 +1,8 @@
+import nanoid from 'nanoid';
+
+const ACAST_RSS_URL = 'https://access.acast.cloud/rss';
```
robsquires: Given there is a prod URL too, maybe we should pass this in as a prop?

components/x-podcast-launchers/src/PodcastLaunchers.jsx
```diff
@@ -0,0 +1,49 @@
+import { h, Component } from '@financial-times/x-engine';
+import generateAppLinks from './generate-app-links';
+import generateRSSUrl from './generate-rss-url';
+
+function mapConceptToAcastSeries (/** concept */) {
```
robsquires: I think we'd need to derive the acast series name from the podcast series concept

components/x-podcast-launchers/src/generate-rss-url.js
```diff
@@ -0,0 +1,8 @@
...
+
+export default function generateRSSUrl(seriesId) {
+	const token = nanoid(10);
```
robsquires: Would it be worth logging these to spoor, just so in the future we'd have a list of all the tokens we'd issued?

components/x-podcast-launchers/src/PodcastLaunchers.jsx
```diff
@@ -0,0 +1,49 @@
...
+	return 'ft-test';
+}
+
+class PodcastLaunchers extends Component {
```
benbarnett: Wondering if this needs to be a class component?

components/x-podcast-launchers/src/PodcastLaunchers.jsx
benbarnett: and something something https://github.com/Financial-Times/n-concept-ids/blob/master/src/index.js

components/x-podcast-launchers/src/generate-rss-url.js
benbarnett: Yeah could live in Vault maybe

components/x-podcast-launchers/src/generate-rss-url.js
benbarnett: Part of me thinks we should either store them properly or not store them at all... where would they actually go if we log them in spoor?

I suppose also we would have the o-tracking stuff on the hrefs on the links themselves?

fenglish:
@robsquires we should delete `components/x-podcast-launchers/package-lock.json`. It causes an issue (although I don't remember the detail of the issue...). We had it when we were working on `x-audio`.

components/x-podcast-launchers/src/PodcastLaunchers.jsx
debugwand: you are much safer relying on the concept ID, that is less likely to change. we could maintain a mapping of concept IDs to list of acast urls within this component

components/x-podcast-launchers/src/generate-rss-url.js
benbarnett: We have an easy way to revoke these with Acast and Matt (PO) doesn't need them

components/x-podcast-launchers/src/generate-rss-url.js
benbarnett: This is now a prop, so its up the apps how they pass this in

components/x-podcast-launchers/src/PodcastLaunchers.scss
```diff
@@ -0,0 +1,52 @@
...
+
+.rss-url__input {
+	max-width: none;
+}
+
+.rss-url__copy-button {
+	padding-left: 0px;
+}
+
+.rss-url__copy-span {
```
i-like-robots: Question: Is there a reason to use BEM style class names rather than JS compatible identifiers, e.g. `.rssUrlCopyButton`?

components/x-podcast-launchers/src/__tests__/PodcastLaunchers.test.jsx
```diff
@@ -0,0 +1,23 @@
+import { h } from '@financial-times/x-engine';
+import { brand } from '@financial-times/n-concept-ids';
+import renderer from 'react-test-renderer';
+jest.mock('../PodcastLaunchers.scss', () => ({}));
```
i-like-robots: Neat, I did wonder if this would be a pain - I guess not!

components/x-podcast-launchers/src/generate-rss-url.js
```diff
@@ -0,0 +1,6 @@
+import nanoid from 'nanoid';
+
+export default function generateRSSUrl(acastHost, seriesId, token) {
+	// ie. https://access.acast.cloud/rss/ft-test/tYPWWHla
+	return `${acastHost}/rss/${seriesId}/${token || nanoid(10)}`;
```
i-like-robots: I think this needs some explanation, why add a random token?

components/x-podcast-launchers/readme.md
```diff
@@ -0,0 +1,63 @@
+# x-podcastlaunchers
```
i-like-robots: Missing dash!

components/x-podcast-launchers/readme.md
```diff
@@ -0,0 +1,63 @@
...
+
+// A == B == C
+const a = PodcastLaunchers(props);
+const b = <PodcastLaunchers {...props} />;
+const c = React.createElement(PodcastLaunchers, props);
+```
+
+```scss
+// within your app's sass file
+@import "x-podcast-launchers/dist/PodcastLaunchers";
```
i-like-robots: I don't think this is the correct identifier

components/x-podcast-launchers/readme.md
```diff
@@ -0,0 +1,63 @@
...
+```
+:warning: This component depends on styles provided by o-forms and o-buttons, and therefore o-forms and o-buttons needs to be imported before x-podcast-launchers.
```
i-like-robots: Importing all of `o-forms` and `o-buttons` can add ~150kb of CSS to a stylesheet, maybe specify which features are actually required?

components/x-podcast-launchers/src/PodcastLaunchers.scss
```diff
@@ -0,0 +1,52 @@
...
+}
+
+.container {
+	font-family: MetricWeb,sans-serif;
+	width: 100%;
+}
+
+.heading {
+	font-weight: 600;
+	font-size: 18px;
```
i-like-robots: Should this use `o-typography`?

components/x-podcast-launchers/src/map-concept-to-acast-series.js
```diff
@@ -0,0 +1,10 @@
+import { brand } from '@financial-times/n-concept-ids';
+
+export default function mapConceptToAcastSeries(conceptId) {
```
i-like-robots: I think a comment here explaining why one thing is weird (and pref. who is responsible for it being weird) would be useful

components/x-podcast-launchers/readme.md
fenglish: 👍

components/x-podcast-launchers/readme.md
fenglish: Oh, yeah it is not correct. I will modify it.

components/x-podcast-launchers/readme.md
fenglish: 👍

components/x-podcast-launchers/src/map-concept-to-acast-series.js
benbarnett: @i-like-robots how do you mean? As in why does only one `conceptId` have a mapping?

i-like-robots: I looked at this in isolation but looking at where it is implemented makes more sense.

components/x-podcast-launchers/readme.md
```diff
@@ -0,0 +1,70 @@
...
+This component also renders a myFT follow button (x-follow-button) for the conceptId provided. This is acts as an onsite way to follow the series should the user's podcast app not be listed.
+
+![screenshot of x-podcast-launchers](https://user-images.githubusercontent.com/21194161/63341610-86996080-c341-11e9-8a21-04da9c8bb6cc.png)
+
+## Installation
+
+This module is compatible with Node 6+ and is distributed on npm.
+
+```bash
+npm install --save @financial-times/x-podcastlaunchers
```
i-like-robots: Missing `-`

components/x-podcast-launchers/readme.md
```diff
@@ -0,0 +1,70 @@
...
+### Properties
+
+Feature              | Type     | Required | Notes
+---------------------|----------|----------|------------------
+`acastRSSHost`       | String   | Yes      | e.g. 'https://acast.access.com'
+`conceptId`          | String   | Yes      |
+`conceptName`        | String   | Yes      | x-follow-button prop.
+`isFollowed`         | Boolean  | Yes      | x-follow-button prop. Whether the podcast concept id is followed or not.
+`csrfToken`          | String   | Yes      | x-follow-button prop. Value included in a hidden form field.
+`renderFollowButton` | Function | No       | x-follow-button prop. Optional render prop.
```
i-like-robots: It might be easier to link off to the `x-follow-button` documentation rather than trying to keep the two in sync.

components/x-podcast-launchers/src/PodcastLaunchers.jsx
```diff
@@ -0,0 +1,105 @@
...
+							{name}
+						</a>
+					</li>
+				))}
+				</ul>
+
+				<div className={rssUrlWrapperStyles}>
+					<input className={rssUrlInputStyles} value={rssUrl} type='text' readOnly/>
+					<div className={rssUrlCopyButtonWrapperStyles}>
+						<button
```
i-like-robots: To be completely safe I think a `type=button` attribute might be needed because the default is `type=submit` which has different semantics.

components/x-podcast-launchers/src/PodcastLaunchers.scss
```diff
@@ -0,0 +1,52 @@
...
+
+.rssUrlInput {
+	max-width: none;
+}
+
+.rssUrlCopyButton {
+	padding-left: 0px;
+}
+
+.rssUrlCopySpan {
```
i-like-robots: Much nicer!

components/x-podcast-launchers/src/generate-app-links.js
```diff
@@ -0,0 +1,13 @@
...
+	return appLinksConfig.map(data => {
+		const url = data.includeProtocol
+					? rssUrl
+					: rssUrl.replace(/^https?:\/\//, '');
+
+		return Object.assign({}, data , {
+			url: data.template.replace(/{url}/, url)
+		})
+	});
+}
```
i-like-robots: I think this file needs Prettier run over it

components/x-podcast-launchers/readme.md
fenglish: Oh 🙈 sorry. I missed `-` again

components/x-podcast-launchers/readme.md
fenglish: 👍

components/x-podcast-launchers/src/PodcastLaunchers.jsx
fenglish: I didn't know this. It's interesting. Thank you for telling!

components/x-podcast-launchers/src/PodcastLaunchers.scss
fenglish: 🙌

components/x-podcast-launchers/src/generate-app-links.js
fenglish: 👍

PR #379, opened 2019-08-05
Add copyToClipboard function

fenglish:
`user-select: none;` is set in ft-app's body. Therefore `span` needs to be set `user-select: text;` otherwise it won't select the text to copy.

I have checked that this copy function works on Chrome, Safari, firefox and IE11.

### DIFF
--- a/components/x-podcast-launchers/src/PodcastLaunchers.css
+++ b/components/x-podcast-launchers/src/PodcastLaunchers.css
@@ -26,13 +26,7 @@
 }
 
 .rss-url__copy-span {
-	-webkit-user-select: text;
-	-khtml-user-select: text;
-	-moz-user-select: text;
-	-ms-user-select: text;
-	-o-user-select: text;
 	user-select: text;
-
 	position: absolute;
 	clip: rect(0 0 0 0);
 	margin: -1px;
--- a/components/x-podcast-launchers/src/PodcastLaunchers.css
+++ b/components/x-podcast-launchers/src/PodcastLaunchers.css
@@ -1,0 +1,45 @@
+.container {
+	width: 100%;
+}
+
+.podcast-app-links__wrapper {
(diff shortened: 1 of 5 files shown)


components/x-podcast-launchers/src/copy-to-clipboard.js
```diff
@@ -0,0 +1,25 @@
...
+		selection.removeAllRanges();
+		selection.addRange(range);
+		inputEl.setSelectionRange(0, 999999);
+	} catch (err) {
+		inputEl.select(); // IE11 etc.
+	}
+	inputEl.contentEditable = oldContentEditable;
+	inputEl.readOnly = oldReadOnly;
+	document.execCommand('copy');
+	inputEl.blur();
```
debugwand: where does the cursor end up after you've done this? is there a way to avoid moving the focus on to the element and then blurring again? that way the user's cursor stays where it is

we've another example of copy code here: https://github.com/Financial-Times/next-myft-page/blob/d106acb16df3b7b2f6991c0cf2426c3fc7e7900e/client/components/alerts/js/rss-preference.js

components/x-podcast-launchers/src/PodcastLaunchers.css
```diff
@@ -24,3 +24,22 @@
 .rss-url__copy-button {
 	padding-left: 0px;
 }
+
+.rss-url__copy-span {
+	-webkit-user-select: text;
```
benbarnett: Should autoprefixer take care of this for us? n-ui should run styles through it, and I'm pretty sure it will be fine in the app.

fenglish: @benbarnett Thank you for pointing this. I have checked next-article and ft-app provide those vendor prefixes.

PR #380, opened 2019-08-06
Add Styles

fenglish:
Screenshot 2019-08-06 at 13 56 00

### DIFF
--- a/components/x-podcast-launchers/readme.md
+++ b/components/x-podcast-launchers/readme.md
@@ -40,7 +40,7 @@ const c = React.createElement(PodcastLaunchers, props);
 // within your app's sass file
 @import "x-podcast-launchers/dist/PodcastLaunchers";
 ```
-:warning: It includes classes to be overwritten o-forms styling. Therefore it needs to be imported after o-forms is imported;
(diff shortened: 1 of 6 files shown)


components/x-podcast-launchers/readme.md
```diff
@@ -15,21 +15,33 @@ The [`x-engine`][engine] module is used to inject your chosen runtime into the c
...
 const a = PodcastLaunchers(props);
 const b = <PodcastLaunchers {...props} />;
 const c = React.createElement(PodcastLaunchers, props);
 ```
 
+```scss
+// within your app's sass file
+@import "x-podcast-launchers/dist/PodcastLaunchers";
+```
+:warning: It includes classes to be overwritten o-forms styling. Therefore it needs to be imported after o-forms is imported;
```
benbarnett: Could this be worded as something like this?

> Warning: this component depends on styles provided by o-forms, and therefore o-forms needs to be imported before x-podcast-launchers.

benbarnett: Or even:

> and therefore o-forms needs to be imported alongside this component.

fenglish: @benbarnett Thank you for your suggestion! I will change the sentences as you suggested.

PR #382, opened 2019-08-09
Map FT concept ID to acast series ID

robsquires:
Maps an FT concept ID from [n-concept-ids](https://github.com/Financial-Times/n-concept-ids/blob/master/src/index.js#L32) to an acast private show ID.

For now just using `ft-test`, until The Rachman Review is created by acast.


robsquires:
*PR of the Year 2019* ?

PR #265, opened 2019-03-11
AT-2015: Use a render prop for the follow button to make it work with…

pavelpichrt:
… x-follow-button 0.0.12

If this is your first `x-dash` pull request please familiarise yourself with the contribution guide before submitting.

**TL;DR**

  - Discuss features first
  - Update the documentation
  - No hacks, experiments or temporary workarounds
  - Reviewers are empowered to say no
  - Reference other issues
  - Update affected stories and snapshots
  - Follow the code style
  - Decide on a version (major, minor, or patch)

PR #383, opened 2019-08-09
Add headings and follow button

fenglish:
Screenshot 2019-08-09 at 16 10 37

### DIFF
--- a/components/x-podcast-launchers/src/PodcastLaunchers.jsx
+++ b/components/x-podcast-launchers/src/PodcastLaunchers.jsx
@@ -70,9 +70,9 @@ class PodcastLaunchers extends Component {
 		return (
 			<div className={styles['container']}>
 				<h2 className={styles['heading']}>Subscribe on a podcast app</h2>
+
 				{rssUrl && (
 					<div>
-
 						<ul className={styles['podcast-app-links__wrapper']}>
 						{generateAppLinks(rssUrl).map(({ name, url, trackingId }) => (
 							<li key={name}>
@@ -97,9 +97,8 @@ class PodcastLaunchers extends Component {
 								</button>
 							</div>
 						</div>
-
 					</div>
-				}
+				)}
 
 				<h2 className={styles['heading']}>Can't see your podcast app?</h2>
 				{renderFollowButton(conceptId, conceptName, csrfToken, isFollowed)}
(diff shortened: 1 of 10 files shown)


components/x-podcast-launchers/src/PodcastLaunchers.jsx
```diff
@@ -53,30 +64,36 @@ class PodcastLaunchers extends Component {
...
-							<div className={rssUrlCopyButtonWrapperStyles}>
-								<button
-									className={basicButtonStyles}
-									onClick={copyToClipboard}
-									data-url={rssUrl}>
-									Copy RSS
-								</button>
-							</div>
+				<h2 className={styles['heading']}>Subscribe on a podcast app</h2>
+				<div>
```
benbarnett: I think given we have these headings now, we might not need this extra wrapping `div`?

components/x-podcast-launchers/src/PodcastLaunchers.jsx
```diff
@@ -32,6 +33,16 @@ const rssUrlCopyButtonWrapperStyles = [
 ].join(' ');
 
 
+function defaultFollowButtonRender (conceptId, conceptName, csrfToken, isFollowed) {
```
benbarnett: I could be convinced either way here, but although having this as a render prop brings more flexibility, it is unlikely we will actually need it. Especially as this component is being used as part of an experiment, it might better to keep things as simple as possible.

The other slightly confusing thing is that this component will always bring in x-follow-button styles, regardless of whether this render prop is used. What do you think?

components/x-podcast-launchers/src/PodcastLaunchers.jsx
fenglish: 👍

components/x-podcast-launchers/src/PodcastLaunchers.jsx
fenglish: 📝 We need to pass a function to render follow button for ft-app

The reason why ft-app needs own function to render follow buttons.
>We need to add tracking and state management, which is not available on the follow button itself. FT.com is currently using events fired on body to handle these.

Related:
https://github.com/Financial-Times/x-dash/pull/265
https://github.com/Financial-Times/ft-app/blob/569500519fb7ff9e02690541a8cbf5a6110a0e43/lib/components/myft/manage/index.js#L58

benbarnett: Thanks @fenglish! Makes total sense 👍

PR #386, opened 2019-08-14
Pass renderFollowButton from props

fenglish:
I forgot to pass the prop...

### DIFF
--- a/components/x-podcast-launchers/src/PodcastLaunchers.jsx
+++ b/components/x-podcast-launchers/src/PodcastLaunchers.jsx
@@ -95,7 +95,7 @@ class PodcastLaunchers extends Component {
 					</div>
 				</div>
 
-				<h2 className={styles['heading']}>Can't see your podcast app?</h2>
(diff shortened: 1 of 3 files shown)


components/x-podcast-launchers/src/PodcastLaunchers.jsx
```diff
@@ -96,7 +96,7 @@ class PodcastLaunchers extends Component {
 				</div>
 
 				<h2 className={styles['heading']}>Can't see your podcast app?</h2>
```
benbarnett: Sorry I should've noticed this before... can we make this a curly apostrophe? (`’`)

fenglish: ohh, I will change it to a curly apostrophe! Thank you!

benbarnett: Thanks, sorry it is so minor 🙈

PR #388, opened 2019-08-20
Update Readme

fenglish:


components/x-podcast-launchers/readme.md
```diff
@@ -50,6 +51,7 @@ All `x-` components are designed to be compatible with a variety of runtimes, no
 
 Feature              | Type     | Required | Notes
 ---------------------|----------|----------|------------------
+`acastRSSHost`       | String   | Yes      |
```
benbarnett: Might be worth adding an example here in the notes?

> e.g. 'https://acast.access.com'

components/x-podcast-launchers/readme.md
```diff
@@ -1,7 +1,8 @@
 # x-podcastlaunchers
 
-This module has these features and scope.
+This module allows a user to subscribe a podcast series on their podcast app. The subscribe urls for each podcast app are generated from rss url with config(`/src/app-links.js`). No elements are returned when the series concept id isn't set as a case in `src/map-concept-to-acast-series.js`. It has myft follow button(x-follow-button) for the concept, in case there was no podcast apps which the user normally uses.
```
benbarnett: I think this is really good and makes all the points it needs to. I have put a suggestion below - what do you think? My thinking is that breaking it into smaller paragraphs can make things slightly easier to read when coming to this component having not seen it before.

```suggestion
This module allows users to open a podcast series in various podcast apps. The subscribe urls for each podcast app are generated from rss url with config (`/src/app-links.js`). 

No elements are returned when the `conceptId` does not map to a known podcast series, as defined in `src/map-concept-to-acast-series.js`. 

This component also renders a myFT follow button (x-follow-button) for the conceptId provided. This is acts as an onsite way to follow the series should the user's podcast app not be listed.
```

PR #389, opened 2019-08-22
Reflect reviews' comments

fenglish:
To address Matt's comments on `x-podcast-launchers` branch.


components/x-podcast-launchers/src/map-concept-to-acast-series.js
```diff
@@ -1,10 +1,9 @@
 import { brand } from '@financial-times/n-concept-ids';
 
+const acastSeriesMap = new Map([
```
robsquires: @fenglish I've tweaked this slightly so we export the Map itself. Think it makes this module a bit clearer

fenglish: @robsquires Thank you for doing that and also thank you for merging this PR.

PR #392, opened 2019-08-28
Add additional data trackables around podcast launchers

benbarnett:
To make it easier to locate the buttons in spoor.

image


fenglish:
Thank you! @robsquires @i-like-robots

PR #394, opened 2019-08-29
add aria label for categories

jenniferemshepherd:
So that the screen reader is slightly more verbose about where, for example, the link Argentina leads us, we now give it an aria-label which will say 'link, Category: Argentina'. This should make it clearer that the link takes you to ft.com/argentina.

### DIFF
--- a/components/x-teaser/src/MetaLink.jsx
+++ b/components/x-teaser/src/MetaLink.jsx
@@ -24,7 +24,7 @@ export default ({ metaPrefixText, metaLink, metaAltLink, metaSuffixText, context
 					className="o-teaser__tag"
 					data-trackable="teaser-tag"
 					href={displayLink.relativeUrl || displayLink.url}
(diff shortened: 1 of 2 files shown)


components/x-teaser/src/MetaLink.jsx
```diff
@@ -23,7 +23,8 @@ export default ({ metaPrefixText, metaLink, metaAltLink, metaSuffixText, context
 				<a
 					className="o-teaser__tag"
 					data-trackable="teaser-tag"
-					href={displayLink.relativeUrl || displayLink.url}>
+					href={displayLink.relativeUrl || displayLink.url}
+					aria-label="Category: {displayLink.prefLabel}">
```
pixelandpage: Because this is JSX we'll need the string interpolation to be slightly different here:
aria-label={`Category: ${displayLink.prefLabel}`}

The outside brackets tell the file to interpret this snippet as JavaScript which lets us use a template literal inside.

PR #397, opened 2019-09-11
Update the aria label on podcast and video teasers

pixelandpage:
Adds an aria-label property to all video and podcast teasers to include the phrase "Watch video" or "Listen to podcast" before the title attribute is read out. This will not affect other teasers as any attributes with a `null` value will be ignored.

This should be released as a new beta version: `v1.0.0-beta.17`

### DIFF
--- a/components/x-teaser/__tests__/__snapshots__/snapshots.test.js.snap
+++ b/components/x-teaser/__tests__/__snapshots__/snapshots.test.js.snap
@@ -28,7 +28,6 @@ exports[`x-teaser / snapshots renders a Hero teaser with article data 1`] = `
       className="o-teaser__heading"
     >
       <a
-        aria-label={null}
         className="js-teaser-heading-link"
         data-trackable="heading-link"
         href="#"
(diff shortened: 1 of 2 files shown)


components/x-teaser/__tests__/__snapshots__/snapshots.test.js.snap
```diff
@@ -28,6 +28,7 @@ exports[`x-teaser / snapshots renders a Hero teaser with article data 1`] = `
       className="o-teaser__heading"
     >
       <a
+        aria-label={null}
```
quarterto: nb although these are `null` in the snapshot, most (all?) DOM/HTML renderers will skip this property in the output

components/x-teaser/src/Title.jsx
```diff
@@ -12,6 +12,7 @@ export default ({ title, altTitle, headlineTesting, relativeUrl, url, indicators
 			<Link {...props} url={displayUrl} attrs={{
 				'data-trackable': 'heading-link',
 				className: 'js-teaser-heading-link',
+				'aria-label': props.type === 'video' ? `Watch video ${displayTitle}` : null
```
i-like-robots: For podcasts/audio should we also add a "listen to" prefix?

pixelandpage: Yes 👍

PR #401, opened 2019-10-03
update related links url property to relativeUrl

pixelandpage:
Replaces `url` with `relatedUrl` in the relatedLinks x-teaser component to more accurately match the data being passed in.

This change will affect the relatedLinks component only.

### DIFF
--- a/components/x-teaser/src/RelatedLinks.jsx
+++ b/components/x-teaser/src/RelatedLinks.jsx
@@ -1,11 +1,11 @@
 import { h } from '@financial-times/x-engine';
 
-const renderLink = ({ id, url, type, title }, i) => (
+const renderLink = ({ id, relativeUrl, type, title }, i) => (
 	<li
 		key={`related-${i}`}
 		data-content-id={id}
 		className={`o-teaser__related-item o-teaser__related-item--${type}`}>
-		<a data-trackable="related" href={url}>
+		<a data-trackable="related" href={relativeUrl}>
 			{title}
 		</a>
 	</li>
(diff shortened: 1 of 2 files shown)


quarterto:
This looks like a breaking change. Where is the `RelatedLinks` component used?

pixelandpage:
> Where is the RelatedLinks component used?

At the moment, nowhere 🙂 It's designed to be used for the related content links which we display in two out of our four top story layouts to go along with the lead story on ft.com. It's pulled in via the main `Teaser` component https://github.com/Financial-Times/x-dash/blob/b196dc20a09b861fce69e4e50fc77f3105f948b7/components/x-teaser/src/Teaser.jsx#L28

Screenshot 2019-10-03 at 14 32 51

quarterto:
Ah, i see :)

components/x-teaser/src/RelatedLinks.jsx
```diff
@@ -1,11 +1,11 @@
 import { h } from '@financial-times/x-engine';
 
-const renderLink = ({ id, url, type, title }, i) => (
+const renderLink = ({ id, relativeUrl, type, title }, i) => (
 	<li
 		key={`related-${i}`}
 		data-content-id={id}
 		className={`o-teaser__related-item o-teaser__related-item--${type}`}>
-		<a data-trackable="related" href={url}>
+		<a data-trackable="related" href={relativeUrl}>
```
i-like-robots: Just a note that we use either `relativeUrl` or `url`  in other places:

https://github.com/Financial-Times/x-dash/blob/master/components/x-teaser/src/Image.jsx#L30

PR #403, opened 2019-10-07
Fix words break issue by adding a white space

chivchila:
This is just fixing a small issue with no word break when two spans next to each have no newline characters in between them. By adding it - CSS can wrap non-fitting characters to the new line.

### DIFF
--- a/components/x-teaser/__tests__/__snapshots__/snapshots.test.js.snap
+++ b/components/x-teaser/__tests__/__snapshots__/snapshots.test.js.snap
@@ -14,6 +14,8 @@ exports[`x-teaser / snapshots renders a Hero teaser with article data 1`] = `
       <div
         className="o-teaser__meta-tag"
       >
+         
+         
         <a
           aria-label="Category: Sexual misconduct allegations"
           className="o-teaser__tag"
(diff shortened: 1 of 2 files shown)


chivchila:
Showing how string wrapping before and after the change
Here the very long string remains in the same line as Meta_prefix
Screenshot 2019-10-07 at 15 45 28
After the change - wraps non-fitting long word into a new line
Screenshot 2019-10-07 at 15 54 09

quarterto:
thanks @chivchila! you'll need to update the Jest snapshots by running `npm run jest -- --updateSnapshots` to get the tests passing

quarterto:
ah, i've just remembered there used to be some whitespace here (slightly differently because it was inside the link) that [was removed](https://github.com/Financial-Times/x-dash/pull/146) because it broke some stream page layouts along with the extra margins if i remember correctly. we should test this branch on some FT.com apps first (and wait until @i-like-robots is back from holiday)

i-like-robots:
Space between the teaser meta elements used to only be based on whitespace in the markup which became a small issue when we started creating components with JSX. I removed the need for HTML whitespace [in this PR](https://github.com/Financial-Times/o-teaser/pull/108) which resulted in the PR Bren linked to above.

Unfortunately as it's now based on margins adding a space can lead to a gap which is larger than intended 😕:

Screenshot 2019-10-11 at 14 53 55

So... I think we could try a zero width space or a `max-width/word-break` dance 🤔

i-like-robots:
Should we close this in favour of https://github.com/Financial-Times/o-teaser/pull/137 @chivchila ?

PR #407, opened 2019-10-23
Add imageDataUrl prop to enable apps offline functionality

willhoward:
## What is this?

This PR adds an `imageDataUrl` boolean prop to x-teaser which can be used to indicate that the image URL provided is a self-contained image and the component does not have to construct an image URL or make a request to the origami image service.

## Why is it necessary?

The apps team are currently working on the capability to store images offline and consume them in teasers directly from this offline store. In order to make this work, we need x-teaser to accept images in data-url format and not to attempt to construct a URL for the image and request it from the origami image service. The new `imageDataUrl` prop achieves this.

I sense-checked this very briefly with Maggie before beginning - please let me know if there's anybody else I should assign to review!

### DIFF
--- a/components/x-teaser/readme.md
+++ b/components/x-teaser/readme.md
@@ -171,7 +171,7 @@ Property        | Type                  | Notes
 `image`         | [media](#media-props) |
 `imageSize`     | String                | XS, Small, Medium, Large, XL or XXL
 `imageLazyload` | Boolean, String       | Output image with `data-src` attribute. If this is a string it will be appended to the image as a class name.
-`imageDataUrl`	|	Boolean								|	Signifies that the image URL is a self-contained data URL.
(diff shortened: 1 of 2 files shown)


i-like-robots:
Interesting use case!

I'm not sure about the implementation... not that there's anything wrong with it, I just wonder if it should detect such a use case automatically? e.g. `image.url.startsWith('data:') ? image.url : imageService(...)`

willhoward:
Absolutely! The only reason I went with adding a boolean prop is because I think it may be more robust, but I can change it over if you'd prefer?

i-like-robots:
Up to you I think as you'll be the main user of it!

If you do choose the flag approach it might need a bit more consideration. The implementation really means "do not use the image service" which could be useful to others, so it could be made more explicit, like `useImageService` (which defaults to `true` because flags which disable features _always_ make a mess IMO!).

willhoward:
Opted for the first solution 👍

PR #414, opened 2019-11-19
Hide podcast app launchers on desktop

alicebartlett:
Hello, this is my first foray into x-dash, so it's very possible I've done something obvious incorrectly. Please be vigilant when reviewing this PR!

We have had customer feedback that these buttons don't do anything on
desktop and it is a concern that we are losing out on opportunities
for engagement by showing people broken buttons.

This commit:
- Uses a media query to hide the app launcher buttons on desktop as
  they don't work
- Introduces an oGrid dependency to get the standard breakpoint for
  desktop

I don't know how x-dash manages Origami dependencies. This is either a patch release because it is a very small change that won't break anything, or it is major because I have introduced a new Origami dependency which could break a dependency tree somewhere. Can anyone advise on this?

✅ Discuss features first (chatted with @robsquires) 
✅ Update the documentation
✅ No hacks, experiments or temporary workarounds
- Reference other issues
- Follow the code style
? Decide on a version (major, minor, or patch)

### DIFF
--- a/components/x-podcast-launchers/src/PodcastLaunchers.scss
+++ b/components/x-podcast-launchers/src/PodcastLaunchers.scss
@@ -1,5 +1,7 @@
 $o-typography-is-silent: true;
+$o-grid-is-silent: true;
 @import 'o-typography/main';
+@import 'o-grid/main';
 
 :global {
 	@import "~@financial-times/x-follow-button/dist/FollowButton";
@@ -48,6 +50,9 @@ $o-typography-is-silent: true;
 }
 
 .podcastAppLink {
+	@include oGridRespondTo(L) {
+		display: none;
+	}
 	width: 100%;
 	margin-top: 8px;
 }
--- a/components/x-podcast-launchers/bower.json
+++ b/components/x-podcast-launchers/bower.json
@@ -4,6 +4,7 @@
   "main": "dist/PodcastLaunchers.es5.js",
   "private": true,
   "dependencies": {
-    "o-typography": "^5.12.0"
+    "o-typography": "^5.12.0",
+    "o-grid": "^4.5.3"
   }
 }


robsquires:
LGTM @alicebartlett 
Re. patch/minor, does it play nicely when you pull this branch into next-article and the app? I can try the app if that helps

alicebartlett:
@robsquires good tip Rob - I'll check that

alicebartlett:
OK - I've checked this on next-article and it looks good. @robsquires is gonna check on the app so I don't have to get all that set up, and then I think we're good!

robsquires:
@alicebartlett we've lost the buttons on a landscape ipad 😢 

Screenshot 2019-11-21 at 17 27 10

components/x-podcast-launchers/readme.md
```diff
@@ -55,5 +55,6 @@ Feature               | Type     | Required | Notes
 `acastRSSHost`        | String   | Yes      | e.g. 'https://acast.access.com'
 `conceptId`           | String   | Yes      |
 `renderFollowButton`  | Function | No       | Optional render prop for the follow button
+`isApp`               | Boolean  | No       | Defaults to true. Currently only used to hide podcast app launchers on wide screen + desktop where the links do nothing because there is no app to launch.
```
apaleslimghost: i'd prefer a more descriptive name for this property, something like `showOnAllBreakpoints`?

components/x-podcast-launchers/src/PodcastLaunchers.scss
```diff
@@ -55,10 +55,18 @@ $o-buttons-is-silent: true;
 	margin: 0 0 24px;
 }
 
+
+/*
+ Hide podcast links at a wide screen width. We only do this when the running application
+ isn't the App so that on any native device you'll always be able to launch a podcast app
+*/
+.hidePodcastLinkAtWide {
+		@include oGridRespondTo(L) {
```
apaleslimghost: this will also hide the buttons on landscape tablets, is that intentional?

components/x-podcast-launchers/readme.md
alicebartlett: Thanks bren. I was going for an easy to understand variable eg is this the app - yes or no? which one could then hang behaviours off of - in this case showing or hiding the podcast launcher buttons. 
I prefer your name and semantic change though - it's simpler. I'll implement.

components/x-podcast-launchers/src/PodcastLaunchers.scss
alicebartlett: 😳 No what I've done here is renamed the class in the CSS but not renamed it in the jsx

```
<li key={name} className={isApp ? "" : styles.hidePodcastLinkAtWide}>
```

I think this solution is a bit gross btw - so if you'd have done it a different more elegant way please tell me!

apaleslimghost: we don't really use anything other than viewport for detecting mobile anywhere else (the other comparable case i can think of is hiding the Whatsapp share icon which [uses the `S` breakpoint](https://github.com/Financial-Times/next-article/blob/cf51a94382cb9fb1b98dfb144781e456565736ca/client/components/share/_mixins.scss#L19-L22))

robsquires:
Woop! 🎯 

Screenshot 2020-01-09 at 11 25 21

components/x-podcast-launchers/src/PodcastLaunchers.jsx
```diff
@@ -48,39 +48,38 @@ class PodcastLaunchers extends Component {
 
 	render() {
 		const { rssUrl } = this.state;
-		const { conceptId, conceptName, csrfToken, isFollowed, renderFollowButton } = this.props
+		const { conceptId, conceptName, csrfToken, isFollowed, renderFollowButton, showLinksOnAllBreakpoints } = this.props
```
robsquires: As per IRL chat, default `showLinksOnAllBreakpoints` to true

robsquires:
🚢👈

components/x-podcast-launchers/src/PodcastLaunchers.jsx
```diff
@@ -48,39 +48,38 @@ class PodcastLaunchers extends Component {
...
-									data-trackable='copy-rss'
-									type='button'>
-									Copy RSS
-								</button>
-						</span>
-					</li>
-				</ul>
+					<ul className={styles.podcastAppLinksWrapper}>
+						{generateAppLinks(rssUrl).map(({ name, url, trackingId }) => (
+							<li key={name} className={showLinksOnAllBreakpoints ? "" : styles.hidePodcastLinkAtWide}>
```
apaleslimghost: it would be better to output `null` instead of empty string here. empty string would output as `class=""` whereas `null` would omit the attribute.

alicebartlett:
```suggestion
							<li key={name} className={showLinksOnAllBreakpoints ? null : styles.hidePodcastLinkAtWide}>
```

apaleslimghost (approved):
sorry this PR got forgotten about! it looks good now 🙂 @robsquires, since Alice is away would you mind shepherding this through to release?

PR #415, opened 2019-11-20
Add support for styles written with Sass to storybook

kateloschinina:
- update storybook webpack config
 - link packages using bower with the post-install script
 - init bower at the root level

Tested that works with the new x-follow-button (see this PR https://github.com/Financial-Times/x-dash/pull/59)

### DIFF
--- a/makefile
+++ b/makefile
@@ -9,6 +9,8 @@ clean:
 install:
 	npm i --no-package-lock
 	npx athloi exec -- npm i --no-package-lock
+	node scripts/link-packages-using-bower.js
+	npx	bower install
 
 build:
 	npm run build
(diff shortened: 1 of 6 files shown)


makefile
```diff
@@ -9,6 +9,8 @@ clean:
 install:
 	npm i --no-package-lock
 	npx athloi exec -- npm i --no-package-lock
+	node scripts/link-packages-using-bower.js
+	npx	bower install
```
i-like-robots: I think the bower dependency will also need to be removed from each package and the `postinstall` step amended to ensure we don't double install everything 🤔

kateloschinina: 🤔 potentially that makes sense... I will try this out

kateloschinina: Potentially - because none of the packages or components other than `x-podcast-launchers` seems to have a bower dependency

edds:
Closing this as this functionality was introduced in #59

PR #416, opened 2019-11-21
Update o-forms and o-buttons in podcast launcher

alicebartlett:
This PR updates o-forms and o-buttons in the PodCast Launcher.

As there are some big Origami changes coming next week it makes sense to upgrade these components now because they'll block the more straightforward upgrades next week.

As part of updating o-forms I realised the way o-forms was being included (declaratively, with the expectation that the consuming client will include all the component sass) is likely to lead to version conflicts that nobody notices. 

To address this, I have brought the code into the component implicitly, and am only using the necessary styles for the rss field. 

I have tested this in next-article and it works as I would expect. I don't anticipate any issues with the app as these are pretty simple changes, but if someone with the apps spun up could help me verify that, that would be good.

BEFORE
Screenshot 2019-11-21 at 15 32 52
Screenshot 2019-11-21 at 15 32 42

AFTER
Screenshot 2019-11-21 at 15 28 18
Screenshot 2019-11-21 at 15 28 25

### DIFF
--- a/__tests__/__snapshots__/snapshots.test.js.snap
+++ b/__tests__/__snapshots__/snapshots.test.js.snap
@@ -569,7 +569,7 @@ exports[`@financial-times/x-podcast-launchers renders a default Example x-podcas
   >
     <li>
       <a
-        className="o-buttons o-buttons--primary o-buttons--big PodcastLaunchers_podcastAppLink__69RMR"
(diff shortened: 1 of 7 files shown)


alicebartlett:
@quarterto welp thanks bren

PR #417, opened 2019-11-25
X podcast launcher cascade

alicebartlett:
As part of the Origami cascade, bump o-typography, o-forms and o-buttons to new majors.

### DIFF
--- a/components/x-podcast-launchers/src/PodcastLaunchers.scss
+++ b/components/x-podcast-launchers/src/PodcastLaunchers.scss
@@ -54,7 +54,10 @@ $o-buttons-is-silent: true;
 }
 
 .podcastAppLink {
-	@include oButtons(big, primary);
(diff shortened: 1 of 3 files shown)


components/x-podcast-launchers/src/PodcastLaunchers.scss
```diff
@@ -16,18 +16,18 @@ $o-buttons-is-silent: true;
...
 .headingNoApp {
-	@include oTypographySansBold($scale: 0);
+	@include oTypographySans($weight: 'semibold', $scale: 0);
 	grid-area: heading;
 	margin: 0;
 }
 
 .textNoApp {
-	@include oTypographySans($scale: 0);
+	@include oTypographySans();
```
notlee: This needs `$scale: 0` to output the font size and line height.
By default it outputs only the font-family.

PR #419, opened 2019-11-25
Modify x-teaser image to support blob: URLs for images without trying to pass them to the image service

rowanbeentje:
To improve speed of loading images in the app, we'd like to switch to `blob:` URLs generated via `URL.createObjectURL()`. At the moment x-teaser doesn't recognise these and so tries to load them via the image service; this is a simple change just adding them to the existing exception that exists for `data:` URLs.

I've added a test for the `data:` uri functionality just to ensure that wasn't broken - thought it was superfluous to add a similar snapshot for a `blob:` url but happy to do so if you'd like...

### DIFF
--- a/components/x-teaser/__fixtures__/article-with-data-image.json
+++ b/components/x-teaser/__fixtures__/article-with-data-image.json
@@ -1,0 +1,30 @@
+{
+  "type": "article",
+  "id": "",
+  "url": "#",
+  "title": "Inside charity fundraiser where hostesses are put on show",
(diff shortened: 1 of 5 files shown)


rowanbeentje:
(This is not nearly as scary as it looks at first glance, it's all snapshot additions, honest!)

rowanbeentje:
CircleCI status not reporting correctly here, likely due to https://status.circleci.com/incidents/vx1kfsmgjyvp , but all good on https://circleci.com/gh/Financial-Times/workflows/x-dash/tree/x-teasers-support-blobs .

PR #420, opened 2019-11-25
Origami cascade

alicebartlett:
This PR updates o-forms, o-message, o-buttons, o-typography and o-loading in x-gift-article.

These components were being required implicitly. This is bad because it means we can't do any programatic detection of version conflicts between dependencies, and we have no build step to check that a version bump in the consuming application isn't going to break the styles of the x-dash component.

Now, x-gift-article provides all of it's own styles using the mixins from various o-components.

I had to do quite a bit of rewriting to get that working which is why this PR is big :(

Happy to break this PR down into smaller ones if that's easier for reviewers. 

I have checked this in n-article and it looks good. It should be checked in the App too.


components/x-gift-article/bower.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "@financial-times/x-gift-article",
+  "description": "",
+  "main": "dist/GiftArticle.es5.js",
+  "private": true,
+  "dependencies": {
+    "o-buttons": "^6.0.0",
+    "o-loading": "^4.0.0"
+  }
```
i-like-robots: I think it may be feasible to make these npm dev dependencies and do away with Bower entirely for this component 🤔

alicebartlett: Hi Matt, sorry, could we have a chat about that, I'm not sure I understand this approach

alicebartlett: As discussed, let's not optimise this right now, as we would need to also update the JavaScript (which is being included onto the page)

components/x-gift-article/readme.md
```diff
@@ -78,7 +77,7 @@ Property                  | Type    | Required | Note
 --------------------------|---------|----------|----
 `isFreeArticle`           | Boolean | yes      | Only non gift form is displayed when this value is `true`.
 `article`                 | Object  | yes      | Must contain `id`, `title` and `url` properties
-`showMobileShareLinks`    | Boolean | no       |
+`showMobileShareLinks`    | Boolean | no       | For ft.com on mobile sharing.
```
chee: :+1::+1::+1::+1:

components/x-gift-article/src/RadioButtonsSection.jsx
```diff
@@ -1,42 +1,44 @@
...
-			onChange={ showNonGiftUrlSection }/>
-
-		<label htmlFor="nonGiftLink" className="o-forms__label">
-			with <span className={ boldTextClassName }>other FT subscribers</span>
+		<label htmlFor="nonGiftLink">
+			<input
+				type="radio"
+				name="gift-form__radio"
+				value="nonGiftLink"
+				className="o-forms__radio"
```
i-like-robots: Should this be `styles['o-forms__radio']`?

components/x-gift-article/src/Loading.jsx
```diff
@@ -1,8 +1,8 @@
 import { h } from '@financial-times/x-engine';
-import styles from './GiftArticle.css';
+import styles from './GiftArticle.scss';
```
chee: good

components/x-gift-article/src/MobileShareButtons.jsx
```diff
@@ -2,70 +2,31 @@ import { h } from '@financial-times/x-engine';
...
-	'o-share__icon',
-	'o-share__icon--linkedin',
-	styles.linkedin
-].join(' ');
-
-const whatsappClassNames = [
-	'o-share__icon',
-	'o-share__icon--whatsapp',
-	styles.whatsapp
-].join(' ');
```
chee: +++

components/x-gift-article/src/lib/variables.scss
```diff
@@ -0,0 +1,3 @@
+// This is needed for calls to the image service for Icons used in Social,
+// and oMessage
+$system-code:'github:Financial-Times/x-dash';
```
chee: is there a way we can set this only be included in the tests, so that apps don't forget to set it?

alicebartlett: Had a look at this but I can't work out how to do it such that it works for storybook and the tests. Going to open an issue for this and move on as I agree it is the right thing to do but it doesn't need to block this PR.

components/x-gift-article/src/Form.jsx
```diff
@@ -4,17 +4,18 @@ import RadioButtonsSection from './RadioButtonsSection';
 import UrlSection from './UrlSection';
 import MobileShareButtons from './MobileShareButtons';
 import CopyConfirmation from './CopyConfirmation';
-import styles from './GiftArticle.css';
+import styles from './GiftArticle.scss';
 
 const formClassNames = [
-	'o-forms',
-	styles.form
+	styles["share-form"]
```
edds: Stylistic point feel free to ignore, but now this is just one class I'd just update line 15 below to be `className={styles['share-form']}` rather than having an intermediary variable.

alicebartlett: Ah yes - I had done that elsewhere but missed this

PR #421, opened 2019-11-27
add a showGuidance data-attribute to the Video.jsx template

pixelandpage:
Adds a `data-o-video-show-guidance` data-attribute to the Video.jsx template which defaults to `true`. 

This will allow us to disable the 'No captions available' guidance for videos which use x-teaser. We want to remove this guidance  for videos on the Front Page as tests have shown that it decreases engagement; [discussion in this trello ticket](https://trello.com/c/qSiuMpf4/110-remove-subtitles-info-from-video-promo-on-ft-homepage). 

Related PR in o-video to support the new attribute: https://github.com/Financial-Times/o-video/pull/149
Related PR to implement this change in next-front-page: https://github.com/Financial-Times/next-front-page/pull/2208

### DIFF
--- a/components/x-teaser/src/Video.jsx
+++ b/components/x-teaser/src/Video.jsx
@@ -10,7 +10,7 @@ const formatData = (props) => JSON.stringify({
 // To prevent React from touching the DOM after mounting… return an empty <div />
 // <https://reactjs.org/docs/integrating-with-other-libraries.html>
 const Embed = (props) => {
-	const showGuidance = typeof(props.showGuidance) === 'boolean' ? props.showGuidance : true;
+	const showGuidance = typeof props.showGuidance  === 'boolean' ? props.showGuidance.toString() : "true";
 	return (
 	<div className="o-teaser__image-container js-image-container">
 		<div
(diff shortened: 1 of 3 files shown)


components/x-teaser/src/Video.jsx
```diff
@@ -20,6 +20,7 @@ const Embed = (props) => (
 			data-o-video-optimumvideowidth="640"
 			data-o-video-autorender="true"
 			data-o-video-playsinline="true"
+			data-o-video-disableGuidance={props.disableGuidance}
```
i-like-robots: this needs adding to the docs

pixelandpage:
After a chat with the Origami team, I've refactored this to use a `showGuidance` rather than `disableGuidance`.

i-like-robots:
Great! It's always better to _opt-in_ rather than opt-out.

components/x-teaser/src/Video.jsx
```diff
@@ -20,12 +22,14 @@ const Embed = (props) => (
 			data-o-video-optimumvideowidth="640"
 			data-o-video-autorender="true"
 			data-o-video-playsinline="true"
+			data-o-video-show-guidance={showGuidance}
```
carlesandres: Maybe we should make `showGuidance` a string to make it consistent with the other data attributes.

PR #422, opened 2019-11-28
Origami cascade x teaser

alicebartlett:
This PR fixes the one issue I've found with the latest Origami versions, and bumps the dependencies in Storybook.

The one problem I came across is video no longer working as it required a system code. I have added this as a prop, and set it in storybook, and documented it.

### DIFF
--- a/components/x-teaser/src/Video.jsx
+++ b/components/x-teaser/src/Video.jsx
@@ -22,7 +22,8 @@ const Embed = (props) => (
 			data-o-video-playsinline="true"
 			data-o-video-placeholder="true"
 			data-o-video-placeholder-info="[]"
-			data-o-video-placeholder-hint="Play video" />
+			data-o-video-placeholder-hint="Play video"
+			data-o-video-systemcode={props.systemCode} />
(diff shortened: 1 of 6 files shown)


components/x-teaser/src/Video.jsx
```diff
@@ -22,7 +22,8 @@ const Embed = (props) => (
 			data-o-video-playsinline="true"
 			data-o-video-placeholder="true"
 			data-o-video-placeholder-info="[]"
-			data-o-video-placeholder-hint="Play video" />
+			data-o-video-placeholder-hint="Play video"
+			data-o-video-systemcode={props.systemCode} />
```
quarterto: this prop should be documented

chee: how do you document props in this library?

chee: oh, https://github.com/Financial-Times/x-dash/pull/422/files#diff-0ffeafa06b86f5b823888944c54c4db8R188-R192

^_^ cool

PR #423, opened 2019-11-29
Remove unused demo

alicebartlett:
This demo was broken by a change to x-handlebars and nobody noticed.

Chatted it through with Bren and rather than fix it and then update the
demo we decided we could probably delete this demo entirely as it's
no longer needed.


quarterto:
my snyk account is fucked, so somebody else will have to delete that package

alicebartlett:
I have deleted the snyk package now.

PR #424, opened 2019-12-05
Output 2x images at lower lossy quality

quarterto:
small teaser images with text look blurry on retina-quality screens:

image

this increases the size of the image, but reduces the quality. in testing this increases the size of the image, but by an insignificant amount (<5%).

i'm not using `srcset`, because it's my intuition that the size increases from including every image URL twice on a page would offset the size reduction from not loading the larger images on every client, and increasing the size of a (probably lazy-loaded) subresource will have less of an impact than increasing the size of the initial page.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,6 @@
     "build": "athloi run build --concurrency 3 && npm run build-storybook",
     "build-only": "athloi run build",
     "jest": "jest -c jest.config.js",
-    "pretest": "npm run build",
     "test": "npm run lint && npm run jest",
     "lint": "eslint . --ext=js,jsx",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -5,7 +5,6 @@
     "build": "athloi run build --concurrency 3 && npm run build-storybook",
     "build-only": "athloi run build",
     "jest": "jest -c jest.config.js",
-    "pretest": "npm run build",
```
quarterto: a minor QOL improvement that i can split out if needed. building every component before test is arduous when iterating on a single component. this lets you do e.g. `make build-x-teaser test`

PR #426, opened 2019-12-09
Add a disable guidance attribute to videos

rowanmanning:
This reapplies @pixelandpage's work to fix video subtitles to a new branch. The `pre-origami-cascade` branch was created from `v1.0.0-beta.21`, and new beta releases should be created from here.

There's definitely some stuff that's slightly wrong about this PR, mostly whitespace, which I'll look into now. I'm going to need help to get this merged and released in a way that doesn't break things.

### DIFF
--- a/components/x-teaser/__tests__/__snapshots__/snapshots.test.js.snap
+++ b/components/x-teaser/__tests__/__snapshots__/snapshots.test.js.snap
@@ -1697,6 +1697,7 @@ exports[`x-teaser / snapshots renders a HeroVideo teaser with video data 1`] = `
             data-o-video-placeholder-hint="Play video"
             data-o-video-placeholder-info="[]"
             data-o-video-playsinline="true"
+            data-o-video-show-guidance={true}
           />
         </div>
       </div>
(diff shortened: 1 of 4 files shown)


components/x-teaser/__tests__/__snapshots__/snapshots.test.js.snap
```diff
@@ -202,7 +202,7 @@ exports[`x-teaser / snapshots renders a Hero teaser with packageItem data 1`] =
         data-trackable="teaser-tag"
         href="#"
       >
-        Financial crisis: Are we safer now? 
+        Financial crisis: Are we safer now?
```
quarterto: these might have been removed by autoformatting or something? i think they're supposed to be there because that's the markup that's actually generated. tenner says the snapshot tests fail

rowanmanning: Yep I'm on this

quarterto:
ah, tests have failed but not for that reason. think you'll need to backport https://github.com/Financial-Times/x-dash/pull/412 as well

PR #428, opened 2019-12-30
Adding github settings.yml

edds:
This lets us centrally define access permissions to our repositories
and other sensible defaults where approrpriate. You can see the
defaults in the [github-apps-config-next][defaults] repo. For full
configuration options check out the [probot settings repo][probot].

[defaults]: https://github.com/Financial-Times/github-apps-config-next/blob/master/.github/settings.yml
[probot]: https://github.com/probot/settings

🤖 This PR was opened automatically.

### DIFF
--- a/.github/settings.yml
+++ b/.github/settings.yml
@@ -1,0 +1,17 @@
+_extends: github-apps-config-next
+
+branches:
+  - name: master
+    protection:
+      required_pull_request_reviews:
+        required_approving_review_count: 1
+        dismiss_stale_reviews: true
+        require_code_owner_reviews: false
+      required_status_checks:
+        strict: true
+        contexts:
+          - "ci/circleci: test"
+      enforce_admins: false
+      restrictions:
+        users: []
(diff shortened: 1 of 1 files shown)


.github/settings.yml
```diff
@@ -0,0 +1,17 @@
...
+  - name: master
+    protection:
+      required_pull_request_reviews:
+        required_approving_review_count: 1
+        dismiss_stale_reviews: true
+        require_code_owner_reviews: false
+      required_status_checks:
+        strict: true
+        contexts:
+          - "ci/circleci: test"
```
i-like-robots: In https://github.com/Financial-Times/dotcom-page-kit/pull/701 there are additional contexts could also apply to this.

edds: You should only need test as it is dependent on `build` passing for it to be run. We can add the others for belts and braces though if you'd like.

i-like-robots: That makes sense

PR #430, opened 2020-01-15
Extract logic from x-teaser-timeline and add tests

alexandercurtis:
This PR does not change behaviour or appearance, simply a small refactor in preparation for fixing this bug: https://trello.com/c/tTPiw9x1/3976-topic-bubbles-fix-latest-news-problem-on-stream-pages

It collects together the presentation logic in teaser-timeline into a single entry point `buildModel`, and puts tests around that function, with the aim that `buildModel` can then be refactored safely without changing its behaviour.

### DIFF
--- a/components/x-teaser-timeline/src/TeaserTimeline.jsx
+++ b/components/x-teaser-timeline/src/TeaserTimeline.jsx
@@ -6,11 +6,19 @@ import styles from './TeaserTimeline.scss';
 import classNames from 'classnames';
 
 const TeaserTimeline = props => {
+	const now = new Date();
 	const {
 		csrfToken = null,
-		showSaveButtons = true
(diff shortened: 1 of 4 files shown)


components/x-teaser-timeline/src/TeaserTimeline.jsx
```diff
@@ -1,29 +1,25 @@
...
+	const now = new Date();
 	const {
 		csrfToken = null,
+		showSaveButtons = true,
 		customSlotContent,
 		customSlotPosition = 2,
 		items,
-		showSaveButtons = true
+		timezoneOffset = now.getTimezoneOffset(),
+		localTodayDate = getDateOnly(now.toISOString()),
```
robsquires: Nice moving these to default props!

PR #435, opened 2020-01-24
Allow Latest News to include articles up to 36h before today

robsquires:
This PR has been recreated from https://github.com/Financial-Times/x-dash/pull/431

Component: x-teaser-timeline

This lets the Latest News section of this stream include news from a configurable time range. Previously it would only include news from the same calendar day.

If latestItemsAgeHours prop is specified, it is used as a maximum age after which articles are no longer considered "latest".

If latestItemsAgeHours is missing, previous behaviour continues, where articles before today are never considered latest.

Trello: https://trello.com/c/tTPiw9x1/3976-topic-bubbles-fix-latest-news-problem-on-stream-pages

Why: Previously, the myft bubbles on the ft-app would have rings around them indicating new content, but clicking them went to a list with no "Latest Articles". When this code is hooked in, there should be some latest articles shown.

Note that, in spite of the branch name, I haven't done any significant refactoring. I recommend this is done at some point though, since this code was a bit smelly and I haven't done anything to freshen it up.


components/x-teaser-timeline/src/lib/transform.js
```diff
@@ -67,15 +50,27 @@ const addItemGroupTitles = (itemGroups, localTodayDate) => {
...
 
-const getItemGroups = ({items, timezoneOffset, localTodayDate, latestItemsTime}) => {
-	if (!items || !Array.isArray(items) || items.length === 0) {
+const getItemGroups = ({sortedItems, timezoneOffset, localTodayDate, latestItemsTime, latestItemsAgeHours}) => {
+	if (!sortedItems || !Array.isArray(sortedItems) || sortedItems.length === 0) {
 		return [];
 	}
 
-	let itemGroups = groupItemsByLocalisedDate(items, timezoneOffset);
+	const isLatestItemTimeWithinRange = latestItemsAgeHours ?
```
robsquires: Might just be me but find this a little tricky to grok, is there a way of making these rules a bit more explicit?

alexandercurtis: @robsquires I've refactored a bit and added a comment. Please have another look.

components/x-teaser-timeline/src/lib/transform.js
```diff
@@ -50,18 +50,41 @@ const addItemGroupTitles = (itemGroups, localTodayDate) => {
...
-const getItemGroups = ({sortedItems, timezoneOffset, localTodayDate, latestItemsTime, latestItemsAgeHours}) => {
-	if (!sortedItems || !Array.isArray(sortedItems) || sortedItems.length === 0) {
+const isTimeWithinAgeRange = (time, localTodayDate, ageRangeHours) =>
+	time && (new Date(localTodayDate) - new Date(time)) < (ageRangeHours * 60 * 60 * 1000);
+
+const isTimeWithinToday = (time, localTodayDate) =>
+	time && (getDateOnly(localTodayDate) === getDateOnly(time));
+
+/**
+ * Groups items (articles) by date
```
robsquires: Nice!

components/x-teaser-timeline/src/lib/transform.js
```diff
@@ -50,18 +50,41 @@ const addItemGroupTitles = (itemGroups, localTodayDate) => {
...
+ */
+const getItemGroups = ({items, timezoneOffset, localTodayDate, latestItemsTime, latestItemsAgeHours: latestItemsAgeRange}) => {
+	if (!items || !Array.isArray(items) || items.length === 0) {
 		return [];
 	}
 
-	const isLatestItemTimeWithinRange = latestItemsAgeHours ?
-		latestItemsTime && (new Date(localTodayDate)-new Date(latestItemsTime)) < (latestItemsAgeHours * 60 * 60 * 1000) :
-		latestItemsTime && (getDateOnly(localTodayDate) === getDateOnly(latestItemsTime));
+	const sortedItems = items ? [...items].sort( (a,b) => a.publishedDate > b.publishedDate ? -1 : 1 ) : [];
```
robsquires: Is the check on items required here?

components/x-teaser-timeline/src/lib/transform.js
```diff
@@ -50,18 +50,41 @@ const addItemGroupTitles = (itemGroups, localTodayDate) => {
...
+const isTimeWithinToday = (time, localTodayDate) =>
+	time && (getDateOnly(localTodayDate) === getDateOnly(time));
+
+/**
+ * Groups items (articles) by date
+ *
+ * Takes an array of article items and groups them into sections by date.
+ * Gives the groups presentable titles, e.g. "Earlier Today", and "Yesterday".
+ * Will try to create a "Latest News" section if `latestItemsTime` is specified and there are articles within the
+ * permitted age range. This range can be set using `latestItemsAgeRange`.
```
robsquires: Would it be worth including the `or since midnight` logic here?

components/x-teaser-timeline/src/lib/transform.js
```diff
@@ -50,18 +50,41 @@ const addItemGroupTitles = (itemGroups, localTodayDate) => {
...
+	if (!items || !Array.isArray(items) || items.length === 0) {
 		return [];
 	}
 
-	const isLatestItemTimeWithinRange = latestItemsAgeHours ?
-		latestItemsTime && (new Date(localTodayDate)-new Date(latestItemsTime)) < (latestItemsAgeHours * 60 * 60 * 1000) :
-		latestItemsTime && (getDateOnly(localTodayDate) === getDateOnly(latestItemsTime));
+	const sortedItems = items ? [...items].sort( (a,b) => a.publishedDate > b.publishedDate ? -1 : 1 ) : [];
+
+	const includeLatestItemsSection = latestItemsAgeRange ?
```
robsquires: I reckon it might be worth taking this refactor one step further,  but totally your call. If `includeLatestItemsSection` was a function I think it would reduce the overhead in understanding what getItemGroups is doing. Would also be able to move the docblock explaining how it works to the function and add that `or since midnight` bit I mentioned above.

robsquires:
Couple comments otherwise looking sharp!

components/x-teaser-timeline/src/lib/transform.js
alexandercurtis: Done. Please have another squizz

robsquires:
🎯 💯 👌

components/x-teaser-timeline/src/lib/transform.js
```diff
@@ -67,15 +50,64 @@ const addItemGroupTitles = (itemGroups, localTodayDate) => {
...
+ */
+const getItemGroups = ({items, timezoneOffset, localTodayDate, latestItemsTime, latestItemsAgeHours: latestItemsAgeRange}) => {
 	if (!items || !Array.isArray(items) || items.length === 0) {
 		return [];
 	}
 
-	let itemGroups = groupItemsByLocalisedDate(items, timezoneOffset);
+	const sortedItems = [...items].sort( (a,b) => a.publishedDate > b.publishedDate ? -1 : 1 );
+
+	const includeLatesNewsSection = isLatestNewsSectionAllowed(localTodayDate, latestItemsTime, latestItemsAgeRange);
```
fenglish: Is this correct name?  Does it need extra `t`? 
`includeLatesNewsSection`  =>`includeLatestNewsSection`?

PR #450, opened 2020-05-13
Update CircleCi Node.js version to 10.13

nickcolley:
Gatsby has made a breaking change in a minor version so we need to update to avoid the build breaking.

https://github.com/gatsbyjs/gatsby/pull/22400#issuecomment-627911322

I have ran this Node.js version locally and ran the various npm scripts and they succeed, so I think we may be OK to upgrade without any changes.

Closes https://github.com/Financial-Times/x-dash/pull/442

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ references:
   container_config_node8: &container_config_node8
     working_directory: ~/project/build
     docker:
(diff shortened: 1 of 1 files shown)


nickcolley:
@andygout just trying to mitigate risk on a project a) I don't understand b) I don't believe our team is going to be actively working on this quarter.

PR #453, opened 2020-05-14
Fix circleci dependency caching

nickcolley:
Sorry about the noise with these pull requests.

The master build is currently failing as node-sass binary has been built for Node.js version 8.

The new environment is Node.js version 10, so my assumption is that the dependency cache needs updating which will prompt CircleCi to build the correct node-sass binary.

As noted by @andygout we should be moving towards Dart sass in the future which does not require native binaries but for now I'm aiming to get master passing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,24 +22,24 @@ references:
   #
   cache_keys_root: &cache_keys_root
     keys:
-      - cache-root-v1-{{ .Branch }}-{{ checksum "./package.json" }}
+      - cache-root-v2-{{ .Branch }}-{{ checksum "./package.json" }}
 
   cache_keys_docs: &cache_keys_docs
     keys:
-      - cache-docs-v1-{{ .Branch }}-{{ checksum "./web/package.json" }}
+      - cache-docs-v2-{{ .Branch }}-{{ checksum "./web/package.json" }}
 
   #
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -5,7 +5,7 @@ references:
   #
   # Workspace
   #
-  container_config_node8: &container_config_node8
+  container_config_node: &container_config_node
```
andygout, line 8: Do we need to specify version 10 with these (and the others in this file) if we have specified `- image: circleci/node:10.13`? E.g.

`container_config_node10: &container_config_node10`

It might also be worth changing `- image: circleci/node:10.13` to `- image: circleci/node:10.x`.

nickcolley: It seemed to be that `container_config_node` was just a name to reference the workspace and only the image number mattered. I'm new to CircleCI though so I could be wrong.

andygout: This is easily possible, which makes we wonder why we include the version number elsewhere… 🤔

nickcolley: I'll update to follow the convention.

nickcolley:
I ran a build that re-runs the build and it passed so I'm hoping that this has resolved the issue. 🙏

Issue #449, opened 2020-05-13
CCPA cookie consent manager


oliverturner:
## Description

As part of the work to achieve CCPA compliance the Privacy & Ads team has been asked to create a component to manage users' consent to their data for programmatic, demographic and behavioural advertising in a unified format per the requirements of the legislation.

This proposal is for a component that is available to FT.com, can be linked to from Specialist Titles, or rendered directly by the FT App.

## Design

![Screenshot 2020-05-13 at 09 18 11](https://user-images.githubusercontent.com/21795/81790310-249b0f80-94fd-11ea-9158-2bcb74eb7c50.png)

apaleslimghost:
thanks Oliver! a few questions:

- is there an existing implementation anywhere, or is that a mockup?
- what state and behaviour does this component have? it looks like it's submitting a form in the background and displaying a success message?
- since this component will presumably only work while online, could the mobile Apps instead link to the FT.com implementation, rather than rendering it directly?
- will this component contain its own styling, or is it Origami-only?

oliverturner:
Great questions!

- I started a quick spike to see how building the component in x-dash would be like: that's a screenshot of a story
- Re state and behaviour, spot on! That's exactly what it does, so I've wrapped the component in `withActions`
- Re linking to FT.com: the thinking behind rendering directly on FT App is that it simplifies cookie handling
- Re styling: the component uses custom CSS modules with Origami styles

oliverturner:
Also worth noting that we're planning to unify the cookie consent management apps, so this component would be a great fit for future development.

apaleslimghost:
yes, please push the branch and open a PR. sounds like this is a good fit for a component ✅

PR #456, opened 2020-05-18
x-privacy-manager

oliverturner:
This PR is a first step towards adding the CCPA Privacy Manager referenced in https://github.com/Financial-Times/x-dash/issues/449 and is intended to give maintainers an early look at the proposal

Smaller, incremental PRs will be used to complete the following features:
- [ ] Sending data to the Consent Proxy 
- [ ] Collaborating with FT Core to accommodate their needs
- [ ] Setting a Consent cookie in the case of logged-out users


nickcolley:
Heya @oliverturner :)

Let us (Platforms) know what feedback / review you'd like on this and when you're ready for it.

oliverturner:
Hola @nickcolley! I'd love some preliminary feedback on any glaring errors or omissions: creating an x-dash component is a new experience for me!

I suspect I've omitted a flag to skip snapshot testing in CI (where snapshots aren't produced, so that test fails), so that would be a v helpful start :)

components/x-privacy-manager/src/radio-btn.scss
```diff
@@ -0,0 +1,83 @@
...
+	background-color: oColorsByName('white');
+
+	.input:checked + & {
+		background-color: oColorsByName('teal');
+		color: oColorsByName('white');
+	}
+	
+	.input:focus + & {
+		outline: 2px solid #7aacfe;
+		outline: 5px auto -webkit-focus-ring-color;
```
nickcolley, line 42: You could consider a comment explaining your reasoning for this.

components/x-privacy-manager/src/radio-btn.jsx
```diff
@@ -0,0 +1,31 @@
...
+				value={value}
+				checked={checked}
+				onChange={onChange}
+			/>
+			<label htmlFor={id} className={s.label}>
+				<span className={s.label__text}>
+					{children}
+				</span>
+
+				<svg className={s.label__icon} viewBox="0 0 36 36">
```
nickcolley, line 24: If this image is used for presentational purposes only, you should consider using `aria-hidden="true"` to ensure it does not impact the accessible name for the label.

You should also consider using `focusable="false"` to prevent Internet Explorer from adding this image to the tab order.

https://stackoverflow.com/questions/29393144/how-to-prevent-svg-elements-from-gaining-focus-with-tabs-in-ie11

components/x-privacy-manager/src/privacy-manager.jsx
```diff
@@ -0,0 +1,115 @@
...
+		};
+	}
+}));
+
+/**
+ * @param {boolean} isLoading
+ * @param {_Response} response
+ * @param {string} referrer
+ */
+function renderMessage(isLoading, response, referrer) {
```
nickcolley, line 43: Forgive me if I've not seen how this happens...

When a screen reader user submits this form, how do they know that a loading or response message has been shown?

If it's not already communicated, you can consider setting focus to the element or something like `aria-live`, but the best method will depend on more context than I have. Can catchup with you more if needed.

components/x-privacy-manager/src/privacy-manager.jsx
```diff
@@ -0,0 +1,115 @@
...
+export function BasePrivacyManager({
+	consent = true,
+	referrer,
+	actions,
+	isLoading,
+	_response = undefined
+}) {
+	return (
+		<div className={s.consent}>
+			<h1 className={s.consent__title}>Do Not Sell My Personal Information</h1>
```
nickcolley, line 71: I'm making an assumption that given this is a component that it may be within a page itself. Is there any risk that a h1 will already be on the page?

components/x-privacy-manager/src/privacy-manager.jsx
```diff
@@ -0,0 +1,115 @@
...
+						<RadioBtn value="true" checked={consent === true} onChange={actions.onConsentChange}>
+							<strong>Allow</strong>
+							<span>See personalised adverts</span>
+						</RadioBtn>
+						<RadioBtn value="false" checked={consent === false} onChange={actions.onConsentChange}>
+							<strong>Block</strong>
+							<span>Only see generic adverts</span>
+						</RadioBtn>
+					</div>
+					<button className={s.form__submit} type="submit" disabled={isLoading}>
```
nickcolley, line 104: Similar to the thoughts about communicating the loading and response above. Is there a possibility that a user on a slow connection may encounter a disabled button and not understand why they cannot interact with it?

components/x-privacy-manager/src/radio-btn.scss
oliverturner: Solid suggestion 👍

components/x-privacy-manager/src/radio-btn.jsx
oliverturner: Nice catch & thank you for the link!

components/x-privacy-manager/src/privacy-manager.jsx
oliverturner: Great question, but in this instance the component is the entirety of the page and the h1's landmarking is accurate here. 
Since this marks our first steps towards a unified Privacy module, and lots of further evolution, it's definitely a point to keep in mind!

components/x-privacy-manager/src/privacy-manager.jsx
oliverturner: A loading notification is displayed immediately, and once that gets upgraded with your excellent suggestion of an `aria-live` value, all users should be kept informed of the state of the app.

(On reflection though, rather than disable the button I think I'd rather simply ignore repeat submissions made when a request is in flight)

oliverturner:
This is excellent feedback: really appreciate the close reading & constructive suggestions 👍

> It'd be great to get this running locally and give it a poke

The recommended method is to clone the X-Dash repo, follow the [set-up instructions](https://financial-times.github.io/x-dash/docs/get-started/installation) and run the bundled storybook. 

> if it's not too difficult could you explain how you're testing this?
I've treated the stories as a form of integration testing, with Jest/Puppeteer providing unit test coverage. 
You can run the tests for this module independently by running `jest -c jest.config.js "-u" "x-privacy"` in the project root

components/x-privacy-manager/src/privacy-manager.jsx
oliverturner: That's an excellent point. I think `aria-live` is probably the right approach here.

I also think that a note on the Origami & X-Dash docs helping developers to be mindful of this scenario would be a fine PR for me to follow up with.

nickcolley: I'd be interested in exploring this more with you. First question I'd ask is does this form need to be submitted via JavaScript or could there be a full page reload. Let me know.

oliverturner: Unfortunately the App UI is stateful

components/x-privacy-manager/.npmignore
```diff
@@ -0,0 +1,3 @@
+src/
+stories/
+rollup.js
```
andygout, line 3: So if this file excludes the mentioned directories from getting included in the package when it is published, presumably the code that does get published is that in the `dist` directory?

components/x-privacy-manager/readme.md
```diff
@@ -0,0 +1,46 @@
+# x-privacy-manager
+
+This module creates an interface giving users the ability to give or withhold consent to the sale of their data under the provisions of the CCPA (California Consumer Protection Act), as a first step towards the FT's journey towards a Unified Privacy solution.
+
+It is intended for use with Page Kit on FT.com, as a component within FT App, and as a brandable page linked to by Specialist Titles.
```
andygout, line 5: > as a brandable page linked to by Specialist Titles.

How can it be served as a page?

components/x-privacy-manager/src/messages.jsx
```diff
@@ -0,0 +1,78 @@
...
+ * Potentially a Specialist Title, FT.com or FT App
+ *
+ * @param {string} referrer
+ */
+function renderReferrerLink(referrer) {
+	let url;
+
+	try {
+		url = new URL(`https://${referrer}`);
+	} catch (err) {
```
andygout, line 15: Worth including the `err` parameter if unused?

components/x-privacy-manager/src/privacy-manager.jsx
```diff
@@ -0,0 +1,119 @@
...
+			const body = JSON.stringify({
+				demographic: consent,
+				behavioural: consent,
+				programmatic: consent
+			});
+
+			try {
+				const res = await fetch(CONSENT_API, { method: 'PATCH', body });
+				return { _response: { ok: res.ok } };
+			} catch (err) {
```
andygout, line 33: Worth including the `err` parameter if unused?

components/x-privacy-manager/src/privacy-manager.jsx
```diff
@@ -0,0 +1,119 @@
...
+					the collection of data by advertisers who advertise on our Sites for the purposes of
+					measuring the effectiveness of their advertising is caught, and requires the ability to
+					object to the use of your personal information. The data collected by advertisers may
+					include online identifiers, such as IP address, and interactions with advertisements.
+				</p>
+				<p>
+					Our advertising Terms and Conditions limit the use of this data by advertisers only to
+					their own business purposes, including analytics and attribution, not for commercial
+					purposes.
+				</p>
```
andygout, line 86: I think I have seen the `next-fow-api` mentioned in relation to this work. Is it a TODO to get this text from there?

components/x-privacy-manager/.npmignore
oliverturner: Spot on.
The compiled output is referenced explicitly by the following keys in `package.json`:
- `main`
- `module`
- `style`

components/x-privacy-manager/readme.md
oliverturner: I'll amend the wording, but the intention is that it can be consumed by sites outside of ft.com:

> "It is intended for use with Page Kit on FT.com, as a component within FT App, and ~as a~ on standalone pages (e.g. as linked to by Specialist Titles)"

components/x-privacy-manager/src/messages.jsx
oliverturner: I'll use `_` instead 👍

components/x-privacy-manager/src/privacy-manager.jsx
oliverturner: Yes, but we may leave this hardcoded this for the time being if it helps us meet our July 1st deadline

PR #459, opened 2020-06-04
Added first useful version of consent-proxy request

carlesandres:
Added integration with `next-consent-proxy` for enhanced experience only
Further work in the tests is needed

### DIFF
--- a/components/x-privacy-manager/src/__tests__/privacy-manager.test.jsx
+++ b/components/x-privacy-manager/src/__tests__/privacy-manager.test.jsx
@@ -46,7 +46,7 @@ function checkPayload(opts, expected) {
 	return worked;
 }
 
-const essentialProps = {
+const constantProps = {
 	consentProxyEndpoints: {
 		createOrUpdateRecord: TEST_CONSENT_URL
 	},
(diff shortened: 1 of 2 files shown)


components/x-privacy-manager/src/__tests__/privacy-manager.test.jsx
```diff
@@ -146,7 +198,10 @@ describe('x-privacy-manager', () => {
 		it('On receiving a response with a non-200 status', () => {
 			const _response = { ok: false, status: 400 };
 			const props = { ...defaultProps, _response };
-			const subject = mount(<BasePrivacyManager {...props} />);
+			const subject = mount(<BasePrivacyManager
+				{...constantProps}
+				{...props}
+			/>);
```
oliverturner, line 205: Sharing this as a purely personal pref, but In general I find that keeping props calculations in the same place involves less context switching between JS/JSX idioms

```tsx
const _response = { ok: false, status: 400 };
const props = { ...defaultProps, ...constantProps, _response };

const subject = mount(<BasePrivacyManager {...props} />);
```

carlesandres:
@oliverturner Nice spot! I overlooked that. I've fixed it now.

PR #463, opened 2020-06-09
Update typings to reflect revised callback signatures

oliverturner:
Running registered callbacks on both success and error responses from the server requires a change to the typings to accept an error message as the first argument (following the Node convention)

### DIFF
--- a/components/x-privacy-manager/src/privacy-manager.jsx
+++ b/components/x-privacy-manager/src/privacy-manager.jsx
@@ -1,5 +1,7 @@
 /// <reference path="./types.d.ts" />
 
+/* eslint no-console: "off" */
+
 import { h } from '@financial-times/x-engine';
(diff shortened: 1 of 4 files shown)


components/x-privacy-manager/src/privacy-manager.jsx
```diff
@@ -70,12 +70,29 @@ export const withCustomActions = withActions(() => ({
...
 				}
 
-				return { _response: { ok: res.ok } };
+				// On response call any externally defined handlers following Node's convention:
+				// 1. Either an error object or `null` as the first argument
+				// 2. An object containing `consent` and `payload` as the second
+				// Allows callbacks to decide how to handle a failure scenario
+
+				if (res.ok === false) {
+					throw new Error(res.statusText);
```
carlesandres, line 79:
```suggestion
					throw new Error(res.statusText || res.status);
```

PR #473, opened 2020-06-16
ACC-433 gift article form field label

thurstontye:
Fixes issue rasied in DAC Q2 2020 audit.
Gift article link input has no associated label or aria-label property.
For more detail see: https://financialtimes.atlassian.net/browse/ACC-433
Adds aria-label to input as this seems the most aproprate fix due to the context of it being in a 'gift article' modal box.
Should be treated as a patch release.

### DIFF
--- a/components/x-gift-article/src/Url.jsx
+++ b/components/x-gift-article/src/Url.jsx
@@ -22,7 +22,7 @@ export default ({ shareType, isGiftUrlCreated, url, urlType }) => {
 				className={ urlClassNames }
 				disabled={ shareType === ShareType.gift && !isGiftUrlCreated }
 				readOnly
-				aria-label="gift-article-link"
+				aria-label="Gift article shareable link"
 			/>
 		</span>
 	);
--- a/components/x-gift-article/src/Url.jsx
+++ b/components/x-gift-article/src/Url.jsx
@@ -12,14 +12,9 @@ const urlClassNames = [
 	styles['url-input']
 ].join(' ');
 
-const urlLabelClassNames = [
-	styles['hidden-label-text']
-].join('');
-
 export default ({ shareType, isGiftUrlCreated, url, urlType }) => {
 	return (
 		<span className={ urlWrapperClassNames }>
-			<label htmlFor={urlType} className={urlLabelClassNames}></label>
 			<input
 				type="text"
 				name={ urlType }
@@ -27,6 +22,7 @@ export default ({ shareType, isGiftUrlCreated, url, urlType }) => {
 				className={ urlClassNames }
 				disabled={ shareType === ShareType.gift && !isGiftUrlCreated }
 				readOnly
+				aria-label="gift-article-link"
 			/>
 		</span>
 	);
--- a/components/x-gift-article/src/Url.jsx
+++ b/components/x-gift-article/src/Url.jsx
@@ -12,9 +12,14 @@ const urlClassNames = [
 	styles['url-input']
 ].join(' ');
 
+const urlLabelClassNames = [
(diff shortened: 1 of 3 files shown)


components/x-gift-article/src/Url.jsx
```diff
@@ -22,6 +22,7 @@ export default ({ shareType, isGiftUrlCreated, url, urlType }) => {
 				className={ urlClassNames }
 				disabled={ shareType === ShareType.gift && !isGiftUrlCreated }
 				readOnly
+				aria-label="gift-article-link"
```
umbobabo, line 25: Why is this kebab case? Wouldn't be better something like...
```suggestion
				aria-label="Gift article shareable link"
```
?
Because that is what is contained in the input.

thurstontye: True, not sure what I was thinking!

nickcolley: Agree that we should try to give it a name that we would describe it in plain english.

For example: "Shareable article link"

nickcolley:
I do worry that this share interaction is just a bit too complicated to understand generally so perhaps best thing to do is to do the simplest change and then recommend reviewing the interaction itself later on.

umbobabo:
> I do worry that this share interaction is just a bit too complicated to understand generally so perhaps best thing to do is to do the simplest change and then recommend reviewing the interaction itself later on.

👍 Sounds like the best approach to me. Let's fix first the issue so to be sure that we are conforming with the DAC Audit and let's maybe push to iterate on that functionality.

nickcolley:
@umbobabo I'll keep a note of this then, thanks :)

components/x-gift-article/src/Url.jsx
thurstontye: Updated as per @umbobabo suggestion. Interestingly VO on mac can read 'gift-article-link' correctly

thurstontye:
> > I do worry that this share interaction is just a bit too complicated to understand generally so perhaps best thing to do is to do the simplest change and then recommend reviewing the interaction itself later on.
> 
> 👍 Sounds like the best approach to me. Let's fix first the issue so to be sure that we are conforming with the DAC Audit and let's maybe push to iterate on that functionality.

I agree. Perhaps I can raise as an issue to review gift article template @umbobabo here in x-dash or Accounts backlog?

components/x-gift-article/src/Url.jsx
andygout: Didn't occur to me either (whoops)!

Interesting to know that VoiceOver can handle kebab-cased text. I know that with uppercase (e.g. HELLO) it will read each letter individually (H-E-L-L-O).

umbobabo:
> > > I do worry that this share interaction is just a bit too complicated to understand generally so perhaps best thing to do is to do the simplest change and then recommend reviewing the interaction itself later on.
> > 
> > 
> > 👍 Sounds like the best approach to me. Let's fix first the issue so to be sure that we are conforming with the DAC Audit and let's maybe push to iterate on that functionality.
> 
> I agree. Perhaps I can raise as an issue to review gift article template @umbobabo here in x-dash or Accounts backlog?

Yes please go on. Add in the backlog so we can report this to the rest of the team.

umbobabo:
@nickcolley @thurstontye FYI I think, after a recent release of the App team about gift article tha you can see here...

image

there are already discussions going on to improve this functionality (see Aisliin reply on Simi release notes).
@rowanbeentje I wonder if your recent release is branching out from the usage of this component and if so if you are still using a x-dash component.

Issue #446, opened 2020-05-05
Live Blogs Post Component


ker-an:
Live blogs have a direct impact on engagement and drives search engine traffic which is why Content Innovation has been tasked with rebuilding (and improving) it. 

Currently, a reader experiences two different versions of a live blog when viewing it on the App or FT.com - with the experience on the App being worse. Content Innovation would like to create a live blogs post component using x-dash so we can provide a consistent reader experience across our platforms.

Here's a screenshot of a current live blogs post:

![image](https://user-images.githubusercontent.com/30316203/81046553-91d0f400-8eb0-11ea-950a-24add0829b2e.png)

apaleslimghost:
i think this is a good idea. i have a couple of questions:

1. i'm guessing it'll follow the FT.com design? using CSS modules in the component? (that's x-dash's convention, see [x-gift-article](https://github.com/Financial-Times/x-dash/blob/master/components/x-gift-article/src/GiftArticle.scss) for an example)
2. is the relative date required? x-teaser [outputs o-date markup](https://github.com/Financial-Times/x-dash/blob/master/components/x-teaser/src/RelativeTime.jsx) but expects the app consuming the component to initialise o-date, would that be okay?
3. same question as 2 but for the share buttons, is it okay for the consuming app to have to handle initialising o-share?
4. the live blogs controller in the mobile apps is... bad (it's okay i can say that i wrote it). using an x-dash component there might be tricky. do you also plan to build a live blogs post list x-dash component? there's nothing really that complex in x-dash yet but i think it would be a good fit; orchestrating the events & posts in the x-interaction model might be hard though, so that might be the time to investigate using Hooks (#248)

GlynnPhillips:
Hey @apaleslimghost here are my responses based on limited knowledge of x-dash.

1. Yes, taking a brief look at gift article I think this seems sensible.
2. Yes this should be fine
3. Yes this should be fine, but another alternative would be to have this component only interested in the "content" of a post and have the share component added by the consuming app. 
4. With a lot of live blogs code base in the mobile apps and next-article we plan to start from scratch unless it makes sense to try and save something. I think your suggestion is interesting and something we should certainly look into and chat with you about for any guidance.

Worth noting we don't have a "final" design yet but we have something we can use as a starting point but this could mean there is room for change in these answers. Eg. If the timestamp changes this is something we would have to take into account and maybe change our decisions

pixelandpage:
[...]

PR #474, opened 2020-06-17
x-live-blog-post component

tbergmen:
`x-live-blog-post` renders live blog posts with an updated design. 

Before:

Screenshot 2020-06-17 10 26 53

After:

Screenshot 2020-06-17 10 25 29

**Related issue:**
https://github.com/Financial-Times/x-dash/issues/446


tbergmen:
I have rebased this branch on `master`.

components/x-content-block/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "@financial-times/x-rollup": "file:../../packages/x-rollup",
+    "bower": "^1.8.8"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/Financial-Times/x-dash.git"
+  },
+  "homepage": "https://github.com/Financial-Times/x-dash/tree/master/components/x-content-block",
+  "engines": {
+    "node": ">= 6.0.0"
```
andygout, line 32: Using `>=` could result in using a non-LTS version of Node.

I would suggest using syntax such "node": "6.x" to lock it onto a specific LTS version.

Although v12 is the current LTS version - can we specify that or is there a specific reason for using v6?

I would suggest using this as an opportunity to bump the Node version of all the consuming apps to v12 rather than keeping this new component at a low enough level to be compatible with them.

components/x-content-block/readme.md
```diff
@@ -0,0 +1,46 @@
+# x-content-block
+
+This module has these features and scope.
```
andygout, line 3: There should be an opening description here of what the component is and what it does.

components/x-content-block/readme.md
```diff
@@ -0,0 +1,46 @@
...
+
+
+## Installation
+
+This module is compatible with Node 6+ and is distributed on npm.
```
andygout, line 8: Earlier Node version comment applies here too.

components/x-content-block/src/ContentBlock.scss
```diff
@@ -0,0 +1,70 @@
...
+.content-block__timestamp-container:after {
+  	content: '';
+  	display: block;
+  	width: oSpacingByName('s4');
+  	border-bottom: 4px solid oColorsMix(black, paper, 90);
+}
+
+.content-block__share-buttons {
+  	margin-top: oSpacingByName('s6');
+  	margin-bottom: oSpacingByName('s8')
```
andygout, line 51: Missing semi-colon. Surprised we do not have linting rule to pick up on this.

components/x-content-block/src/ContentBlock.scss
```diff
@@ -0,0 +1,70 @@
...
+	text-transform: uppercase;
+  	margin-top: oSpacingByName('s4');
+}
+
+.content-block__breaking-news:before {
+  	content: '';
+  	display: inline-block;
+  	width: 8px;
+  	height: 8px;
+	margin-right: 4px;
```
andygout, line 67: Should `oSpacingByName` be used here (as with the other classes in this file)?

components/x-content-block/src/Timestamp.jsx
```diff
@@ -0,0 +1,45 @@
+import { h } from '@financial-times/x-engine';
+import styles from './ContentBlock.scss';
+
+// this function extracts time in HH:mm format from a given date
```
andygout, line 4: Does `o-date` or any other existing package do this? Seems like it must be a common use case for elsewhere on FT.com.

components/x-content-block/package.json
tbergmen: I made it compatible with 10.x and above, because the FT app is currently pinned to 10.x.

components/x-content-block/readme.md
tbergmen: Added the description.

components/x-content-block/readme.md
tbergmen: Done

components/x-content-block/src/ContentBlock.scss
tbergmen: Fixed

components/x-content-block/src/ContentBlock.scss
tbergmen: Done

components/x-content-block/package.json
andygout: I think the `>=` might still allow usage of a non-LTS Node version?

Or maybe this is not an issue for components, but only the apps, i.e. does the component have to fall in line with the Node version used by the app consuming it? 🤔

components/x-content-block/src/Timestamp.jsx
tbergmen: Now using `o-date` for rendering the exact date.

tbergmen:
@andygout For accessibility, I added more descriptive aria labels to the sharing buttons. Apart from that, I couldn't see any issues. Do you have anything else in mind?

components/x-content-block/src/Timestamp.jsx
andygout: Can the code that calculates the time ll.5-23 not use one of the `o-date` methods listed under the [JavaScript](https://github.com/Financial-Times/o-date#javascript) header?

It seems like what you are trying to achieve must be a common use case and I would be surprised if it was not achievable with one of these methods.

andygout:
Re. accessibility, I think as long as you have tested it yourself with a screenreader and found that the component can be navigated in an understandable manner (maybe in the context of an app) then that should be okay.

components/x-content-block/src/Timestamp.jsx
tbergmen: I avoid using Origami Javascript in the x-dash component because of this comment here: https://github.com/Financial-Times/x-dash/issues/446#issuecomment-623949600

andygout: Ah, interesting.

apaleslimghost:
i see this has been tested in `next-article`, has it been tested in `ft-app` yet? that's a requirement for release

tbergmen:
@andygout  @apaleslimghost 

I tested the component in `ft-app` as well. The only issue was that the share buttons wouldn't work in the app. Therefore I created a PR that adds the option to hide share buttons. - https://github.com/Financial-Times/x-dash/pull/494

The rest of the features work well in the app.

andygout:
So https://github.com/Financial-Times/x-dash/pull/494 and this PR will be included in the same release?

tbergmen:
> So #494 and this PR will be included in the same release?

@andygout Yes, I'm planning to merge #494 into this branch before merging this to master. Do you think that would work?

andygout:
Might be worth doing one final check on both FT.com and the app now that https://github.com/Financial-Times/x-dash/pull/494 has been merged into this branch to ensure that its functionality works as expected for both.

tbergmen:
@andygout I tested the component including the latest changes on `next-article` and on `ft-app`. It works well in both platforms. 

Do you think we can go ahead with the release today?

PR #481, opened 2020-06-18
ACC-434 Added aria to copy button

AniaMakes:
This is a minor touch up to improve accessibility and make it clear what the user is copying.

### DIFF
--- a/components/x-gift-article/src/Buttons.jsx
+++ b/components/x-gift-article/src/Buttons.jsx
@@ -44,7 +44,7 @@ export default ({
 						className={ ButtonWithGapClassNames }
 						type="button"
 						onClick={ shareType === ShareType.gift ? actions.copyGiftUrl : actions.copyNonGiftUrl }
-						aria-label="Copy the link to your clipboard to share this article"
+						aria-label="Copy the gift article link to your clipboard"
 						>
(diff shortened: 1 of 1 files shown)


components/x-gift-article/src/Buttons.jsx
```diff
@@ -44,6 +44,7 @@ export default ({
 						className={ ButtonWithGapClassNames }
 						type="button"
 						onClick={ shareType === ShareType.gift ? actions.copyGiftUrl : actions.copyNonGiftUrl }
+						aria-label="Copy the link to your clipboard to share this article"
```
andygout, line 47:
```suggestion
						aria-label="Copy the gift article link to your clipboard"
```

The current wording suggests that the action of pressing the button will result in the article being shared, so this suggestion is to try and make the wording clear that all it does is put the link on the clipboard.

AniaMakes: I copy pasted what the accessibility team suggested, but your comment makes more sense!

andygout:
Please be sure to follow the [Release Guidelines](https://github.com/Financial-Times/x-dash/blob/master/docs/components/release-guidelines.md).

PR #482, opened 2020-06-18
ACC-432 radio buttons group reference

thurstontye:
Groups radio button inputs as a multiple form input for Assistive technology(AT).. 

Addresses the DAC issue raised in the Q2 audit see: https://financialtimes.atlassian.net/browse/ACC-432

DAC recommendation was to group in fieldset and ledgend but the grouping has been applied using role on the containing element and aria-labelledby properties for reasons described in o-forms accessibility docs: 
https://github.com/Financial-Times/o-forms/blob/master/ACCESSIBILITY.md#multiple-inputs

### DIFF
--- a/components/x-gift-article/src/GiftArticle.scss
+++ b/components/x-gift-article/src/GiftArticle.scss
@@ -1,4 +1,5 @@
 @import 'src/lib/variables';
+@import 'o-normalise/main';
 
 $o-buttons-is-silent: true;
 @import 'o-buttons/main';
@@ -35,6 +36,22 @@ $o-typography-is-silent: true;
 
 	.radio-button-section {
 		margin-bottom: 12px;
+	}
+
+	.share-option-title {
+		@include oNormaliseVisuallyHidden();
+	}
+
+	.share-option-title {
+		@include oNormaliseVisuallyHidden();
+	}
+
+	.share-option-title {
+		@include oNormaliseVisuallyHidden();
+	}
+
+	.share-option-title {
+		@include oNormaliseVisuallyHidden();
 	}
 }
 
--- a/components/x-gift-article/src/RadioButtonsSection.jsx
+++ b/components/x-gift-article/src/RadioButtonsSection.jsx
@@ -11,7 +11,8 @@ const radioSectionClassNames = [
 ].join(' ');
 
 export default ({ shareType, showGiftUrlSection, showNonGiftUrlSection }) => (
-	<div className={ radioSectionClassNames }>
+	<div className={ radioSectionClassNames } role="group" aria-labelledby="article-share-options">
+		<span className={ styles["share-option-title"] } id="article-share-options">Article share options</span>
 		<label htmlFor="giftLink">
 			<input
 				type="radio"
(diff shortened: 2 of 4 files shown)


components/x-gift-article/src/GiftArticle.scss
```diff
@@ -35,6 +36,13 @@ $o-typography-is-silent: true;
 
 	.radio-button-section {
 		margin-bottom: 12px;
+
+
+	}
```
andygout, line 41: Whitespace?

components/x-gift-article/src/GiftArticle.scss
```diff
@@ -35,6 +36,13 @@ $o-typography-is-silent: true;
...
+
+	.share-option-title {
```
andygout, line 43: The class that gets produced from `styles["share-option-title"]` in the snapshot test is `GiftArticle_share-option-title__1CFAK`.

I am not entirely sure about what is the logic behind `styles["share-option-title"]` but the results indicate that this `.share-option-title` class would have no elements to which to apply styling.

I have probably missed something - could you fill me in pls?

components/x-gift-article/src/GiftArticle.scss
```diff
@@ -35,6 +36,13 @@ $o-typography-is-silent: true;
...
+		color: blue;
```
andygout, line 44: Is there a specific blue from the `o-colors` palette that can be used?

thurstontye: @andygout was using that to debug my css should have removed before commit. Lesson in not using hacky methods to debug things 😬

components/x-gift-article/src/GiftArticle.scss
thurstontye: `.share-option-title` becomes `GiftArticle_share-option-title__1CFAK` in the distributed code.

My understanding of the logic is it's rollup css-js module bundling 'magic' that transpiles the class names from those in the .scss to [Component_name]_your-class_[some-random-suffix] done with the config for: https://github.com/egoist/rollup-plugin-postcss

andygout: Thanks for the explanation.

I would also have accepted: "Magic ✨" 😆

components/x-gift-article/src/RadioButtonsSection.jsx
```diff
@@ -11,7 +11,8 @@ const radioSectionClassNames = [
 ].join(' ');
 
 export default ({ shareType, showGiftUrlSection, showNonGiftUrlSection }) => (
-	<div className={ radioSectionClassNames }>
+	<div className={ radioSectionClassNames } role="group" aria-labelledby="article-share-options">
```
umbobabo, line 14: I wonder if `role="group"` here is really necessary considering that is already on the parent. [See here](https://github.com/Financial-Times/x-dash/pull/482/files#diff-c19ca0f6bf98fd083196ea1f6a85f037R514). I don' remeber if need to be the immediate parent or not, maybe worth checking @thurstontye ?

thurstontye: Will do but my understanding is they are different roles and so need to be declared separately.

umbobabo:
I'm happy with the answer to my question. For the rest it probably worth rebase and squash before merge.

components/x-gift-article/src/GiftArticle.scss
thurstontye: done

PR #485, opened 2020-06-19
Rename x-content-block to x-live-blog-post

tbergmen:
Renames `x-content-block` to `x-live-blog-post`.

For `x-content-block` component, please see this PR: https://github.com/Financial-Times/x-dash/pull/474


components/x-live-blog-post/package.json
```diff
@@ -1,11 +1,11 @@
 {
-  "name": "@financial-times/x-content-block",
+  "name": "@financial-times/x-live-blog-post",
   "version": "0.0.0",
   "description": "",
-  "main": "dist/ContentBlock.cjs.js",
-  "module": "dist/ContentBlock.esm.js",
-  "browser": "dist/ContentBlock.es5.js",
-  "style": "dist/ContentBlock.css",
+  "main": "dist/LiveBlogPpost.cjs.js",
```
ker-an, line 5: Small typo - there is an extra `p` here that needs to be removed

components/x-live-blog-post/readme.md
```diff
@@ -1,14 +1,14 @@
-# x-content-block
+# x-live-blog-post
 
-This module displays a block of content with an id, title, body and timestamp. Main use case is the new version of live blogs, although we don't want to limit its usage to it.  
+This module displays a live blog post with an id, title, body and timestamp.
```
ker-an, line 3: lol thanks to Simon we now know time stamp is two words 😄 so just changing this to `time stamp` from `timestamp`

components/x-live-blog-post/package.json
tbergmen: Thanks Keran 👍 

Should be fixed now.

components/x-live-blog-post/readme.md
andygout, line 3: Pedantically, this sounds like the post displays the ID, which it doesn't. Although it does display the writer's name and social media links, which might be worth mentioning.

This is a side note and unrelated to the purpose of this PR, but I am raising it in case you want to do a small follow-up PR.

tbergmen: Will do that. Thank you!

PR #493, opened 2020-06-22
Added troubleshooting Chrome error to installation instructions

AniaMakes:
I doubt I'm the only one encountering this issue, and that way what helped me is there already...

### DIFF
--- a/docs/get-started/installation.md
+++ b/docs/get-started/installation.md
@@ -41,3 +41,10 @@ To aid the development of interactive components with Storybook it is recommende
     make build
     npm run start-storybook
     ```
+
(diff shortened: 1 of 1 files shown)


apaleslimghost:
this isn't an x-dash specific problem, it can happen with any FT.com app, so i think maybe it should be in the [wiki](https://github.com/financial-times/next/wiki) instead?

AniaMakes:
Mostly, I ran into this problem in this specific case, and even if the details go into the wiki (any suggestions *where* to put it are appreciated), I think it would be beneficial to link it from here. I have read some of the wiki when I joined but I don't recall much of it and it wouldn't necessarily be the first place I'd look for troubleshooting (because it's so large).

PR #494, opened 2020-06-22
x-live-blog-post option for hiding share buttons

tbergmen:
FT app is using native sharing, so we want to hide the share buttons when using this component in the app.

### DIFF
--- a/components/x-live-blog-post/src/LiveBlogPost.jsx
+++ b/components/x-live-blog-post/src/LiveBlogPost.jsx
@@ -11,7 +11,7 @@ const LiveBlogPost = (props) => {
 		publishedTimestamp,
 		isBreakingNews,
 		articleUrl,
-		showShareButtons,
+		showShareButtons = false,
 	} = props;
 
 	return (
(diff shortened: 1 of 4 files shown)


components/x-live-blog-post/src/LiveBlogPost.jsx
```diff
@@ -10,7 +10,8 @@ const LiveBlogPost = (props) => {
 		content,
 		publishedTimestamp,
 		isBreakingNews,
-		articleUrl
+		articleUrl,
+		showShareButtons = true,
```
GlynnPhillips, line 14: I think we should switch this to be defaulted to false. I think this because having it as true relies on the consuming app (next-article or ft-app) to initialize o-share which for me makes more sense that both parts are done together. eg. `showShareButtons=true` and initialize o-share. Rather than needing to know to set `showShareButtons` as false because you are not using o-share.

I also think this might make a more obvious development flow that leads you to the documentation. 

eg. "why aren't the share buttons working?" - Look at docs - "Oh I need to set `showShareButtons` to true and initialize o-share"

tbergmen: @GlynnPhillips this is now set to false by default. 

Also added this property to the README file.

andygout: > @GlynnPhillips this is now set to false by default.
It now has no default value. Might be better to explicitly state `false` as the default value rather than rely on the `null` or `undefined` value which I guess is currently the case?

PR #499, opened 2020-06-25
Version 1.2.1

oliverturner:


oliverturner:
Thanks for putting that list together: only reason I didn’t was because I felt it duplicated what was in the referenced PR.

I’ll add and resubmit 👍

PR #500, opened 2020-06-29
X privacy manager v1.2.2

oliverturner:
Updates `x-privacy-manager`
  - Adds more Spoor tracking
  - Adds `loginPrompt` prop to allow consuming apps to provide customisable messaging to logged-out users
  - Allows stories to opt out of having HTML input escaped

The biggest change is the addition of a set of helper functions extracted from `next-control-centre`. Consolidating them under this component avoids consuming apps having to replicate the same functionality every time, and appropriately colocates highly coupled code, ensuring that they can be  tested together.

PR #502, opened 2020-07-02
Add an optional cookieDomain to PrivacyManager

vbachev:
Updates `x-privacy-manager`

Adds an optional `cookieDomain` prop that allows consuming apps to override the domain of the consent cookie being set with the successful response from Consent API. When provided, the `cookieDomain` prop is attached to the payload being sent over to Consent Proxy and overrides the default ".ft.com" here https://github.com/Financial-Times/next-consent-proxy/blob/master/src/controllers/set-consent-record-cookie.ts#L58

Background:
This component is used on specialist titles (e.g. https://cookies.ftadviser.com/preferences/privacy-ccpa?referrer=www.ftadviser.com&cookieDomain=.ftadviser.com&legislation=ccpa ) where the consent cookie from Consent API needs to be set to the appropriate domain for the title. 
[This PR in `next-control-centre`](https://github.com/Financial-Times/next-control-centre/pull/189) enables the `&cookieDomain=.ftadviser.com` URL parameter from the example above to be passed as prop to `<PrivacyManager />`. The changes in this current PR are the logical next step that allows for the consent cookie to correctly get set on any specialist title.

### DIFF
--- a/components/x-privacy-manager/readme.md
+++ b/components/x-privacy-manager/readme.md
@@ -26,16 +26,17 @@ The [`x-engine`][engine] module is used to inject your chosen runtime into the c
 Feature                     | Type       | Notes
 ----------------------------|------------|-----------------------------------------------
 `consentSource`             | string     | Name of the consuming app to be included in requests to Consent Proxy (e.g. "next-control-centre")
-`consentProxyEndpoints`    | object     | Dictionary containing already-formed Consent Proxy Endpoints to use (including userId). It must include, at least, `consentProxyEndpoints.createOrUpdateRecord`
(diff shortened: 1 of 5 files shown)


vbachev:
Thanks! Was just about to push that change 👍  Is this the place to bump the component version in its `package.json`? Or does that happen in a follow-up PR?

carlesandres:
AFAIK, you don't set the version in code for `x-dash` components. It gets assigned upon releasing.

vbachev:
Pushed some spoor tracking improvements that @oliverturner suggested I copy from https://github.com/Financial-Times/x-dash/pull/500/commits/daaf6bf464b9ce16b540271b1dfe1ba854eced2a

PR #504, opened 2020-07-03
Version 1.3.0

vbachev:
Includes only changes to `x-privacy-manager`

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,12 @@
 
 ## v1
 
+### v1.3.0
+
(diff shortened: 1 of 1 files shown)


apaleslimghost:
thanks for the PR @vbachev! since you're outside of Customer Products, by our [merging guidelines](https://github.com/Financial-Times/next/wiki/Who-can-push-to-customer-products-repositories%3F) we should merge this ourselves. but since this is a documentation change only, i'll do that as soon as the build passes 👍

PR #505, opened 2020-07-10
Update unit tests of x-privacy-manager

vbachev:
This PR fills some gaps in unit-test coverage of `x-privacy-manager`:
- testing whether the optional `cookieDomain` prop is attached to fetch request and callback payloads
- testing that callbacks are called with an error argument in case of fetch request failure

The unit test layout is changed so we can extract the `consent choice handling` logic from the `initial state` cases.

### DIFF
--- a/components/x-privacy-manager/src/__tests__/privacy-manager.test.jsx
+++ b/components/x-privacy-manager/src/__tests__/privacy-manager.test.jsx
@@ -34,18 +34,11 @@ const buildPayload = (consent) => ({
 			},
 		},
 	},
-	cookieDomain: '.ft.com',
 	formOfWordsId: 'privacyCCPA',
 });
 
-function checkPayload(opts, expected) {
-	const consents = JSON.parse(String(opts.body)).data;
-
-	let worked = true;
-	for (const category in consents) {
-		worked = worked && consents[category].onsite.status === expected;
-	}
-	return worked;
+function getLastFetchPayload() {
+	return JSON.parse(fetchMock.lastOptions().body);
 }
 
 const defaultProps = {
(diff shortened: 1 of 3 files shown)


components/x-privacy-manager/src/__tests__/privacy-manager.test.jsx
```diff
@@ -34,18 +34,11 @@ const buildPayload = (consent) => ({
...
-function checkPayload(opts, expected) {
-	const consents = JSON.parse(String(opts.body)).data;
-
-	let worked = true;
-	for (const category in consents) {
-		worked = worked && consents[category].onsite.status === expected;
-	}
-	return worked;
+function getLastFetchPayload() {
+	return JSON.parse(fetchMock.lastOptions().body);
```
vbachev, line 41: With this change the fetch payload is now check for deep-equality instead of just checking the category status flags. In my opinion it's more straight-forward and relies on less code

components/x-privacy-manager/src/__tests__/privacy-manager.test.jsx
```diff
@@ -56,20 +49,10 @@ const defaultProps = {
...
 describe('x-privacy-manager', () => {
 	describe('initial state', () => {
-		beforeEach(() => {
-			fetchMock.reset();
-			const okResponse = {
-				body: { a: 'b' },
-				status: 200,
-			};
-			fetchMock.mock(TEST_CONSENT_URL, okResponse, { delay: 500 });
-		});
```
vbachev, line 71: This is moved below within `describe('handling consent choices', () => {`

oliverturner:
Closing this as the tests have been integrated into https://github.com/Financial-Times/x-dash/pull/565

PR #507, opened 2020-07-16
CI-261 Remove `&amp;` from share URLs

ker-an:
The `x-live-blog-post` share URLs are not working in `next-article` because at some point they are being double encoded. As a result, users who click on the LinkedIn share link on a live blog post are being sent to an error page:

image

[Content Innovation Ticket](https://financialtimes.atlassian.net/browse/CI-261)


ker-an:
Thanks for the 👍 - I forgot to update the tests (broken build) so will do that now!

PR #508, opened 2020-07-23
Changed message in x-privacy-manager component

carlesandres:
A simple message change as requested in [this ticket](https://financialtimes.atlassian.net/browse/ADSDEV-566). 

It could make sense to release as a minor update. Thoughts?

### DIFF
--- a/components/x-privacy-manager/src/messages.jsx
+++ b/components/x-privacy-manager/src/messages.jsx
@@ -49,11 +49,11 @@ export function ResponseMessage({ success, referrer }) {
 	const statusDict = {
 		true: {
 			cls: 'o-message--success',
-			msg: 'Your setting has been saved.',
+			msg: 'Your settings have been saved on this device. Please apply your preferences to all of the devices you use to access our Sites.',
 		},
 		false: {
 			cls: 'o-message--error',
-			msg: 'Your setting could not be saved. Please try again later.',
+			msg: 'Your settings could not be saved. Please try again later.',
 		},
 	};
 


oliverturner:
Even a patch release would be fine

components/x-privacy-manager/src/messages.jsx
```diff
@@ -49,11 +49,11 @@ export function ResponseMessage({ success, referrer }) {
 	const statusDict = {
 		true: {
 			cls: 'o-message--success',
-			msg: 'Your setting has been saved.',
+			msg: 'Your settings have been saved on this device. Please apply your preferences to all of the devices you use to access our Sites.',
```
andygout, line 52:
```suggestion
			msg: 'Your settings have been saved on this device. Please apply your preferences to all of the devices you use to access our sites.',
```
Sites is not a proper noun in this context.

carlesandres: There is a precedent for using "Sites" somewhere else.

image

It was chosen for consistency by the Privacy Product Owner (Dan Fish) but if you think it's worth debating, I can go back to him @andygout.

andygout: @carlesandres I think it is worth asking.

This reads like 'Sites' is a named thing, e.g. a named product, but being preceded with 'our' indicates that this is likely not the intention.

dmfish1131: Hi Andy,

Sites (with a capital S) is a term used by Legal and Compliance as the manage cookies/privacy policies/CCPA pages is also used by Specialist titles. It's a way for them to show that this does not pertain only to FT.com. 

I hope that gives a bit more clarity. 

Dan

andygout: A-ha - so I guess the asterisk expands on that a bit more and that Sites is deemed a proper noun in this context because it's a definition in some legal agreement. Makes sense - thanks for the clarification.

carlesandres: Thank you @andygout for being so attentive and @dmfish1131 for the clarification.

PR #509, opened 2020-07-29
x-live-blog-wrapper

tbergmen:
`x-live-blog-wrapper` is designed to handle initial rendering of live blog posts at the server side and handle live updates at the client side.

**Current features:**
* Render initial posts
* Render hydration data at server side rendering
* Live updates

**Developer notes:**
In next-article, there is a hacky branch that renders this component with data from Wordpress live blogs. This branch is local testing purposes only and is not meant to be merged or used in production.
https://github.com/Financial-Times/next-article/tree/x-live-blog-wrapper-test
https://github.com/Financial-Times/next-article/commit/1ea0698a7b838c1180a8f49c3b901021caaff6c1


components/x-live-blog-wrapper/src/HydratedLiveBlogWrapper.jsx
```diff
@@ -0,0 +1,17 @@
...
+import { listenToLiveBlogEvents } from './LiveEventListener';
+
+const HydratedLiveBlogWrapper = (props) => {
+	const serialiser = new Serialiser();
+
+	return (
+		<Fragment>
+			<LiveBlogWrapper {...props} serialiser={serialiser} />
+			<HydrationData serialiser={serialiser} />
+		</Fragment>
```
apaleslimghost, line 13: i'd originally intended that using the serialiser would be left for the consuming app, since you might want to reuse the same serialiser for multiple components to reduce the output, and you might have particular requirements for where the `HydrationData` gets output

also i don't know if we need to think coordinating this with the various Page Kit components that output server data; `x-interaction` predates Page Kit and the hydration stuff hasn't been used in production yet

happy to chat further about this! 🙂

components/x-live-blog-wrapper/src/LiveEventListener.js
```diff
@@ -0,0 +1,98 @@
...
+		isKeyEvent: Boolean(post.keytext),
+		title,
+		content
+	};
+};
+
+const dispatchLiveUpdateEvent = (eventType, data) => {
+	// consumer app will need to consume this event after the component is rendered. therefore,
+	// we defer dispatching of this event.
+	window.setTimeout(() => document.dispatchEvent(new CustomEvent(eventType, { detail: data })), 0);
```
apaleslimghost, line 40: please don't dispatch global events from x-dash components. if this component is used in `ft-app`, they way it renders article pages means up to 3 instances of this component could be on the page at once, because of preloaded swipe panes. it looks like you should be able to use `x-interaction` for this use case though?

components/x-live-blog-wrapper/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "@financial-times/x-rollup": "file:../../packages/x-rollup",
+    "@financial-times/x-test-utils": "file:../../packages/x-test-utils"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/Financial-Times/x-dash.git"
+  },
+  "homepage": "https://github.com/Financial-Times/x-dash/tree/master/components/x-live-blog-wrapper",
+  "engines": {
+    "node": ">= 6.0.0"
```
ker-an, line 32: We prob wanna update the node version here or we can remove the field altogether

components/x-live-blog-wrapper/src/LiveBlogWrapper.jsx
```diff
@@ -0,0 +1,41 @@
...
+		};
+	},
+
+	deletePost (postId) {
+		return ({ posts }) => {
+			const updatedPosts = posts.filter((post) => post.postId !== postId);
+			return { posts: updatedPosts };
+		};
+	}
+});
```
simonplend, line 30: I'm wondering if there are pros or cons to creating new arrays when handling each event (as is happening here) vs modifying the existing posts array directly e.g.

- `insertPost` - Add `post` to the beginning of the `posts` array with [`Array.prototype.unshift()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/unshift)
- `updatePost` - Find post with [`Array.prototype.findIndex()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) and then do `posts[index] = updated`
- `deletePost` - Find post with [`Array.prototype.findIndex()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) and then use [`Array.prototype.splice()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice) to remove it

The `posts` array could get quite large, so the main benefit I can think of when modifying the array in place is that you're not using more memory by creating copies of the posts array every time we receive a new event. I'm not sure how much this matters though :thinking: Interested to hear your thoughts on this!

components/x-live-blog-wrapper/src/LiveBlogWrapper.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+const LiveBlogWrapper = liveBlogWrapperActions(({ posts = [], articleUrl, showShareButtons }) => {
+	const postElements = posts.map((post) => <LiveBlogPost key={`live-blog-post-${post.postId}`} {...post} articleUrl={articleUrl} showShareButtons={showShareButtons}/>);
```
simonplend, line 33: I think it would improve the readability of this statement if it was formatted across multiple lines.

components/x-live-blog-wrapper/src/LiveEventListener.js
```diff
@@ -0,0 +1,98 @@
+// TODO: parsing posts will not be needed when we use v2 route on next-live-events-api
```
simonplend, line 1: :tada:

components/x-live-blog-wrapper/src/HydratedLiveBlogWrapper.jsx
tbergmen: @apaleslimghost that makes sense. Currently, one of the consumer apps (`next-article`) is rendering this component using `x-handlebars`. I couldn't find a way to use serialiser with it. Do you think it is possible?

Would it be acceptable to offer a plain version of this component that doesn't serialise the hydration data as well as having this one as an option?

components/x-live-blog-wrapper/src/LiveEventListener.js
tbergmen: The purpose of this event is to notify the consumer app when the state of this component is updated. Would that be possible with `x-interaction`?

components/x-live-blog-wrapper/src/LiveBlogWrapper.jsx
ker-an, line 33: So this `key` here is used by `next-article` to initialise the Origami components?

components/x-live-blog-wrapper/src/LiveEventListener.js
```diff
@@ -0,0 +1,98 @@
...
+			return;
+		}
+
+		invokeAction('deletePost', [ post.postId ])
+		dispatchLiveUpdateEvent('LiveBlogWrapper.DELETE_POST', { postId: post.postId });
+	});
+
+	// TODO: do we handle live blog status updates in this component?
+	eventSource.addEventListener('close', (event) => {
+		// TODO
```
ker-an, line 93: Good question, I'm not sure. Is the alternative that the topper deals with this?

components/x-live-blog-wrapper/src/LiveEventListener.js
tbergmen: To be more specific;

We are listening to server sent events in this component and add new posts (by adding `x-live-blog-post` components to the markup) based on those events. 

`x-live-blog-post` component has Origami components that need to be initialised after rendering the markup. Because we want the consumer app to initialise Origami components, we notify the app using that event when the markup is changed.

components/x-live-blog-wrapper/src/LiveBlogWrapper.jsx
tbergmen: @simonplend Thanks for the suggestion. I have created PR for this: https://github.com/Financial-Times/x-dash/pull/517

components/x-live-blog-wrapper/src/LiveBlogWrapper.jsx
tbergmen: @ker-an key is a required property by React for items in a list. In this case we are creating a list of LiveBlogPost items and we assign a unique key to each post so that React can identify which posts are changed.

There is more info about this on React documentation: https://reactjs.org/docs/lists-and-keys.html#keys

components/x-live-blog-wrapper/src/HydratedLiveBlogWrapper.jsx
apaleslimghost: sorry, i missed these comments. i think setting `res.locals.serialiser` will work and get passed through to the component. you might need to do `serialiser=@root.serialiser` when rendering it

components/x-live-blog-wrapper/src/LiveEventListener.js
apaleslimghost: ah, okay. you could dispatch the events from the wrapper element and let them bubble instead?

apaleslimghost (commented):
overall this looks good, i think if you address my comment about event dispatching it'll be good to go!

components/x-live-blog-wrapper/src/LiveEventListener.js
tbergmen: Thanks @apaleslimghost 

Sorry I forgot to update this thread; we have already updated it to raise the event from the wrapper element: https://github.com/Financial-Times/x-dash/blob/313afe7a016b2911dfabafd6bac251d104832296/components/x-live-blog-wrapper/src/LiveEventListener.js#L52

Do you think it looks ok now?

apaleslimghost: ah! sorry, i didn't check the current state of this file. yes, that looks good!

apaleslimghost (approved):
okay, this looks ready! per the [release guidelines](https://github.com/Financial-Times/x-dash/blob/master/docs/components/release-guidelines.md) this should be tested in `ft-app` before merging. this could be as simple as installing it locally and writing a quick spike implementation, and checking that with the Apps team (you might already have done this, idk!)

components/x-live-blog-wrapper/src/HydratedLiveBlogWrapper.jsx
tbergmen: That's good to know. Thanks!

components/x-live-blog-wrapper/src/LiveBlogWrapper.jsx
```diff
@@ -0,0 +1,55 @@
+import { h } from '@financial-times/x-engine';
+import { LiveBlogPost } from '@financial-times/x-live-blog-post';
+import { withActions } from '@financial-times/x-interaction';
+import { listenToLiveBlogEvents } from './LiveEventListener';
+
+const withLiveBlogWrapperActions = withActions({
+	insertPost (post) {
```
nickcolley, line 7: When posts are updated and inserted, does the application take charge of informing the user about it?

GlynnPhillips: @nickcolley there is currently a notification that appears and takes you to the new post should you want to. This component is new so we are making some of these architectural decisions along the way and one that we have defined yet is how much of that should live in here and how much should live in the application.

The only version of these pages wasn't amazing for accessibility and I don't think it told you about updates or deletes and only mentioned creates. So we must improve this and I think it will be the main factor in "whose" responsibility that is, the app or the component. At this stage I am leaning towards component.

GlynnPhillips: Sorry that was a bit of a ramble to acknowledge that yes we need to make sure the user experience is accessible

apaleslimghost (dismissed):
smashing that approve button again since the recent commits dismissed my previous review

nickcolley:
Heya @tbergmen,

Did you get a chance to read this conversation that I had with Glynn?

https://github.com/Financial-Times/x-dash/pull/509#discussion_r478512642

ker-an (approved):
This is really great work and the comments/documentation really helped me.

tbergmen:
> Heya @tbergmen,
> 
> Did you get a chance to read this conversation that I had with Glynn?
> 
> [#509 (comment)](https://github.com/Financial-Times/x-dash/pull/509#discussion_r478512642)

Hello @nickcolley, yes, accessibility of live updates is in our to-do list. In the first release of live blogs, we won't have live updates. We will revisit this component when we are building that feature.

nickcolley:
@tbergmen thanks for the update, sounds great nice work!

PR #511, opened 2020-08-05
CI-288 Update timestamp to use latest version of `o-date`

ker-an:
The latest version allows us to render a date multiple times within the
same `o-date` component by including multiple `data-o-date-printer`
elements.

FYI, there is no visual change to the timestamp.

See https://github.com/Financial-Times/o-date/releases/tag/v4.1.0.

### DIFF
--- a/components/x-live-blog-post/src/Timestamp.jsx
+++ b/components/x-live-blog-post/src/Timestamp.jsx
@@ -27,19 +27,19 @@ export default ({ publishedTimestamp }) => {
 			<time
 				data-o-component="o-date"
 				className={`o-date ${styles['live-blog-post__timestamp']}`}
-				itemProp="datePublished"
+				dateTime={publishedTimestamp}
 				data-o-date-format={format}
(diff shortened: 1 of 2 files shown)


components/x-live-blog-post/src/Timestamp.jsx
```diff
@@ -27,19 +27,19 @@ export default ({ publishedTimestamp }) => {
 			<time
 				data-o-component="o-date"
 				className={`o-date ${styles['live-blog-post__timestamp']}`}
-				itemProp="datePublished"
+				dateTime={publishedTimestamp}
 				data-o-date-format={format}
-				dateTime={publishedTimestamp}>{formatted}
+				itemProp="datePublished">
+					<span data-o-date-printer>{formatted}</span>
```
ker-an, line 33: FYI, this `{formatted}` date is overridden by `o-date` and only appears if JS is disabled:

image

GlynnPhillips: Yep, I think this looks good to me

PR #512, opened 2020-08-06
Use next-live-event-api v2 endpoint to listen to live updates

tbergmen:
Connects to `next-live-events-api` v2 endpoint for live updates.

### DIFF
--- a/components/x-live-blog-wrapper/src/LiveEventListener.js
+++ b/components/x-live-blog-wrapper/src/LiveEventListener.js
@@ -1,37 +1,11 @@
-// TODO: parsing posts will not be needed when we use v2 route on next-live-events-api
 const parsePost = (event) => {
 	const post = JSON.parse(event.data);
 
-	if (!post || !post.mid || !post.textrendered) {
+	if (!post || !post.postId) {
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


components/x-live-blog-wrapper/src/LiveEventListener.js
```diff
@@ -51,11 +25,9 @@ const listenToLiveBlogEvents = ({ liveBlogWrapperId, blogPath }) => {
 		);
 	};
 
-	// TODO: use next-live-events-api v2 routes
-	// const eventSource = new EventSource(`https://next-live-event.ft.com?eventid=${blogPath}&formatted=false`, { withCredentials: true });
-	const eventSource = new EventSource(`http://localhost:3000/events`, { withCredentials: false });
+	const eventSource = new EventSource(`https://next-live-event.ft.com/v2/liveblog/${liveBlogArticleUuid}`, { withCredentials: true });
```
simonplend, line 28: Now that this is integrating with our new architecture, I think that `liveBlogArticleUuid` should be `liveBlogPackageUuid`.

tbergmen: Good catch. This is not updated.

PR #513, opened 2020-08-07
Allow users to specify the data-trackable attribute for the teaser

bjfletcher:
This introduces another attribute to the teaser: data-trackable

There are data-trackable for various bits within the teaser (e.g. link type) however we need a data-trackable for the teaser as a whole. One example is to add details where the teaser is placed on the frontpage:

- `data-trackable="slice:hero-slice-1;slicePos:1;layout:standaloneimage;listPos:1"` (first top story)

- `data-trackable="slice:freeform-twocolumns;slicePos:5;layout:standaloneimage;listPos:14"` (last top story)

See https://github.com/Financial-Times/next-front-page/pull/2392 for the code there to add the above tracking. That PR would follow this PR (& release).

So, in short, this PR simply allows users of `x-teaser` to specify, optionally, a tracking value for the teaser.

### DIFF
--- a/components/x-teaser/storybook/top-story.js
+++ b/components/x-teaser/storybook/top-story.js
@@ -17,6 +17,7 @@ exports.knobs = [
 	'metaPrefixText',
 	'metaSuffixText',
 	'metaLink',
+	'dataTrackable',
 	// Title
 	'showTitle',
 	'title',
--- a/components/x-teaser/__fixtures__/top-story.json
+++ b/components/x-teaser/__fixtures__/top-story.json
@@ -8,6 +8,7 @@
   "altStandfirst": "Groping and sexual harassment at black-tie dinner charity event",
   "publishedDate": "2018-01-23T15:07:00.000Z",
   "firstPublishedDate": "2018-01-23T13:53:00.000Z",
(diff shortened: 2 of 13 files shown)


components/x-teaser/storybook/top-story.js
```diff
@@ -17,6 +17,7 @@ exports.knobs = [
 	'metaPrefixText',
 	'metaSuffixText',
 	'metaLink',
+	'dataTrackable',
```
taraojo, line 20: This only shows for the top story teaser in Storybook and not the other teasers, I'm not sure how important it is for it to show up for all of them. Was there a specific reason you only added to `top-story` @bjfletcher?

bjfletcher: No good reason (we're currently analysing only top stories) and I think that's a good idea to add to all the other types for the storybook. So I've just pushed a commit to add the knob to all the other teaser types.

taraojo (approved):
Looks good to me, thanks Ben

taraojo:
I'll look at merging and releasing this tomorrow morning @bjfletcher as I'm a lot less familiar with that process

PR #514, opened 2020-08-20
Remove HydratedLiveBlogWrapper

tbergmen:
Because we will soon be using JSX instead of handlebars in next-article for rendering this component, we don't need to serialise the hydration data in the component anymore. This will need to be done in the consumer app from now on.


components/x-live-blog-wrapper/src/LiveBlogWrapper.jsx
```diff
@@ -29,13 +30,21 @@ const liveBlogWrapperActions = withActions({
...
-	);
-});
 
-export { LiveBlogWrapper };
+const LiveBlogWrapper = liveBlogWrapperActions(function LiveBlogWrapper ({ posts = [], articleUrl, showShareButtons }) {
+		const postElements = posts.map(post =>
+			<LiveBlogPost key={`live-blog-post-${post.postId}`}
+										{...post}
+										articleUrl={articleUrl}
+										showShareButtons={showShareButtons}/>
```
simonplend, line 39: Indentation looks a little odd here.

components/x-live-blog-wrapper/src/LiveBlogWrapper.jsx
```diff
@@ -29,13 +30,21 @@ const liveBlogWrapperActions = withActions({
...
-	const postElements = posts.map((post) => <LiveBlogPost key={`live-blog-post-${post.postId}`} {...post} articleUrl={articleUrl} showShareButtons={showShareButtons}/>);
-	return (
-		<div className='x-live-blog-wrapper'>
-			{postElements}
-		</div>
-	);
-});
 
-export { LiveBlogWrapper };
+const LiveBlogWrapper = liveBlogWrapperActions(function LiveBlogWrapper ({ posts = [], articleUrl, showShareButtons }) {
```
ker-an, line 34: I'm a bit confused by this line and how `LiveBlogWrapper` calls itself here but let me take a look at `liveBlogWrapperActions`!

PR #517, opened 2020-08-24
Modify posts array instead of creating a new array when updating state

tbergmen:
This will improve performance and memory usage.

### DIFF
--- a/components/x-live-blog-wrapper/src/LiveBlogWrapper.jsx
+++ b/components/x-live-blog-wrapper/src/LiveBlogWrapper.jsx
@@ -5,32 +5,32 @@ import { listenToLiveBlogEvents } from './LiveEventListener';
 
 const withLiveBlogWrapperActions = withActions({
 	insertPost (post) {
-		return ({ posts }) => {
-			posts.unshift(post)
+		return (props) => {
+			props.posts.unshift(post);
 
-			return { posts };
+			return props;
 		};
 	},
 
 	updatePost (updated) {
-		return ({ posts }) => {
-			const index = posts.findIndex(post => post.postId === updated.postId);
+		return (props) => {
+			const index = props.posts.findIndex(post => post.postId === updated.postId);
 			if (index >= 0) {
-				posts[index] = updated;
+				props.posts[index] = updated;
 			}
 
-			return { posts };
+			return props;
 		};
 	},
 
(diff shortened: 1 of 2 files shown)


andygout (approved):

components/x-live-blog-wrapper/src/LiveBlogWrapper.jsx
```diff
@@ -6,26 +6,31 @@ import { listenToLiveBlogEvents } from './LiveEventListener';
 const withLiveBlogWrapperActions = withActions({
 	insertPost (post) {
 		return ({ posts }) => {
-			const updatedPosts = [ post, ...posts ];
+			posts.unshift(post)
```
simonplend, line 9: Missing `;`

components/x-live-blog-wrapper/src/__tests__/LiveBlogWrapper.test.jsx
```diff
@@ -32,3 +32,53 @@ describe('x-live-blog-wrapper', () => {
...
+		// liveBlogActions are not exported from the module, but we can access them via
+		// the props of LiveBlogWrapper component.
+		const liveBlogWrapper = LiveBlogWrapper({});
+		actions = liveBlogWrapper.props.actions;
+	});
+
+	it('inserts a new post to the top of the list', () => {
+		const post3 = {
+			postId: '3'
+		}
```
simonplend, line 52: Missing `;`

components/x-live-blog-wrapper/src/LiveBlogWrapper.jsx
```diff
@@ -6,26 +6,31 @@ import { listenToLiveBlogEvents } from './LiveEventListener';
...
 
-			return { posts: updatedPosts };
+			return { posts };
```
simonplend, line 11: Rather than creating and returning a new props object at the end of each of these action methods, I'm wondering if this might be a little clearer:

```javascript
return (props) => {
	props.posts.unshift(post);

	return props;
};
```

My thinking is that this makes it explicit that the argument to this method is a props object and that we're mutating the posts in place.

What do you think?

simonplend (approved):
This looks great :+1:

simonplend (approved):
Super :ok_hand:

PR #519, opened 2020-08-26
Fix height of x-live-blog-post component

simonplend:
In order for the `Element.clientHeight` property of the `x-live-blog-post` component's main container element to correctly reflect the height of all child elements, including their margins, we need to ensure that the container fully wraps all elements.

Related pull request: https://github.com/Financial-Times/next-article/pull/3929


ker-an (approved):

GlynnPhillips (approved):
This change is truly a work of art

PR #522, opened 2020-09-01
x-teaser hi-dpi images

invisiblegaudi:
`v1.5.0-beta` x-teaser

part of ticket: https://financialtimes.atlassian.net/browse/CON-330

allowing for additional `srcset` attribute for loading of images on hi-dpi displays

No impact on page performance was seen (finish time not an indicator due to lazy loading)

Tested on the How To Spend It section and saw resource load increase by 1.2MB, from 6.8MB to 8.0MB total
Screenshot 2020-09-08 at 17 43 06
Screenshot 2020-09-08 at 17 44 10


components/x-teaser/src/Image.jsx
```diff
@@ -17,20 +17,24 @@ const aspectRatio = ({ width, height }) => {
...
-	<img className="o-teaser__image" src={src} alt="" />
+const NormalImage = ({ src, srcSet }) => (
+	<img className="o-teaser__image" src={src} srcSet={srcSet} alt="" />
 );
 
-const LazyImage = ({ src, lazyLoad }) => {
+const LazyImage = ({ src, srcSet, lazyLoad }) => {
 	const lazyClassName = typeof lazyLoad === 'string' ? lazyLoad : '';
-	return <img className={`o-teaser__image ${lazyClassName}`} data-src={src} alt="" />;
+	return <img className={`o-teaser__image ${lazyClassName}`} data-src={src} srcSet={srcSet} alt="" />;
```
taraojo, line 26: Could we end up loading one of the `srcSet` images here even though they should be lazy loaded? Should this also be a data attribute that gets swapped when it's coming into the viewport?

components/x-teaser/src/concerns/image-service.js
```diff
@@ -1,14 +1,19 @@
...
+});
 
 /**
  * Image Service
  * @param {String} url
  * @param {Number} width
  * @param {String} options
  */
-export default function imageService(url, width, options) {
+export default function imageService(url, width, addOptions, overrideOptions={}) {
```
taraojo, line 14: Maybe not one for now, but it seems that `addOptions` and `overrideOptions` are almost the same. I wonder if we should replace `addOptions` with `overrideOptions` as the way you've added `overrideOptions` is more flexible and handles both scenarios. What do you think?

It looks like we're only using the previous options argument in one other place (Headshot component) to add a tint which could also be done with `overrideOptions`.

taraojo:
Hey John, this is looking cool 😎  
I've added a few notes, the main one is around how the lazy loading will work with `srcSet`. Could you also update the readme in the `x-teaser` directory to document the new `showHiResImages` prop?

components/x-teaser/src/concerns/image-service.js
invisiblegaudi: thanks for checking that, I wasn't 100% confident that it was the only dependency, yeah it'd be nice be rid of addOptions altogether. Not entirely out of the scope if we can be sure nothing else will break!

kavanagh, line 14: I don't really get why two different but almost identical params are needed? Perhaps you could refactor the other call sites to pass in an object rather than a string for this argument?

I like the use of the URLSearchParams class, although you should get an instance of one when you create an instance of a URL. so you could do something like this...

```js
const result = new URL(`${BASE_URL}/${encodeURIComponent(url)}`);
if (typeof options === 'string') { // might not be needed if this is refactored?
 options = Object.fromEntries(options.split('&').map(s => s.split('=')))
}
result.search = new URLSearchParams({...OPTIONS, ...options })
result.searchParams.set('width', width);
return result.href;
```
This ensures that a query param isn't set twice (which I don't think is valid for the image service).

components/x-teaser/src/Image.jsx
```diff
@@ -17,20 +17,24 @@ const aspectRatio = ({ width, height }) => {
 	return null;
 };
 
-const NormalImage = ({ src }) => (
-	<img className="o-teaser__image" src={src} alt="" />
+const NormalImage = ({ src, srcSet }) => (
+	<img className="o-teaser__image" src={src} srcSet={srcSet} alt="" />
```
kavanagh, line 21: If `srcSet` is null then the attribute will be empty.

```js
NormalImage({ src: 'foo', srcSet: null }) // `<img className="o-teaser__image" src="foo" srcSet="" alt="" />
```

I think it shouldn't be set if the value is falsey.

components/x-teaser-timeline/stories/content-items.json
```diff
@@ -42,7 +42,8 @@
       "height": 1152
```
kavanagh, line 42: Surprised to see a diff for snapshots in the x-teaser-timeline directory. Is this necessary?

components/x-teaser/src/Image.jsx
```diff
@@ -17,20 +17,24 @@ const aspectRatio = ({ width, height }) => {
...
 };
 
-export default ({ relativeUrl, url, image, imageSize, imageLazyLoad, ...props }) => {
+export default ({ relativeUrl, url, image, imageSize, imageLazyLoad, showHiResImages, ...props }) => {
 	const displayUrl = relativeUrl || url;
 	const useImageService = !(image.url.startsWith('data:') || image.url.startsWith('blob:'));
 	const imageSrc = useImageService ? imageService(image.url, ImageSizes[imageSize]) : image.url;
 	const ImageComponent = imageLazyLoad ? LazyImage : NormalImage;
+	const srcSet = showHiResImages && useImageService ? [
+		imageService(image.url, ImageSizes[imageSize], null, {dpr:1}),
```
kavanagh, line 35: Is dpr=1 actually needed? Wouldn't the browser use the `src` attribute if the screen density isn't high enough? Or does the browser need this for some other reason?

Also, have you had a look to see it's worth also altering compression/quality settings in addition to the dpr?

kavanagh (changes requested):
Could you please also add documentation to the component readme. I would also expect to see changes in components/x-teaser/storybook so this is previewable in storybook.

Also thinking about the release version, I think it might be worth this having a beta version to allow us to test the impact of this in production. 

Have you got any data on how this might impact the html size and image sizes?

components/x-teaser/src/Image.jsx
invisiblegaudi: I changed this as it seems that dpr=2 was used to cover hi-dpi and downscaled via CSS for normal size... I presumed it might help performance not to downscale but also using the image service as intended. A nice side-effect has been it makes it helps to see when the browser is using the default image size and not the hi-res.

from image service:
> Device Pixel Ratio, i.e. number of image pixels per CSS pixel. The default is 1, which gives the standard 1:1 ratio of image pixels to CSS pixels. Set to 2 for high-DPI ("Retina") displays. Request to the image service with width=100&dpr=2 will output image 200 pixels wide, but optimized to be displayed with CSS width:100px

components/x-teaser/src/concerns/image-service.js
invisiblegaudi: this is nice code 😎

components/x-teaser-timeline/stories/content-items.json
invisiblegaudi: it was breaking the build, the appearance of `scrSet` was causing the snapshot tests to fail

components/x-teaser/src/Image.jsx
invisiblegaudi: It's `dpr:2` by default in the image service handler, but this is scaled down for non-hi-density displays, with the intention of covering both, I think.
By setting `dpr:1` we should be downloading smaller images for non-hi-density displays and easing the load on the browser caused by this scaling
This behavour is recommended by the image service:
> Device Pixel Ratio, i.e. number of image pixels per CSS pixel. The default is 1, which gives the standard 1:1 ratio of image pixels to CSS pixels. Set to 2 for high-DPI ("Retina") displays. Request to the image service with width=100&dpr=2 will output image 200 pixels wide, but optimized to be displayed with CSS width:100px

Note that browsers don't automatically adjust intrinsic size of images with higher DPI, so you need'll need to use some "responsive image" solution, e.g. set explicit width or height, use image-set() or srcset attribute.

Unless this is intended to increase the quality of the image via the scaling?

apaleslimghost (changes requested):
i did some investigation of this a while ago. i discovered that, at the sizes usually used on the homepage, although using `dpr=1` for the devices that wanted it gave a ~5% download size reduction, that was totally offset by the addition of ~60 `srcset` attributes on the front page. we ended up using `dpr=2` for everything as that was the best tradeoff between size and quality.

since adding this attribute gives a `1.5MB` increase in download size, i don't think it's worth it for a small data saving on some devices, especially since the devices that would be using `1x` images are probably desktops (where we don't care about data)

components/x-teaser/src/Image.jsx
invisiblegaudi: reverted to dp=2 for all (set 2x to use highest images quality)

components/x-teaser/src/Image.jsx
```diff
@@ -17,20 +17,33 @@ const aspectRatio = ({ width, height }) => {
...
+	? <img className={`o-teaser__image ${lazyClassName}`} data-src={src} srcSet={srcSet} data-srcSet={srcSet} alt="" />
+	: <img className={`o-teaser__image ${lazyClassName}`} data-src={src} alt="" />;
 };
 
-export default ({ relativeUrl, url, image, imageSize, imageLazyLoad, ...props }) => {
+export default ({ relativeUrl, url, image, imageSize, imageLazyLoad, imageHighDensity, ...props }) => {
 	const displayUrl = relativeUrl || url;
 	const useImageService = !(image.url.startsWith('data:') || image.url.startsWith('blob:'));
+	const optionsHighDensity = {
+		quality: 'highest',
```
apaleslimghost, line 38: to me there's absolutely no perceptible difference between `highest` and `medium` quality on 2x screens. i don't think this warrants the extra markup output, download size, and code complexity.

## `medium`
screenshot of how to spend it with medium quality images


## `highest`
screenshot of how to spend it with highest quality images

apaleslimghost (changes requested):

apaleslimghost (changes requested):
i don't understand what the benefit of this PR is without changing image quality or resolutions? in its current state it's adding `srcset` with the same image URLs as the `src` attributes, which would have no effect (apart from increasing markup output size)?

apaleslimghost:
we're [already setting `dpr=2`](https://github.com/Financial-Times/x-dash/blob/master/components/x-teaser/src/concerns/image-service.js#L2), with `o-teaser` setting the width, that's what i'm saying. this PR doesn't do anything apart from add redundant `srcset` attributes

invisiblegaudi:
does o-teaser explicitly set the image width for hi-dpi displays?

apaleslimghost:
it sets _a_ width no matter what the size of the image is. this width works for 1x and 2x image sources.

invisiblegaudi:
> i did some investigation of this a while ago. i discovered that, at the sizes usually used on the homepage, although using `dpr=1` for the devices that wanted it gave a ~5% download size reduction, that was totally offset by the addition of ~60 `srcset` attributes on the front page. we ended up using `dpr=2` for everything as that was the best tradeoff between size and quality.
> 
> since this PR gives a 20% increase in download size, i don't think it's worth it for a small data saving on some devices, especially since the devices that would be using `1x` images are probably desktops (where we don't care about data)

It seems that this ticket arose from the need to mitigate certain specific situations, namely the large teaser images in the HTSI section, where a loss of quality _is_ detectable (I have no specific examples unfortunately), so allowing a teaser to have it's image's quality set to highest is desirable in this case...

invisiblegaudi:
> it sets _a_ width no matter what the size of the image is. this width works for 1x and 2x image sources.

in this case I can remove `srcSet`... I see the intrinsic size is there, my mistake!

apaleslimghost:
that looks like we can close this PR then, since the `imageHighDensity` option won't be doing anything without `srcSet` in there?

if you still want increased image quality for the large HTSI images maybe we could instead add `quality=highest` (or `high`, which might be good enough?) only to `XXL` images? keeping the surface area of additional options low would be preferred and i think all `XXL` images would benefit from that (and they're used rarely enough that it wouldn't have much overall impact)

components/x-teaser/src/concerns/image-service.js
```diff
@@ -8,7 +12,8 @@ const OPTIONS = ['source=next', 'fit=scale-down', 'dpr=2'];
  * @param {String} options
  */
 export default function imageService(url, width, options) {
-	const encoded = encodeURIComponent(url);
-	const href = `${BASE_URL}/${encoded}?${OPTIONS.join('&')}&width=${width}`;
-	return options ? href + '&' + options : href;
+  const imageSrc = new URL(`${BASE_URL}/${encodeURIComponent(url)}`);
```
apaleslimghost, line 15: officially, `x-teaser` supports any Node version from 6; `URL` is only available on the global object from v10. i don't know if anything consuming `x-teaser` is on older Node versions, but so we don't have to remove support for those versions (which would be a breaking change), you should add `const {URL, URLSearchParams} = require('url')`

components/x-teaser/src/Image.jsx
```diff
@@ -26,10 +26,11 @@ const LazyImage = ({ src, lazyLoad }) => {
 	return <img className={`o-teaser__image ${lazyClassName}`} data-src={src} alt="" />;
 };
 
-export default ({ relativeUrl, url, image, imageSize, imageLazyLoad, ...props }) => {
+export default ({ relativeUrl, url, image, imageSize, imageLazyLoad, imageHighestQuality, ...props }) => {
 	const displayUrl = relativeUrl || url;
 	const useImageService = !(image.url.startsWith('data:') || image.url.startsWith('blob:'));
-	const imageSrc = useImageService ? imageService(image.url, ImageSizes[imageSize]) : image.url;
+  const options = imageSize === 'XXL' && imageHighestQuality ? { quality: 'highest' } : {};
```
apaleslimghost, line 32: please fix the indentation to use tabs

components/x-teaser/src/concerns/image-service.js
```diff
@@ -1,5 +1,9 @@
 const BASE_URL = 'https://www.ft.com/__origami/service/image/v2/images/raw';
-const OPTIONS = ['source=next', 'fit=scale-down', 'dpr=2'];
+const OPTIONS = Object.freeze({
```
apaleslimghost, line 2: what's the purpose of `Object.freeze` here? there's no way for this object to be mutated,, it seems unnecessary

apaleslimghost (changes requested):
a few things to fix, but we're close! thanks for your patience

components/x-teaser/src/concerns/image-service.js
invisiblegaudi: ok, i guess it's a bit of a future-proof

components/x-teaser/src/Image.jsx
invisiblegaudi: 😱

components/x-teaser/src/concerns/image-service.js
andygout: Once the upgrade of all Customer Products apps to Node v12 is complete, and if all the App's repos likewise have a minimum of v10, we can then use the `URL` from the global object with confidence.

components/x-teaser/storybook/article.js
```diff
@@ -36,6 +36,7 @@ exports.knobs = [
 	'showImage',
 	'image',
 	'imageSize',
+  'imageHighestQuality',
```
apaleslimghost, line 39: it wasn't just that one line that had spaces, all the indentation in the PR is! 😬

invisiblegaudi: 😱 😱

apaleslimghost (approved):

PR #523, opened 2020-09-03
Updates to use this component at the client side

tbergmen:
There are two major updates to use this component at the client side:

## 1. Add a property named data-x-live-blog-wrapper-id

We need to render a property named `data-x-live-blog-wrapper-id` in the main element. This is now used to identify the element at the client side for live updates.

Previously we used `data-x-dash-id` for this purpose, but that property is only rendered when using this component with server side rendering.

## 2. Pass actions object to listenToLiveBlogEvents function

With server side rendering and hydrating at the client side, component actions can be triggered by dispatching a custom event named `x-interaction.trigger-actions`. This method can't be used with client side rendering. Therefore we need pass the `actions` object to `listenToLiveEvents` function as an argument.

`actions` object can be gathered by passing a callback function to the component in a property named actionsRef.

For more information, see https://github.com/Financial-Times/x-dash/tree/master/components/x-interaction#triggering-actions-externally

Usage of `listenToLiveEvents` function is now documented in more detail in the readme file.

### DIFF
--- a/components/x-live-blog-wrapper/readme.md
+++ b/components/x-live-blog-wrapper/readme.md
@@ -34,6 +34,24 @@ All `x-` components are designed to be compatible with a variety of runtimes, no
 
 [jsx-wtf]: https://jasonformat.com/wtf-is-jsx/
 
+### Client side rendering
+This component can be used at the client side. To access the actions, a function needs to be passed into the actionsRef property of the LiveBlogWrapper element.
+
+```jsx
+import { LiveBlogWrapper } from '@financial-times/x-live-blog-wrapper';
+
+const actionsRef = actions => {
+    // Use actions to insert, update and delete live blog posts
+};
+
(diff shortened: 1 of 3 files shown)


ker-an (approved):
Thanks for the detailed descriptions, it definitely helped me better understand this 😄

components/x-live-blog-wrapper/readme.md
```diff
@@ -64,13 +82,59 @@ hydrate();
 ```
 
 ### Live updates
-This component exports a function named `listenToLiveBlogEvents` which is used for listening to client side live blog updates. This function should be called after hydrating the component if it is rendered at the server side. 
+This component exports a function named `listenToLiveBlogEvents` which is used for listening to client side live blog updates.
```
GlynnPhillips, line 85: Is it worth mentioning here that these come in the form of server sent events sent by next-live-event-api?

GlynnPhillips (approved):
Looks great to me. I added one small comment about some extra detail but I don't hold a strong opinion about this

PR #524, opened 2020-09-11
Update x-live-blog-post arguments

GlynnPhillips:
This component was originally produced whilst we were receiving data
from the wordpress cms and we didn't have a schema for the Spark CMS
yet.

Whilst we are running both systems side by side I have added some new
arguments that will in the longer term replace some of the old
arguments.

I have made this change within the x-dash component to avoid having to
map property names in the systems that use this component.

This allows us to release this change as a minor release rather than a
major breaking change and then release a major when we remove these
arguments in the future.

**Argument mapping**

New argument      | Old argument 
------------------|---------------
`id`                        | `postId`
`bodyHTML`        | `content`
`publishedDate ` | `publishedTimestamp`

### DIFF
--- a/components/x-live-blog-post/readme.md
+++ b/components/x-live-blog-post/readme.md
@@ -36,12 +36,18 @@ All `x-` components are designed to be compatible with a variety of runtimes, no
 
 ### Properties
 
+Deprecated properties should only be used when data comes from the Wordpress CMS.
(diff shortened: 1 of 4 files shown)


tbergmen (approved):
👍

tbergmen (approved):

PR #525, opened 2020-09-11
Add a data-x-component attribute to live blog post

GlynnPhillips:
Add a `data-component="x-live-blog-post"` to each instance of the
component which can then be used by client side to target the component.

An example for this component is that it adds share buttons but the
parent application is responsible for initilising the o-share. This
allows us to query every post and initilise o-share for that instance.


tbergmen (dismissed):
👍

PR #530, opened 2020-09-21
Fix live-blog-post padding

tbergmen:
The bottom padding for the live blog post used to rely on sharing buttons to be visible. In the ft-app we don't display the share buttons in this component and the padding doesn't look right.

This fixes the bottom padding regardless of displaying or hiding the share buttons.

**Before:**
Screenshot 2020-09-21 at 11 48 20

**After:**
Screenshot 2020-09-21 at 11 48 43

### DIFF
--- a/components/x-live-blog-wrapper/storybook/index.jsx
+++ b/components/x-live-blog-wrapper/storybook/index.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { storiesOf } from '@storybook/react';
 import { withKnobs, text } from '@storybook/addon-knobs';
 import { LiveBlogWrapper } from '../src/LiveBlogWrapper';
+import '../../x-live-blog-post/dist/LiveBlogPost.css';
 
 const defaultProps = {
 	message: 'Test',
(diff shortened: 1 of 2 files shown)


components/x-live-blog-wrapper/storybook/index.jsx
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 import { storiesOf } from '@storybook/react';
 import { withKnobs, text } from '@storybook/addon-knobs';
 import { LiveBlogWrapper } from '../src/LiveBlogWrapper';
+import '../../x-live-blog-post/dist/LiveBlogPost.css';
```
simonplend, line 5: How was the styling of the component in Storybook working without this?

simonplend (approved):

components/x-live-blog-wrapper/storybook/index.jsx
tbergmen: It wasn't working in the x-live-blog-wrapper storyboard.

PR #537, opened 2020-10-09
Fix live blog post blockquote styling on FT App

tbergmen:
We add `article--body` class to the wrapper div of live blog posts to fix the rendering of blockquotes in the FT App.

Before:
Screenshot 2020-10-09 at 14 56 04

After:
Screenshot 2020-10-09 at 14 56 24


ker-an (approved):

ker-an:
This is interesting because on FT.com the quotes are styled differently from the App. On FT.com live blogs, they are intended to look like this: 
image

Something we can bring up with Josh at some point but it's not super urgent.

tbergmen:
> This is interesting because on FT.com the quotes are styled differently from the App. On FT.com live blogs, they are intended to look like this:
> image
> 
> Something we can bring up with Josh at some point but it's not super urgent.

I don't know about the reasons or background but ft-app overrides a few of the content styles. This is one of them.

PR #541, opened 2020-10-14
Add `article--content` class to live blog wrapper

ker-an:
This class is required by the App. It tells `o-topper` where to place the live blog content.

`o-topper` uses grid template areas [here](https://github.com/Financial-Times/o-topper/blob/fb29a5620bda23aa4c3fb1fd81c99f341c12b337/src/scss/_grid.scss#L10) and `ft-app` sets the `grid-area` based on the `article--content` class [here](https://github.com/Financial-Times/ft-app/blob/c6d89334b325edc2dfa625b878c19ba915d0fa4a/lib/css/article/main.scss#L119).

Before
image

After
image

### DIFF
--- a/components/x-live-blog-wrapper/src/LiveBlogWrapper.jsx
+++ b/components/x-live-blog-wrapper/src/LiveBlogWrapper.jsx
@@ -60,7 +60,10 @@ const BaseLiveBlogWrapper = ({ posts = [], articleUrl, showShareButtons, id, liv
 	);
 
 	return (
-		<div className='x-live-blog-wrapper' data-live-blog-wrapper-id={id} ref={liveBlogWrapperElementRef}>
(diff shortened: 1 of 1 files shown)


tbergmen (approved):

simonplend (commented):
I think we should try and avoid adding FT.com or App specific classes into our `x-live-blog-*` components as the styles for those classes don't exist within those components. My understanding is that these components should be self-contained (JSX + CSS).

If it's possible to put the `article--content` class on an element which wraps `<LiveBlogWrapper />`, I think that would be the ideal scenario.

ker-an:
@simonplend 🤦🏽 yah @tbergmen ignore me, this can be done in my ft-app PR and it works there too. Gonna close!

PR #544, opened 2020-10-28
CPP-350 - Update linting to align with Page Kit and run on all files

nickcolley:
1. Add linting configuration used in Page Kit.
2. Run linting on all files.

Platforms team is trying to make x-dash and Page Kit's tooling and style more consistent.

By adding linting in this way reviewers of pull requests can avoid having as many conversations about style as most issues will be fixed automatically pre-commit.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,9 @@
+{
+  "printWidth": 110,
+  "semi": false,
+  "singleQuote": true,
+  "bracketSpacing": true,
+  "arrowParens": "always",
+  "jsxBracketSameLine": true,
+  "trailingComma": "none"
+}
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,4 +1,5 @@
 module.exports = {
+	parser: '@typescript-eslint/parser',
 	env: {
 		node: true,
 		browser: true,
@@ -12,7 +13,8 @@ module.exports = {
 		// https://github.com/yannickcr/eslint-plugin-react
 		'plugin:react/recommended',
 		// https://github.com/evcohen/eslint-plugin-jsx-a11y
-		'plugin:jsx-a11y/recommended'
+		'plugin:jsx-a11y/recommended',
+		'prettier'
 	],
 	parserOptions: {
 		ecmaFeatures: {
@@ -53,4 +55,4 @@ module.exports = {
 			}
 		}
 	]
-};
+}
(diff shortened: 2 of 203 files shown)


apaleslimghost:
i think it would be best to run prettier against every existing file and commit that rather than wait for it to trickle in as people edit them, what do you think?

nickcolley:
@apaleslimghost yeah just giving that a go now before raising for realsies.

nickcolley:
Hmm eslint is getting sad with Prettier since Prettier is adding spaces and then Eslint doesnt want spaces. Will need to keep poking.

apaleslimghost:
oh, we've got `eslint-config-prettier` _installed_ but it's not actually referenced in the config, adding that should make `eslint` not care about formatting at all

nickcolley:
When I read the docs it recommended to avoid that way of doing things which is why I ended up trying prettier-eslint, let me know what you think.

https://prettier.io/docs/en/integrating-with-linters.html#notes

nickcolley:
Sorry I'm confusing `config` and `plugin`.

apaleslimghost:
that's `eslint-plugin-prettier` not `eslint-config-prettier`, they're different things 😅 `eslint-plugin-prettier` runs prettier as part of eslint, which is what they're recommending against; `eslint-config-prettier` disables eslint rules that conflict with prettier, which they do suggest using in that doc. i don't think `prettier-eslint` will give us any benefit over running them separately with that config?

apaleslimghost:
oh you commented before me lol. yeah let's do that

nickcolley:
Synk failed with an error that suggests to try making a new pull request:

https://support.snyk.io/hc/en-us/articles/360000920678-Failed-to-detect-issues

.prettierrc
```diff
@@ -0,0 +1,9 @@
+{
```
antonsamper, line 1: is there a common prettier config that is used across FT apps? similar to the `eslint` config?

nickcolley: This is copied from Page Kit, I'm not aware of other projects using prettier but if there are we should consider publishing a common config like we do with eslint.

Probably out of scope for this given what we know atm.

antonsamper: It looks like a lot of changes comes from removing the semi-colon at the end of the line. Should we keep it?

nickcolley: I personally prefer to keep the configuration to stock and not change it as style is subjective.

Semi-colons or no colons is controversial so the intention here is to mirror Page Kit, so I'd say any discussion about our general style should be had when/if we consider a single prettier config for FT.

apaleslimghost (approved):

PR #545, opened 2020-10-28
Update to Node v12

andygout:
This update is made incorporating the points raised by @simonplend in this issue: https://github.com/Financial-Times/tooling-helpers/issues/60.

Essentially, if we aim to support (or state compatibility for) the packages on multiple Node versions then we should be testing the packages on all of those Node versions as part of the CI process accordingly. However, that would result in very protracted build times.

The benefit of having made the Node upgrades across Customer Products so that all apps are on the same version (v12) is that we only need to worry about supporting packages on a single Node version. I have checked with @rowanbeentje and all the FT App repos that consume `x-dash` are also all on Node v12.

This PR amends various Node values and README statements to make clear that only v12 is currently supported for the packages, and updates the CircleCI image used so that packages are only tested in a Node v12 environment.

---

## 

- Discuss features first
- Update the documentation
- Reference other issues
- Follow the code style

### DIFF
--- a/packages/x-rollup/src/rollup-config.js
+++ b/packages/x-rollup/src/rollup-config.js
@@ -28,7 +28,7 @@ module.exports = ({ input, pkg }) => {
 				plugins: [
 					babel(
 						babelConfig({
-							targets: [{ node: 6 }]
+							targets: [{ node: 12 }]
 						})
 					),
 					...plugins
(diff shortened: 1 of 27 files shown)


packages/x-rollup/src/rollup-config.js
```diff
@@ -44,11 +44,17 @@ module.exports = ({ input, pkg }) => {
 				input,
 				external,
 				plugins: [
+<<<<<<< HEAD
```
emortong, line 47: @andygout unresolved merge conflict

andygout: Good catch. I will just do an amend commit on top of the existing one.

emortong (approved):

PR #551, opened 2020-11-11
Fix bugs with triggering live blog wrapper actions

ker-an:
I am in the process of rendering live updates to FT.com. While doing this, I came across a couple of bugs which prevented us from triggering the actions to create, update and delete a post. I broke up the fixes into separate commits and explained the "why" in the commit message. But long story short, we were using the wrong property name for accessing the post ID and we weren't bubbling an event to the parent when we should have been.

We are now able to render a live update to FT.com:

image


tbergmen (approved):
👍

nickcolley (approved):

simonplend (approved):
Solid fixes, solid commit messages :raised_hands:

PR #554, opened 2020-11-16
CI-392 Fix bug with accessing post ID

ker-an:
I forgot to update how we access post ID in this commit: https://github.com/Financial-Times/x-dash/pull/551/commits/be96eb0e44b6b0ef70a3112716fc768b302daf36.

Essentially, `next-live-blog-trigger` renamed the post ID property from `id` to `postId` which means when we receive the event, we need to access the post id as `updated.postId`. 

I tested this out in `next-article` and the most recent blog post was successfully deleted.

Before
image

After
image


tbergmen (approved):
👍

PR #555, opened 2020-11-16
Remove failing Snyk CircleCI command

ker-an:
A Snyk error is preventing me from releasing a new version of `x-dash`.
Snyk is aware of the issue and Andy is working with them to debug.
Because we don't know if there will be a fix soon, I am going to remove
the Snyk command, publish a new release and then re-add the Snyk
command.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -124,7 +124,6 @@ jobs:
       - run:
           name: shared-helper / npm-store-auth-token
           command: .circleci/shared-helpers/helper-npm-store-auth-token
-      - run: npx snyk monitor --org=customer-products --project-name=Financial-Times/x-dash --prune-repeated-subdependencies
       - run:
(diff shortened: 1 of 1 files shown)


nickcolley (approved):

andygout:
It turns out that this issue is caused by the root directory `package.json` not having any dependencies (it only has devDependencies, which are not monitored by Snyk).

With the help of the Snyk Support Team I have revised the command whereby it monitors all the `package.json` files, i.e. all that exist in subdirectories (which it currently does not do) except for the `/private` directory. The last hurdle is how to exclude the root directory `package.json` - I made a few attempts but all have failed, so I have asked the Snyk Support Team for further advice.

ker-an:
@andygout thank you for the update! let me know if you need me to do some debugging to help you out 😄

PR #557, opened 2020-11-17
Remove Snyk monitor command from CircleCI config

andygout:
This Snyk command is not currently doing anything (in either of its two appearances). I learned from the Snyk Support Team that the `snyk monitor` command only looks at `dependencies`, not `devDependencies`. The `package.json` in the root dir of this repo only has `devDependencies` and so the command fails because it cannot find any `dependencies` to examine.

So there is no net change from removing this command, other than CI builds will no longer fail owing to this Snyk-related failure.

I am currently corresponding with the Snyk Support Team to get a command that will monitor all the `package.json` files in subdirectories (which would be an increase in scope of what the existing code sought to achieve), but ignore the one in the root dir given the problems it is causing us, as well as ignoring the one in the `/private` directory which also fails because its node_modules are not installed as the others are. This is how far I have got:

`npx snyk monitor --org=customer-products --all-projects --exclude=private`

Somehow this is still scoped exclusively to `x-dash` despite not declaring `--project-name=Financial-Times/x-dash`, which is useful given that the combination of the `project-name` and `all-projects` flags is currently not supported.

However, this command still tries to monitor the root dir `package.json` (and so fails the CI workflow on that basis). I have tried the following to exclude that file too, but all these attempts failed for the described reasons:
- `--exclude=.,private` results in `Internal error (reference: 3b3cea61-008d-4b5e-8288-ae84875fc843)` (i.e. the original error that caused us to start removing this Snyk command)

- `--exclude=package.json,private` does not throw an error but it is only monitoring one of the subdirectories' package.json (and it seems relevant that this is the only subdirectory that has a package-lock.json, indicating that this command ignores all `package.json` files and only examines `package-lock.json` files).

- `--exclude=./package.json,private` results in error `The --exclude argument must be a comma separated list of directory names and cannot contain a path.`

I am waiting to hear back from the Snyk Support Team as to how to exclude the root dir `package.json` file.

---

## 

- Discuss features first
- Update the documentation
- Reference other issues
- Follow the code style

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -124,7 +124,6 @@ jobs:
       - run:
           name: shared-helper / npm-store-auth-token
           command: .circleci/shared-helpers/helper-npm-store-auth-token
-      - run: npx snyk monitor --org=customer-products --project-name=Financial-Times/x-dash --prune-repeated-subdependencies
       - run:
           name: Bump version
           command: npx athloi version ${CIRCLE_TAG}
@@ -139,7 +138,6 @@ jobs:
       - run:
           name: shared-helper / npm-store-auth-token
           command: .circleci/shared-helpers/helper-npm-store-auth-token
-      - run: npx snyk monitor --org=customer-products --project-name=Financial-Times/x-dash --prune-repeated-subdependencies
       - run:
           name: Extract tag name and version number
           command: |


antonsamper (approved):

andygout:
I received this response from Love Bhardwaj on the Snyk Support Team as to whether we can exclude the root dir `package.json` file but still monitor all the `package.json` files in subdirectories. The answer is no, we cannot.

> This is currently not possible. Doing so would logically mean that you are telling Snyk to not test the project directory itself and the contents of it. In other words, Snyk would ignore the directory itself and any other sub projects that may exist in subdirectories of the project. Which is equivalent to not testing the project in the first place.

> For example, if your project directory is called test and there are two subprojects in it `test-1` and `test-2`. Further `test-1` and `test-2` further have two sub-projects `test-3` and `test-4` respectively.

> ```
> test
> |-test-1->test-3
> |-test-2->test-4
> ```

> Now when you run `snyk test --all-projects --exclude=test-1` from the root directory i.e test. Snyk will scan directories test, `test-2`, and `test-4` while excluding `test-1` and `test-3`. If I were to go to the parent directory of test and run `snyk test --all-projects --exclude=test`. Snyk will exclude directory `test`, `test-1`, `test-2`, `test-3`, and `test-4` from testing. 

> So if it were possible to ignore the root directory itself then as mentioned earlier this would mean not running Snyk at all, which is logically invalid. What would be more suited here is an option to exclude a file instead of the directory from scanning, which unfortunately is not available at the moment.

PR #558, opened 2020-11-17
Add additional context for tracking

tbergmen:
We want to track post_id and button attributes when a user shares a live blog posts. To achieve that we use data-trackable-context- attribute on sharing buttons.


GlynnPhillips (approved):

GlynnPhillips:
Adding this comment here so we don't forget. We think theres a chance we don't need this functionality and that it might be wrong placed in here anyway.

We are keeping the PR for now but will close it soon if it turns out we are right and this is not needed

tbergmen:
Closing this PR because these attributes are not needed for this purpose.

PR #564, opened 2020-12-02
Give the x-live-blog-wrapper component a displayName - CI-423

GlynnPhillips:
Without this the component wouldn't work with x-interaction when the code is transpiled for production.

It was working with development code I believe because it was relying in on the standard Javascript name property of the function.

When the code was transpiled for production the functions were minified the function names were changed or made anonymous.

By hardcoding this displayName as a string it allows x-interaction to register the component and reference it by name.

### DIFF
--- a/components/x-live-blog-wrapper/src/LiveBlogWrapper.jsx
+++ b/components/x-live-blog-wrapper/src/LiveBlogWrapper.jsx
@@ -68,6 +68,10 @@ const BaseLiveBlogWrapper = ({ posts = [], articleUrl, showShareButtons, id, liv
 	)
 }
 
+// A displayName is required for this component
+// This enables the component to work with x-interaction
+BaseLiveBlogWrapper.displayName = 'BaseLiveBlogWrapper'
+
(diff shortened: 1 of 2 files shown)


ker-an (approved):
😫 I honestly tried this exact line of code but it didn't work for me 🤔

tbergmen (approved):
This looks good 👍 

I think it would be great to mention this in the x-interaction readme file so that others who use x-interaction is aware of this.

https://github.com/Financial-Times/x-dash/blob/master/components/x-interaction/readme.md

GlynnPhillips:
@ker-an - I have had weird cases where I have not seen it work as well and I think it was to do with my environment maybe 🤷‍♂️ I am this stage I am crossing my fingers a lot 🤞 and hoping it works in prod and if not we can continue to look further.

@tbergmen Thats a good suggestion - I was also thinking of updating all components that use x-interaction even if at the moment they don't have an issue it would be nice to avoid future issues. I will do this once I am 100% sure this works

apaleslimghost:
it might also be worth adding an useful error message if the component isn't registered when hydrating

PR #567, opened 2020-12-07
Allow `next-live-event-api` endpoint URL to be overriden

nickcolley:
Make it easier for us to test the component locally.

I tested this locally in `next-article` by building the component locally and configuring the plugin as is documented.


ker-an (approved):
nice

tbergmen (approved):

simonplend (approved):

PR #568, opened 2020-12-08
CPP-351 Editing x-dash components does not update Storybook

emortong:
[Ticket CPP-351](https://financialtimes.atlassian.net/browse/CPP-351)

When Webpack rebuilds on changes it checks files ending with `.jsx`, since some Storybook stories were importing components from `/dist` files ending in `.cjs`, these were not getting rebuilt. 

Changing Storybook so that it imports the `.jsx` components instead, made some build errors get thrown, which were fixed by upgrading `o-typography`, `o-grid`, `o-colors`. The changes in `TeaserTimeline.scss` are related to these version upgrades.

### DIFF
--- a/components/x-podcast-launchers/storybook/index.jsx
+++ b/components/x-podcast-launchers/storybook/index.jsx
@@ -1,4 +1,4 @@
-const { PodcastLaunchers } = require('../dist/PodcastLaunchers.cjs')
+import { PodcastLaunchers } from '../src/PodcastLaunchers'
 const { brand } = require('@financial-times/n-concept-ids')
 import React from 'react'
 import { Helmet } from 'react-helmet'
--- a/components/x-podcast-launchers/storybook/index.jsx
+++ b/components/x-podcast-launchers/storybook/index.jsx
@@ -1,5 +1,5 @@
 import { PodcastLaunchers } from '../src/PodcastLaunchers'
-const { brand } = require('@financial-times/n-concept-ids')
+import { brand } from '@financial-times/n-concept-ids'
 import React from 'react'
 import { Helmet } from 'react-helmet'
 import BuildService from '../../../.storybook/build-service'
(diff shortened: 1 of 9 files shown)


components/x-teaser-timeline/bower.json
```diff
@@ -3,8 +3,8 @@
   "private": true,
   "main": "dist/TeaserTimeline.es5.js",
   "dependencies": {
-    "o-colors": "^4.7.2",
-    "o-grid": "4.4.4",
-    "o-typography": "^5.7.5"
+    "o-typography": "^6.4.5",
+    "o-colors": "^5.3.0",
+    "o-grid": "^5.2.9"
```
apaleslimghost, line 8: is this `bower.json` only used for local development? i've noticed that the apps using `x-teaser-timeline` (`next-myft-page` and `ft-app`) both have different versions of these components, so this probably isn't actually a breaking change?

components/x-podcast-launchers/storybook/index.jsx
```diff
@@ -1,4 +1,4 @@
-const { PodcastLaunchers } = require('../dist/PodcastLaunchers.cjs')
+import { PodcastLaunchers } from '../src/PodcastLaunchers'
 const { brand } = require('@financial-times/n-concept-ids')
```
apaleslimghost, line 2: non-merge-blocking, is there any reason to keep these as `require` or should we standardise on `import`? if there's a reason it should probably be documented or commented?

components/x-teaser-timeline/bower.json
apaleslimghost: oh, i see, they're being used in the component's Sass files which are compiled to CSS before reaching the apps

apaleslimghost (dismissed):
looks good!

components/x-podcast-launchers/storybook/index.jsx
emortong: nope, there is no particular reason, at least for this case... it's leftover pre-migration code, I'll push the change

apaleslimghost (approved):

PR #569, opened 2020-12-10
Update breaking news label based on Spark data

GlynnPhillips:
The data we get from Spark and UPP for breaking news is `standout.breakingNews`
This change uses that property to render the breaking news label.

For now it also still uses the `isBreakingNews` property that is used for wordpress data.


tbergmen (approved):
👍

PR #574, opened 2021-01-20
CI-479: Check for existing posts before inserting new ones.

nickcolley:
We want to make sure that if a new post is received that has an id of one that already exists that we don't add it again, this will stop the user seeing multiples of the same post.

This makes this logic most resilient against issues upstream, for example the event stream sending multiple inserts, or inserts being sent due to race conditions.

### DIFF
--- a/components/x-live-blog-wrapper/src/LiveBlogWrapper.jsx
+++ b/components/x-live-blog-wrapper/src/LiveBlogWrapper.jsx
@@ -2,21 +2,28 @@ import { h } from '@financial-times/x-engine'
 import { LiveBlogPost } from '@financial-times/x-live-blog-post'
 import { withActions } from '@financial-times/x-interaction'
 import { listenToLiveBlogEvents } from './LiveEventListener'
+import { dispatchEvent } from './dispatchEvent'
 
 const withLiveBlogWrapperActions = withActions({
-	insertPost(post) {
+	insertPost(newPost) {
 		return (props) => {
-			props.posts.unshift(post)
+			// Check if the new post already exists in the page, so we don't end up with duplicates.
+			const newPostExists = props.posts.find((post) => post.id === newPost.id)
+			if (!newPostExists) {
+				props.posts.unshift(newPost)
+				dispatchEvent('LiveBlogWrapper.INSERT_POST', { post: newPost })
+			}
 
 			return props
 		}
 	},
 
-	updatePost(updated) {
+	updatePost(updatedPost) {
 		return (props) => {
-			const index = props.posts.findIndex((post) => post.id === updated.postId)
+			const index = props.posts.findIndex((post) => post.id === updatedPost.id)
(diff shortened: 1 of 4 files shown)


components/x-live-blog-wrapper/src/LiveBlogWrapper.jsx
```diff
@@ -4,9 +4,13 @@ import { withActions } from '@financial-times/x-interaction'
 import { listenToLiveBlogEvents } from './LiveEventListener'
 
 const withLiveBlogWrapperActions = withActions({
-	insertPost(post) {
+	insertPost(newPost) {
 		return (props) => {
-			props.posts.unshift(post)
+			// Check if the new post already exists in the page, so we don't end up with duplicates.
+			const newPostExists = props.posts.find((existingPost) => existingPost.id === newPost.id)
```
nickcolley, line 10: I have seen `post.postId` and `post.id` used in this codebase, which is correct?

components/x-live-blog-wrapper/src/__tests__/LiveBlogWrapper.test.jsx
```diff
@@ -74,6 +74,19 @@ describe('liveBlogWrapperActions', () => {
 		expect(posts[0].id).toEqual('3')
 	})
 
+	it('does not insert a new post if a duplicate', () => {
+		const post2 = {
```
nickcolley, line 78: I'm going to change this logic so it intentionally duplicates an element in the original posts array

GlynnPhillips (changes requested):
**Change Request:** This is almost perfect but unfortunately in the LiveEventListener.js file it also dispatches an event called `LiveBlogWrapper.INSERT_POST` I think this shouldn't happen if we don't actually insert the post

https://github.com/Financial-Times/x-dash/blob/master/components/x-live-blog-wrapper/src/LiveEventListener.js#L86

nickcolley:
@GlynnPhillips good catch, it looks like the update and delete ones would have the same issue too... 🤔

components/x-live-blog-wrapper/src/LiveBlogWrapper.jsx
GlynnPhillips: I think this is a wordpress to spark cross over thing. I can't be certain if each case is using the right one without reviewing them all but I believe its that `postId` was a wordpress thing and `id` is Spark.

There might be a reason why we didn't standardise them but I am not sure 😬

nickcolley: I'm going to update to `postId` for now since it seems the most current thing:

https://github.com/Financial-Times/x-dash/commit/be96eb0e44b6b0ef70a3112716fc768b302daf36

nickcolley:
Closed in favour of https://github.com/Financial-Times/x-dash/pull/576

PR #575, opened 2021-01-20
Check for existence of image as soon as image teaser function starts

jkerr321:
Check for existence of `image` as soon as image teaser function starts so that the line
```js
const useImageService = !(image.url.startsWith('data:') || image.url.startsWith('blob:'))
```
doesn't error if `image` is undefined. This has happened twice on the homepage when a gif has been used as the main image. when this errors on the home page the whole home page returns a 500.

Now when image is null stories render without images rather than the whole page going down:
image

### DIFF
--- a/components/x-teaser/src/Image.jsx
+++ b/components/x-teaser/src/Image.jsx
@@ -27,28 +27,27 @@ const LazyImage = ({ src, lazyLoad }) => {
 export default ({ relativeUrl, url, image, imageSize, imageLazyLoad, imageHighestQuality, ...props }) => {
 	if (!image || (image && !image.url)) {
 		return null
-	} else {
-		const displayUrl = relativeUrl || url
-		const useImageService = !(image.url.startsWith('data:') || image.url.startsWith('blob:'))
-		const options = imageSize === 'XXL' && imageHighestQuality ? { quality: 'highest' } : {}
-		const imageSrc = useImageService ? imageService(image.url, ImageSizes[imageSize], options) : image.url
-		const ImageComponent = imageLazyLoad ? LazyImage : NormalImage
-
-		return (
-			<div className="o-teaser__image-container js-teaser-image-container">
-				<Link
-					{...props}
-					url={displayUrl}
-					attrs={{
-						'data-trackable': 'image-link',
-						tabIndex: '-1',
-						'aria-hidden': 'true'
-					}}>
-					<div className="o-teaser__image-placeholder" style={{ paddingBottom: aspectRatio(image) }}>
-						<ImageComponent src={imageSrc} lazyLoad={imageLazyLoad} />
-					</div>
-				</Link>
-			</div>
-		)
 	}
+	const displayUrl = relativeUrl || url
+	const useImageService = !(image.url.startsWith('data:') || image.url.startsWith('blob:'))
+	const options = imageSize === 'XXL' && imageHighestQuality ? { quality: 'highest' } : {}
+	const imageSrc = useImageService ? imageService(image.url, ImageSizes[imageSize], options) : image.url
+	const ImageComponent = imageLazyLoad ? LazyImage : NormalImage
+
+	return (
+		<div className="o-teaser__image-container js-teaser-image-container">
+			<Link
+				{...props}
+				url={displayUrl}
+				attrs={{
+					'data-trackable': 'image-link',
+					tabIndex: '-1',
+					'aria-hidden': 'true'
+				}}>
+				<div className="o-teaser__image-placeholder" style={{ paddingBottom: aspectRatio(image) }}>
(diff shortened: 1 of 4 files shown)


components/x-teaser/src/Image.jsx
```diff
@@ -25,26 +25,30 @@ const LazyImage = ({ src, lazyLoad }) => {
 }
 
 export default ({ relativeUrl, url, image, imageSize, imageLazyLoad, imageHighestQuality, ...props }) => {
-	const displayUrl = relativeUrl || url
-	const useImageService = !(image.url.startsWith('data:') || image.url.startsWith('blob:'))
-	const options = imageSize === 'XXL' && imageHighestQuality ? { quality: 'highest' } : {}
-	const imageSrc = useImageService ? imageService(image.url, ImageSizes[imageSize], options) : image.url
-	const ImageComponent = imageLazyLoad ? LazyImage : NormalImage
+	if (image) {
```
fenglish, line 28: `if (!image) { return null }` on the top of this function would be clearer?

components/x-teaser/src/Image.jsx
```diff
@@ -25,26 +25,30 @@ const LazyImage = ({ src, lazyLoad }) => {
...
+		const displayUrl = relativeUrl || url
+		const useImageService = !(image.url.startsWith('data:') || image.url.startsWith('blob:'))
+		const options = imageSize === 'XXL' && imageHighestQuality ? { quality: 'highest' } : {}
+		const imageSrc = useImageService ? imageService(image.url, ImageSizes[imageSize], options) : image.url
+		const ImageComponent = imageLazyLoad ? LazyImage : NormalImage
 
-	return image ? (
```
umbobabo, line 34: Question: Wasn't the ternary operator applied here producing the same effect? I mean that if `image` is undefined it should return null.

fenglish: line 29 causes issues. we need to check `image` before the line
https://sentry.io/organizations/financial-times/issues/2005812713/?project=79503

components/x-teaser/src/Image.jsx
jkerr321: Oo yes! early return! I love that!! Great suggestion! Thank you 🙌

components/x-teaser/src/Image.jsx
jkerr321: Yep @fenglish has got it - this code runs before we check for the presence of `image` and errors with `Cannot read property 'startsWith' of undefined`:

```js
const useImageService = !(image.url.startsWith('data:') || image.url.startsWith('blob:'))
const options = imageSize === 'XXL' && imageHighestQuality ? { quality: 'highest' } : {}
const imageSrc = useImageService ? imageService(image.url, ImageSizes[imageSize], options) : image.url
```

umbobabo: `Cannot read property 'startsWith' of undefined` means that the `url` is undefined, not `image`. Image could be defined but not having a `url`. If Image is `undefined` how is that bit of code executed? The ternary operator should prevent that. Is it possible that `image` exists but doesn't have `url`?

umbobabo: If you run the existing function with no argument I would expect to return null, if you run it with an empty literal object like `{}` I would expect you to see the same error, cause `url` will not exists

umbobabo: Sorry, I was wrong about the ternary cause that is executed after, I misread in the rush. Pardon me 🙏

components/x-teaser/src/Image.jsx
```diff
@@ -25,26 +25,30 @@ const LazyImage = ({ src, lazyLoad }) => {
...
 
 export default ({ relativeUrl, url, image, imageSize, imageLazyLoad, imageHighestQuality, ...props }) => {
-	const displayUrl = relativeUrl || url
-	const useImageService = !(image.url.startsWith('data:') || image.url.startsWith('blob:'))
-	const options = imageSize === 'XXL' && imageHighestQuality ? { quality: 'highest' } : {}
-	const imageSrc = useImageService ? imageService(image.url, ImageSizes[imageSize], options) : image.url
-	const ImageComponent = imageLazyLoad ? LazyImage : NormalImage
+	if (!image || (image && !image.url)) {
+		return null
+	} else {
```
fenglish, line 30: I think we can remove this `else`

```js
if (!image || (image && !image.url)) {
  return null
}

const displayUrl = relativeUrl || url
...
```

umbobabo (approved):

components/x-teaser/src/Image.jsx
fenglish: I think less nesting would be better. In case when we need to add other conditions in the future.

fenglish (approved):
🙌

PR #577, opened 2021-01-22
CPP-419 x-interaction component registration

emortong:
[Ticket CPP-419](https://financialtimes.atlassian.net/browse/CPP-419)

From the ticket description: 
> x-interaction allows you to serialise component data for server-rendered components, and hydrate those components on the client side to enable actions with the client-side version of the component. registering components happens implicitly within x-interaction, and relies on heuristics to determine the component name. these must work the same on the client and server for the component to hydrate properly, which is unreliable, and is more likely to break in production environments than development.

This PR makes x-interaction export an explicit register function instead of implicitly registering. It also sets checks in both server (Serialiser) and client (Hydrate) code to ensure that all components are correctly registered. It also updates the README to reflect this change, (grammar/wording corrections are welcomed).

This has been tested in `next-article` using Live Blog Post content https://local.ft.com:5050/content/b47a8dd0-81dc-4136-807c-c071c75c8fe2

For it to work you'll need to turn the `liveBlogsV2` flag on.

The commits have been made based on the ticket's checklist order, so reviewing this PR by commit could be easier.

-----------------

## 

- Discuss features first
- Update the documentation
- Reference other issues
- Follow the code style

### DIFF
--- a/components/x-interaction/src/Hydrate.jsx
+++ b/components/x-interaction/src/Hydrate.jsx
@@ -52,6 +52,12 @@ export function hydrate() {
 
 		const Component = getComponent(component)
 
+		if (!Component) {
+			throw new Error(
+				`x-interaction hydrate was called using unregistered component: ${Component}. please verify you'are registering your component using x-interaction's registerComponent function before attempting to hydrate.`
+			)
+		}
+
 		while (wrapper.firstChild) {
 			wrapper.removeChild(wrapper.firstChild)
 		}
--- a/components/x-interaction/src/Hydrate.jsx
+++ b/components/x-interaction/src/Hydrate.jsx
@@ -1,5 +1,5 @@
 import { h, render, Component } from '@financial-times/x-engine'
-import { getComponent } from './concerns/register-component'
+import { getComponentByName } from './concerns/register-component'
 
 export class HydrationWrapper extends Component {
 	render() {
@@ -50,11 +50,11 @@ export function hydrate() {
 			)
 		}
 
-		const Component = getComponent(component)
+		const Component = getComponentByName(component)
 
 		if (!Component) {
 			throw new Error(
-				`x-interaction hydrate was called using unregistered component: ${Component}. please verify you'are registering your component using x-interaction's registerComponent function before attempting to hydrate.`
+				`x-interaction hydrate was called using unregistered component: ${component}. please verify you're registering your component using x-interaction's registerComponent function before attempting to hydrate.`
 			)
 		}
 
--- a/components/x-interaction/src/concerns/register-component.js
+++ b/components/x-interaction/src/concerns/register-component.js
@@ -1,9 +1,15 @@
 const registeredComponents = {}
 
-export function registerComponent(component) {
-	registeredComponents[component.wrappedDisplayName] = component
+export function registerComponent(component, name) {
+	const xInteractionName = Symbol('x-interaction-name')
+	component[xInteractionName] = name
+	registeredComponents[name] = component
 }
 
-export function getComponent(name) {
+export function getComponent(component) {
+	const xInteractionSymbol = Object.getOwnPropertySymbols(component).find(
+		(key) => key.toString() === `Symbol(x-interaction-name)`
+	)
+	const name = component[xInteractionSymbol]
 	return registeredComponents[name]
 }
--- a/components/x-interaction/src/concerns/register-component.js
+++ b/components/x-interaction/src/concerns/register-component.js
@@ -1,15 +1,25 @@
 const registeredComponents = {}
+const xInteractionName = Symbol('x-interaction-name')
 
-export function registerComponent(component, name) {
-	const xInteractionName = Symbol('x-interaction-name')
-	component[xInteractionName] = name
-	registeredComponents[name] = component
+export function registerComponent(Component, name) {
+	if (registeredComponents[name]) {
+		throw new Error(
+			`x-interaction a component has already been registered under that name, please use another name.`
+		)
+	}
+	Component[xInteractionName] = name
+	registeredComponents[name] = Component
 }
 
-export function getComponent(component) {
-	const xInteractionSymbol = Object.getOwnPropertySymbols(component).find(
-		(key) => key.toString() === `Symbol(x-interaction-name)`
-	)
-	const name = component[xInteractionSymbol]
+export function getComponent(Component) {
+	const name = Component[xInteractionName]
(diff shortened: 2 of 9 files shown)


components/x-interaction/src/Hydrate.jsx
```diff
@@ -52,6 +52,12 @@ export function hydrate() {
 
 		const Component = getComponent(component)
 
+		if (!Component) {
```
ker-an, line 55: **praise:** This is fab and would have come in handy when I was dealing with some hydration issues with `x-live-blog-wrapper`

components/x-interaction/src/Hydrate.jsx
```diff
@@ -52,6 +52,12 @@ export function hydrate() {
...
+			throw new Error(
+				`x-interaction hydrate was called using unregistered component: ${Component}. please verify you'are registering your component using x-interaction's registerComponent function before attempting to hydrate.`
```
ker-an, line 57: Suggestion to replace 

```
please verify you'are registering
```

with

```
please verify you're registering
```

components/x-interaction/src/concerns/register-component.js
```diff
@@ -1,9 +1,15 @@
 const registeredComponents = {}
 
-export function registerComponent(component) {
-	registeredComponents[component.wrappedDisplayName] = component
+export function registerComponent(component, name) {
+	const xInteractionName = Symbol('x-interaction-name')
```
ker-an, line 4: Maybe more of a Javascript question but I've never seen `Symbol` be used before 🙈 Reading up on it looks like it helps create a unique name here? I'm a tiny bit confused about what is going on here.

ker-an:
Just an overall suggestion (non-blocking), I had to visit the Jira ticket to understand why this was being done because I had totally forgotten about all the component registration issues I've encountered in the past. It might be nice to add a sentence or two in this pull request description on why this work is being done 😄

ker-an (approved):
Really great! Though I'd prob feel a bit more comfortable knowing Kara reviewed this 😂 I know a bit about `x-interaction` but not as much as her.

components/x-interaction/src/concerns/register-component.js
andygout: I have also not encountered `Symbol` before, and also had a question about its usage.

> Every symbol value returned from `Symbol()` is unique.

> `Symbol('foo') === Symbol('foo')  // false`

(Ref. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol)

If the benefit of using Symbol is that it is unique, is this perhaps undone by later identifying it by converting it into a string (``key.toString() === `Symbol(x-interaction-name)` ``), i.e. wouldn't all `xInteractionName` consts whose values are `Symbol('x-interaction-name')` all be the same?

Is there a risk from them all being the same or could we just use a string instead?

(I don't know much about `Symbol` so very likely I have missed something here.)

andygout (commented):
Looking good 👍, and likewise think it is worthwhile Kara having a look.

Is it possible to add tests for the new errors being thrown?

components/x-interaction/src/concerns/register-component.js
emortong: This was done as part of the step: 
> remove reliance on component `displayName`/`name` in `registerComponent`, instead have an explicit `name` argument, and store that name as an opaque property on the component itself

Having the name be a Symbol would make it semi-private. So if you did `Object.keys(component)` or `for (key in component)`, the `Symbol(x-interaction-name)` won't show up.

But your point does make me realize that if a component gets registered more than once, it will create multiple Symbols storing the name. I am not sure if the component getting registered multiple times is something that could happen, but it might be worth to add a check if the xInteractionName Symbol already exists. I am also not sure why the name needs to be a private property, @apaleslimghost do you remember why you proposed that on the ticket?

nickcolley (commented):
Not a full review, just commenting to say this is a great improvement and that this will be a breaking change.

components/x-interaction/src/Hydrate.jsx
apaleslimghost, line 57: this should log `component` (i.e. the name of the component), not `Component` which is going to be `undefined` here

components/x-interaction/src/concerns/register-component.js
apaleslimghost: i suggested it as a way of hiding an internal implementation detail from users of these components, so they wouldn't be able to use something that might break in the future. it's not necessary though.

i was intending for the `Symbol` to be defined outside of the scope of `registerComponent`, so it would always be the same symbol used for multiple component's keys. that would mean you don't have to do the `toString` later.

emortong: ahh!! that makes much more sense! I'll change it.. thanks Kara!

components/x-interaction/src/concerns/serialiser.js
```diff
@@ -1,6 +1,7 @@
 import { h, render } from '@financial-times/x-engine'
 import getComponentName from './get-component-name'
```
apaleslimghost, line 2: `getComponentName` is used below when adding serialisation data, but it still relies on using `.displayName`/`.name`. it should be moved into `register-component.js` and retrieve the name using the `x-interaction` name property

components/x-interaction/src/concerns/register-component.js
```diff
@@ -1,9 +1,15 @@
...
+	component[xInteractionName] = name
+	registeredComponents[name] = component
 }
 
-export function getComponent(name) {
+export function getComponent(component) {
```
apaleslimghost, line 9: this function is used by `hydrate` to get a component by its name, but here it seems to be taking a component as an argument. i think this needs to be split into two functions, `getComponent` which is the same as previously and `getComponentName` to replace `get-component-name.js` and return the component's name from the hidden property

apaleslimghost (commented):
a few comments, it's looking good but not quite ready yet. the documentation looks great!

components/x-interaction/src/concerns/register-component.js
apaleslimghost: oh, another thought, it would probably be a good idea to throw an error, or at least log a warning, if multiple components are registered with the same name. there's no reason to do this intentionally, and it might lead to the wrong component, or a different version of the same component, being used accidentally, which could break things

apaleslimghost (approved):
nice! this is fantastic

PR #579, opened 2021-01-27
ACC-749 x-gift-article improvements wrapped in a flag

AniaMakes:
This PR is https://github.com/Financial-Times/x-dash/pull/578 wrapped in a flag so that we can run an AB test. 

Screenshot 2021-01-21 at 11 43 54
Screenshot 2021-01-21 at 11 27 27

This PR is one of many steps to get the AB test rolled out:

- [x] create a flag in flags API, with the status off
- [ ] merge a PR in next-article that passes the flag status to x-gift-article - https://github.com/Financial-Times/next-article/pull/4117
- [ ] merge this PR, release it, make sure the release propagates where needed
- [ ] run a manual QA with flag on in toggler to make sure it all works as intended
- [ ] turn the AB test on

### DIFF
--- a/components/x-gift-article/src/CopyConfirmation.jsx
+++ b/components/x-gift-article/src/CopyConfirmation.jsx
@@ -9,12 +9,12 @@ const confirmationClassNames = [
 	styles['copy-confirmation']
 ].join(' ')
 
-export default ({ hideCopyConfirmation, isTestVariant }) => (
+export default ({ hideCopyConfirmation, isArticleSharingUxUpdates }) => (
 	<div className={confirmationClassNames} role="alert">
 		<div className={styles['o-message__container']}>
 			<div className={styles['o-message__content']}>
 				<p className={styles['o-message__content-main']}>
-					{isTestVariant ? (
+					{isArticleSharingUxUpdates ? (
 						<span>The link has been copied to your clipboard</span>
 					) : (
 						<span className={styles['o-message__content-highlight']}>
--- a/components/x-gift-article/src/CopyConfirmation.jsx
+++ b/components/x-gift-article/src/CopyConfirmation.jsx
@@ -9,14 +9,18 @@ const confirmationClassNames = [
 	styles['copy-confirmation']
 ].join(' ')
 
-export default ({ hideCopyConfirmation }) => (
+export default ({ hideCopyConfirmation, isTestVariant }) => (
 	<div className={confirmationClassNames} role="alert">
 		<div className={styles['o-message__container']}>
 			<div className={styles['o-message__content']}>
 				<p className={styles['o-message__content-main']}>
-					<span className={styles['o-message__content-highlight']}>
-						The link has been copied to your clipboard
-					</span>
+					{isTestVariant ? (
+						<span>The link has been copied to your clipboard</span>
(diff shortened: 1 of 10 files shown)


components/x-gift-article/src/CopyConfirmation.jsx
```diff
@@ -9,14 +9,18 @@ const confirmationClassNames = [
...
-export default ({ hideCopyConfirmation }) => (
+export default ({ hideCopyConfirmation, isTestVariant }) => (
 	<div className={confirmationClassNames} role="alert">
 		<div className={styles['o-message__container']}>
 			<div className={styles['o-message__content']}>
 				<p className={styles['o-message__content-main']}>
-					<span className={styles['o-message__content-highlight']}>
-						The link has been copied to your clipboard
-					</span>
+					{isTestVariant ? (
```
thurstontye, line 17: I would expect if this is an A/B test to return either 'control' or 'variant' so I think this would need to be `isTestVariant === 'variant'`

AniaMakes: Ah, well spotted - I haven't worked much with flags before, I assumed it would be "flag - true / false"

components/x-gift-article/src/Message.jsx
```diff
@@ -12,8 +12,41 @@ export default ({
 	monthlyAllowance,
 	nextRenewalDateText,
 	redemptionLimit,
-	invalidResponseFromApi
+	invalidResponseFromApi,
+	isArticleSharingUxUpdates
 }) => {
+	if (isArticleSharingUxUpdates) {
```
thurstontye, line 18: Not sure but I think currently this will always be true if the 'isArticleSharingUxUpdates ' flag exists and you are using a A/B flag. Unless I've missed something you should expect 'control' or 'variant' here or in the case of no flag 'false'. So the only condition this would be 'false' if the flag is not available.

thurstontye (changes requested):
The A/B condition naming needs to be more consitant and decide if you are casting the flag value to boolean or not

components/x-gift-article/src/Form.jsx
```diff
@@ -25,7 +26,10 @@ export default (props) => (
 		</form>
 
 		{props.showCopyConfirmation && (
-			<CopyConfirmation hideCopyConfirmation={props.actions.hideCopyConfirmation} />
+			<CopyConfirmation
+				hideCopyConfirmation={props.actions.hideCopyConfirmation}
+				isTestVariant={props.isArticleSharingUxUpdates}
```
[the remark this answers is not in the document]
AniaMakes: I don't have a good answer to this. I think sometimes I get confused trying to track a prop with the same name across the whole repo and when I was writing this maybe that's what I was thinking? 🤷  anyhow, I fixed it to be the same name as everywhere else.

thurstontye (approved):

thurstontye:
@AniaMakes not essential but it might be worth adding an explanation of the flag in x-gift-article readme https://github.com/Financial-Times/x-dash/blob/main/components/x-gift-article/readme.md

oluoluoxenfree (approved):

PR #582, opened 2021-02-08
CPP-464 register wrapper components and allow accessing wrapper components from original component

apaleslimghost:
if the original component is registered, when we try to hydrate it it'll render the markup but it doesn't have the actions. previously, if the wrapper component was registered, the serialiser would complain because it's trying to get the name of the component from the original component, which doesn't know it.

i'm not set up to test live blogs locally, so @GlynnPhillips i'd really appreciate if you could kick the tyres :)

### DIFF
--- a/components/x-interaction/src/concerns/wrap-component-name.js
+++ b/components/x-interaction/src/concerns/wrap-component-name.js
@@ -1,7 +1,5 @@
-import { getComponentName } from './register-component'
-
 function wrapComponentName(Component, Enhanced) {
-	const originalDisplayName = getComponentName(Component)
+	const originalDisplayName = Component.displayName || Component.name
 	Enhanced.displayName = `withActions(${originalDisplayName})`
 	Enhanced.wrappedDisplayName = originalDisplayName
 }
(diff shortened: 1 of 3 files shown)


emortong:
looks good, although I am curious about the failing test relating to the `displayName`

GlynnPhillips (approved):
I have quickly tested this on live blogs and it seems to fix our issue. Thanks Kara 👍 💯

GlynnPhillips:
Is the test thats failing sort of redundant now?

apaleslimghost:
the `displayName` is still useful for debugging, as it's what shows up in React Devtools, so i've made it use the actual original `displayName` not the registered name as it's not actually relied on by any code so it doesn't matter if it's different in production

components/x-interaction/src/concerns/wrap-component-name.js
```diff
@@ -1,7 +1,5 @@
-import { getComponentName } from './register-component'
-
 function wrapComponentName(Component, Enhanced) {
-	const originalDisplayName = getComponentName(Component)
+	const originalDisplayName = Component.displayName || Component.name
```
emortong, line 2: is `Component.name` available? since when the component is getting registered it does `Component[xInteractionName] = name`

apaleslimghost: `Component.name` here is the built-in Javascript function `name` so it should be there, `wrapComponentName` is only for debugging purposes, so it doesn't really matter :)

emortong (approved):

PR #587, opened 2021-03-01
CI-558, CI-559: Update Live Blog Post byline

nickcolley:
From Josh: "We need to adjust the styling of the byline so that it works better within the title lock-up."

| Before | After |
| - | - |
| Screenshot 2021-03-01 at 14 22 58 |  |


components/x-live-blog-post/storybook/index.jsx
```diff
@@ -9,7 +14,12 @@ export default {
 }
 
 export const ContentBody = (args) => {
-	return <LiveBlogPost {...args} />
+	return (
+		<div className="story-container">
+			{dependencies && <BuildService dependencies={dependencies} />}
```
GlynnPhillips, line 19: I didn't know about this way of including the build service in x-dash

GlynnPhillips (dismissed):
Code looks good and bigger more readable text makes me happy

noelenwenede (approved):

PR #588, opened 2021-03-01
CI-564: Remove overflow: auto from live blog post

nickcolley:
This was preventing us using `scroll-margin` to make the linking to the post via anchor better.

It was previously added to [allow for better calculations of the height of a post for scroll adjustments](https://github.com/Financial-Times/x-dash/commit/c90b2a459b5cc8a8a07a85faf1726ad358187444), but we no longer do this so we can remove it.


AliceArmstrong (dismissed):
✅

AliceArmstrong (approved):
🚢 🇮🇹

PR #589, opened 2021-03-02
Minimal fix: refactor targets properties alone

oliverturner:
### Context
Last week I found that a clean install of x-dash was failing to compile any of the components: running `make build` would error with the following message:

```
✖ [BABEL] <path-to-repo>/components/x-teaser/src/Teaser.jsx: @babel/helper-compilation-targets: '[object Object]' is not a valid browserslist query (While processing: "<path-to-repo>/packages/x-babel-config/node_modules/@babel/preset-env/lib/index.js")
```

Today @nickcolley reported having the same issue, so this PR is submitted as the quickest way to get compiling.

### PS
I investigated making a fuller update to the rollup dependencies (since the current versions have moved under the @rollup namespace) but had issues with emitting the `babelHelpers` in a form that consuming apps (such as `next-control-centre`) could pick up


nickcolley (approved):
This has made my day way easier, thank you Oliver!

PR #590, opened 2021-03-08
Remove unused listener logic

nickcolley:
We [moved this into the applications](https://github.com/Financial-Times/next-article/pull/4140) for now, if there is a good abstraction in the future we may bring some more of this logic back into x-dash.

This will be a breaking change.


components/x-live-blog-wrapper/readme.md
```diff
@@ -81,64 +76,43 @@ import { hydrate } from '@financial-times/x-interaction';
...
+    `[data-live-blog-wrapper-id="x-dash-element-id"]`
+);
+
+const post = {
+    id: '00000000-0000-0000-0000-000000000000',
+    ...
+};
+
+const action = 'insert-post';
+// wrapperElement must be the last argument.
```
nickcolley, line 99: If we made the `id` prop required we might be able to avoid asking people to pass the wrapper which is a bit weird...

GlynnPhillips (dismissed):

components/x-live-blog-wrapper/readme.md
GlynnPhillips: I don't feel too strongly about this - I do agree with you it might be nicer but I have also approved this for now as I think we could do this part later if we feel strongly about improving it?

AliceArmstrong (approved):
🙌

PR #591, opened 2021-03-18
NOPS-722 Fix blue headshots

jkerr321:
Currently on site headshots on the front page are not tinted blue, which they should be:
image

This fixes them so the tint is correctly applied:
image

### The fix:
The headshot file creates an options variable **string** containing the tint values which is later passed into an `imageService` function used to construct the URL to the Origami image service:
```js
const options = `tint=${headshotTint || DEFAULT_TINT}`
...
src={imageService(headshot, ImageSizes.Headshot, options)}
```

[The `imageService` function](https://github.com/Financial-Times/x-dash/blob/main/components/x-teaser/src/concerns/image-service.js) expects an **object** (which is spread into a base object to create the URL) and so incorrectly handles the incoming string and produces a garbage query param on the constructed URL:

```js
imageSrc.search = new URLSearchParams({ ...OPTIONS, ...options })
```

So the fix is to update the value passed to the `imageService` function to be an object.

URL before fix (see tint param): 
image

URL after fix: 
image

----
### How did this even work in the first place?!
Background to be discussed later with Kara:

- This bug was introduced in this PR: https://github.com/Financial-Times/next-front-page/commit/53dc5bd4bd2798c3558cc91a1de1aecf4a7c6b66 which bumped `x-teaser` in the front page from `1.6.4` to `3.3.0` on 22nd Jan
- In `v1.6.4` the code for the `imageService` function is set up to receive and process an `options` argument of type `string`, which makes sense on the surface:- somewhere between `v1.6.4` and `v3.3.0` that code was refactored to expect an object but updating the options type in the `headshot` function which calls `imageService` was forgotten:
```
 const BASE_URL = 'https://www.ft.com/__origami/service/image/v2/images/raw';
const OPTIONS = ['source=next', 'fit=scale-down', 'dpr=2'];

export default function imageService(url, width, options) {
	const encoded = encodeURIComponent(url);
	const href = `${BASE_URL}/${encoded}?${OPTIONS.join('&')}&width=${width}`;
	return options ? href + '&' + options : href;
}
```
- When `v3.3.0` was applied to the front-page this oversight manifested itself in not-blue headshots, and this PR is the fix for it

HOWEVER THIS IS NOT THE CASE!
- The code which refactored the `imageService` function to expect an object was actually written, released, and then _reverted_ on Dec 6th **2019** - https://github.com/Financial-Times/x-dash/pull/425 and the [commit history for the file](https://github.com/Financial-Times/x-dash/commits/main/components/x-teaser/src/Headshot.jsx) shows there have been no other changes since then bar one linting update:
image

- SO the object-expecting version of the `imageService` function should have been removed from the codebase entirely in 2019, and yet it was somehow re-applied at some point between `v1.6.4` and `v3.3.0` and yet there's no record of it in GitHub. Howwww?!!!

### DIFF
--- a/components/x-teaser/__tests__/__snapshots__/snapshots.test.js.snap
+++ b/components/x-teaser/__tests__/__snapshots__/snapshots.test.js.snap
@@ -216,7 +216,7 @@ exports[`x-teaser / snapshots renders a Hero teaser with opinion data 1`] = `
       aria-hidden="true"
       className="o-teaser__headshot"
       height={75}
-      src="https://www.ft.com/__origami/service/image/v2/images/raw/fthead-v1%3Agideon-rachman?0=t&1=i&2=n&3=t&4=%3D&5=0&6=5&7=4&8=5&9=9&10=3&11=%2C&12=d&13=6&14=d&15=5&16=d&17=3&source=next&fit=scale-down&dpr=2&width=75"
+      src="https://www.ft.com/__origami/service/image/v2/images/raw/fthead-v1%3Agideon-rachman?source=next&fit=scale-down&dpr=2&tint=054593%2Cd6d5d3&width=75"
       width={75}
     />
   </div>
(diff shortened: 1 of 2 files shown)


taraojo (approved):
Nice sleuthing 🎉  Interestingly we should have caught this in the snapshot test diff, but I guess we didn't notice that particular change (which happens a lot with snapshots). Would be good to have some visual regression tests for these kind of components.

PR #596, opened 2021-04-08
Send setConsentCookie only when legislationId is 'gdpr'

oliverturner:
It has transpired that the Privacy Manager component should send the `setConsentCookie` flag based on  legislation: CCPA users opting in/out of the sale of their data should not have `FTConsent_GDPR` set to true because they should still be shown the cookie banner.

This PR [encapsulates that requirement](https://github.com/Financial-Times/x-dash/pull/596/files#diff-5c14b4096cd7b936e70498295744fa9fa2a6d91df6a92a676892bd6fefd26614R70) so that consuming apps do not need to make any changes

Additionally 
- Tests have been refactored to simplify the mounting a configured instance of the component
- Type definitions have been updated to make reference easier under an `XPrivacyManager` namespace

### DIFF
--- a/components/x-privacy-manager/jsconfig.json
+++ b/components/x-privacy-manager/jsconfig.json
@@ -1,7 +1,6 @@
 {
   "compilerOptions": {
     "outDir": "dist/types",
-    "allowJs": true,
     "target": "es2015",
     "module": "commonjs",
     "strict": true,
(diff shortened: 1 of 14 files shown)


alevito (dismissed):
Good job!

andrewgeorgiou1981 (approved):

andrewgeorgiou1981 (approved):
All LGTM - nice updates to the test-helpers.

PR #599, opened 2021-05-13
handle multiple customSlot components for x-teaser-timeline

gitGremlin:
This update is a minor patch to enable multiple customSlots to be added to the `x-teaser-timeline` component.

Ideally I'd liked to have make this a new import, but I didn't want to make the change a breaking one. `x-teaser-timeline` can still accept the old style of customSlot (a String and a Number), but now also allows for multiple components.

### DIFF
--- a/components/x-teaser-timeline/__tests__/lib/transform.test.js
+++ b/components/x-teaser-timeline/__tests__/lib/transform.test.js
@@ -322,5 +322,32 @@ describe('buildModel', () => {
 				groupedItems[2]
 			])
 		})
+		test('returns correct model for multiple custom slots off end of all groups', () => {
+			const result = buildModel({
+				items,
+				timezoneOffset: 0,
+				localTodayDate: '2020-01-14',
+				customSlotContent: [{ foo: 1 }, { bar: 2 }],
+				customSlotPosition: [0, 10]
+			})
+			expect(result).toEqual([
+				{
+					...groupedItems[0],
+					items: [
+						{ foo: 1 },
+						groupedItems[0].items[0],
+						groupedItems[0].items[1],
+						groupedItems[0].items[2],
+						{ bar: 2 },
+						groupedItems[0].items[3]
+					]
+				},
+				groupedItems[1],
+				{
+					...groupedItems[2],
+					items: [...groupedItems[2].items, { bar: 2 }]
+				}
+			])
+		})
 	})
 })
--- a/components/x-teaser-timeline/__tests__/lib/transform.test.js
+++ b/components/x-teaser-timeline/__tests__/lib/transform.test.js
@@ -338,7 +338,6 @@ describe('buildModel', () => {
 						groupedItems[0].items[0],
 						groupedItems[0].items[1],
 						groupedItems[0].items[2],
-						{ bar: 2 },
 						groupedItems[0].items[3]
 					]
 				},
--- a/components/x-teaser-timeline/__tests__/lib/transform.test.js
+++ b/components/x-teaser-timeline/__tests__/lib/transform.test.js
@@ -322,6 +322,30 @@ describe('buildModel', () => {
 				groupedItems[2]
 			])
 		})
+		test('returns correct model for a non-zero custom slot', () => {
+			const result = buildModel({
+				items,
+				timezoneOffset: 0,
+				localTodayDate: '2020-01-14',
+				customSlotContent: [{ foo: 1 }, { bar: 2 }],
+				customSlotPosition: [2, 3]
+			})
+			expect(result).toEqual([
+				{
+					...groupedItems[0],
+					items: [
+						groupedItems[0].items[0],
+						groupedItems[0].items[1],
+						{ foo: 1 },
+						groupedItems[0].items[2],
+						{ bar: 2 },
+						groupedItems[0].items[3]
+					]
+				},
+				groupedItems[1],
+				groupedItems[2]
+			])
+		})
 		test('returns correct model for multiple custom slots off end of all groups', () => {
 			const result = buildModel({
 				items,
--- a/components/x-teaser-timeline/__tests__/lib/transform.test.js
+++ b/components/x-teaser-timeline/__tests__/lib/transform.test.js
@@ -298,5 +298,29 @@ describe('buildModel', () => {
 				groupedItems[2]
 			])
 		})
+		test('returns correct model for multiple custom slots', () => {
+			const result = buildModel({
+				items,
+				timezoneOffset: 0,
+				localTodayDate: '2020-01-14',
+				customSlotContent: [{ foo: 1 }, { bar: 2 }],
+				customSlotPosition: [0, 3]
+			})
+			expect(result).toEqual([
+				{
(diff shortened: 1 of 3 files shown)


components/x-teaser-timeline/src/lib/transform.js
```diff
@@ -165,15 +165,19 @@ export const buildModel = ({
...
-
 	if (itemGroups.length > 0 && customSlotContent) {
-		const insertPosition = Math.min(customSlotPosition, items.length)
-		const insert = getGroupAndIndex(itemGroups, insertPosition)
-		const copyOfItems = [...itemGroups[insert.group].items]
+		customSlotContent = Array.isArray(customSlotContent) ? customSlotContent : [customSlotContent]
+		customSlotPosition = Array.isArray(customSlotPosition) ? customSlotPosition : [customSlotPosition]
+
+		customSlotContent.forEach((item, index) => {
+			const insertPosition = Math.min(customSlotPosition[index], items.length)
```
alevito, line 173: Considering that the `items` array only includes the news articles, I wonder if the total items should be `items.length + customSlotPosition.length`?

e.g.,
```suggestion
			const insertPosition = Math.min(customSlotPosition[index], customSlotPosition.length + items.length)
```

components/x-teaser-timeline/src/lib/transform.js
```diff
@@ -165,15 +165,19 @@ export const buildModel = ({
 		latestItemsTime,
 		latestItemsAgeHours
 	})
-
 	if (itemGroups.length > 0 && customSlotContent) {
-		const insertPosition = Math.min(customSlotPosition, items.length)
-		const insert = getGroupAndIndex(itemGroups, insertPosition)
-		const copyOfItems = [...itemGroups[insert.group].items]
+		customSlotContent = Array.isArray(customSlotContent) ? customSlotContent : [customSlotContent]
```
alevito, line 169: Maybe the variable names customSlotContent and customSlotPosition can be pluralised now that they are respectively an array of slots, and an array of position.
I would also consider improving the prototype of the `X-Teaser-Timeline` component to take a `customSlots` prop like
```
customSlots: Record<position, content>
```
which would still allow us to loop over the keys (slot positions) or over the values (slot contents).

alevito, line 169: This loop could be rewritten as
```suggestion
		for (const slotPosition of customSlotPositions) { ... }
```
which would give direct access to the values of the array of slot positions.

components/x-teaser-timeline/src/lib/transform.js
gitGremlin: You're right, this was atrocious of me to miss.

gitGremlin: It would have to be `items.length + customSlotPosition.length - 1`, otherwise it would overflow the array.

components/x-teaser-timeline/src/lib/transform.js
gitGremlin: I agree with the above, though this would be better for a future refactor. I don't want to make this a breaking change and doing so will but in order to make this change a non-breaking change, I need to think about this some more.

components/x-teaser-timeline/src/lib/transform.js
```diff
@@ -165,15 +165,19 @@ export const buildModel = ({
...
+		customSlotPosition = Array.isArray(customSlotPosition) ? customSlotPosition : [customSlotPosition]
+
+		customSlotContent.forEach((item, index) => {
```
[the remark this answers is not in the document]
gitGremlin: I changed over to something like this. :)

oliverturner (approved):
This looks clear and well-tested to me: nice job

components/x-teaser-timeline/src/lib/transform.js
```diff
@@ -165,15 +165,21 @@ export const buildModel = ({
...
+			const insertPosition = Math.min(customSlotPositionArray[index], items.length + index)
+			const insert = getGroupAndIndex(itemGroups, insertPosition)
+			const copyOfItems = [...itemGroups[insert.group].items]
 
-		copyOfItems.splice(insert.index, 0, customSlotContent)
+			copyOfItems.splice(insert.index, 0, slotContent)
 
-		itemGroups[insert.group].items = copyOfItems
+			itemGroups[insert.group].items = copyOfItems
+		}
```
oliverturner, line 182: Suggestion [non-blocking]: extracting this into a named function might make the purpose of this loop easier to grok at a glance

gitGremlin: I think I've done this.

components/x-teaser-timeline/__tests__/lib/transform.test.js
```diff
@@ -298,5 +298,55 @@ describe('buildModel', () => {
...
+			})
+			expect(result).toEqual([
+				{
+					...groupedItems[0],
+					items: [
+						{ foo: 1 },
+						groupedItems[0].items[0],
+						groupedItems[0].items[1],
+						groupedItems[0].items[2],
+						{ bar: 2 },
```
Lily2point0, line 317: That's index 4. If I read the input properly it would expect it at index 3?

Lily2point0: Reading the code, I understand that it leaves 3 items between the first insert position and the next custom slot. I think the "Position" variable name is a bit confusing in that case.

components/x-teaser-timeline/__tests__/lib/transform.test.js
```diff
@@ -298,5 +298,55 @@ describe('buildModel', () => {
...
+				groupedItems[2]
+			])
+		})
+		test('returns correct model for multiple custom slots off end of all groups', () => {
+			const result = buildModel({
+				items,
+				timezoneOffset: 0,
+				localTodayDate: '2020-01-14',
+				customSlotContent: [{ foo: 1 }, { bar: 2 }],
+				customSlotPosition: [0, 10]
```
Lily2point0, line 331: I think it would be worth having a test case where the first insert position isn't 0, to validate the offset

gitGremlin: I've added another test [2,3] to show the effects at a non-0 position.

components/x-teaser-timeline/__tests__/lib/transform.test.js
gitGremlin: The confusing part is that the insertion is performed _simultaneously_. It's not leaving a 3 offset, but it's inserting at positions 0 and 3 before the change.
ie. it inserts something before `groupedItems[0].items[0]` and before `groupedItems[0].items[3]`
By adding another test case (positions 2 and 3), I'm hoping that makes it clearer.

I think this is preferred because the `customSlot` addition is for non-news items being inserted, so we wouldn't want to have to factor in that added slot in the decision of adding another slot.

alevito (approved):

Lily2point0 (approved):

PR #603, opened 2021-06-10
At 4670 always display timestamps on x-teasers

rashedul-hassan-app:
I personally love `x-teaser` and its a great opportunity to work on it being part of the apps team.

This PR modifies the current behaviour of `x-teaser's` sub-component `Status` - to add an extra feature.

We are trying to work on a new feature in Apps team where we require the `x-teaser` to display the timestamp signature to be available all the time. As we all know, `x-teaser` has a default configuration for the `Status` component, where if an article was published less than 4 hours ago, the time-stamp is being displayed. To make the timestamp appear always, this PR has been created.

**Approach**:
I have applied a very simplistic logic where, if a certain property - `alwaysDisplayTimestamp` is injected within a teaser data, we do a logical check on what to display in the timestamp. If its on the same day, we display the timestamp otherwise we display the full date.

To avoid code-duplication, I have refactored the code and introduced a prop within the Relative Time component, which conditionally adds `data-o-date-format` to the DOM. The default behaviour is kept as it is, and this element is removed when the `alwaysDisplayTimestamp` is required.

**Testing**:
Since the work revolves around the `Status` component of `x-teaser`, the full bit of the test has been done via running the storybook.

I could not find a place where the Status component is documented in detail, so I could add that information. As such, I have left comments in the code for clarity.


components/x-teaser/src/AlwaysShowTimestamp.jsx
```diff
@@ -0,0 +1,21 @@
+import { h } from '@financial-times/x-engine'
+import TimeStamp from './TimeStamp'
+import RelativeTime from './RelativeTime'
+import { differenceInCalendarDays } from 'date-fns'
+import { getDateOnly } from '../../x-teaser-timeline/src/lib/date'
```
apaleslimghost, line 5: **issue**: to use `x-teaser-timeline` you'd need to add it as a dependency of `x-teaser` and refer to it by package name (since they're separate packages, you can't rely on relative paths). however, it doesn't feel right to have `x-teaser` depend on `x-teaser-timeline`; could this function be extracted into a common library, or is there maybe something we could use from `date-fns`?

components/x-teaser/src/AlwaysShowTimestamp.jsx
```diff
@@ -0,0 +1,21 @@
...
+ */
+export default (props) => {
+	const localTodayDate = getDateOnly(new Date().toISOString())
+	const dateToCompare = getDateOnly(new Date(props.publishedDate).toISOString())
+
+	if (differenceInCalendarDays(localTodayDate, dateToCompare) >= 1) {
+		return <TimeStamp {...props} />
+	} else {
+		return <RelativeTime {...props} override4hLimit={true} />
+	}
```
apaleslimghost, line 20: **thought**: it might be nicer to move this logic into `o-date` as a new [format](https://github.com/Financial-Times/o-date#format-options)? it's fine here though

components/x-teaser/src/Status.jsx
```diff
@@ -2,14 +2,17 @@ import { h } from '@financial-times/x-engine'
...
 export default (props) => {
 	if (props.status) {
 		return <LiveBlogStatus {...props} />
 	}
 
 	if (props.publishedDate) {
-		if (props.useRelativeTime) {
+		if (props.alwaysDisplayTimestamp) {
+			return <AlwaysShowTimestamp {...props} />
+		} else if (props.useRelativeTime) {
```
apaleslimghost, line 15: **suggestion**: it would be good to add this variant to Storybook & the snapshot tests somewhere, i can't remember how to do that though 😬

apaleslimghost (commented):
thanks @rashedul-hassan-app! a few comments, one that needs addressing before merge.

(note, my review comments use the [Conventional Comments](https://conventionalcomments.org/) format)

components/x-teaser/src/Status.jsx
```diff
@@ -2,14 +2,17 @@ import { h } from '@financial-times/x-engine'
...
+import AlwaysShowTimestamp from './AlwaysShowTimestamp'
 
 export default (props) => {
 	if (props.status) {
 		return <LiveBlogStatus {...props} />
 	}
 
 	if (props.publishedDate) {
-		if (props.useRelativeTime) {
+		if (props.alwaysDisplayTimestamp) {
```
rashedul-hassan-app, line 13: **Thoughts**: Do you have any thoughts on the choice of name? I didn't want to make it too big, but then again difficult to portray what it is doing - 'Relative time for the same day, non-relative for any other'.

Other thought I have is to, plug in the name of the work that it is tied with e.g. how LiveBlogStatus is done. In this case it would be something along the line of 'NewsfeedTimestamp'

rashedul-hassan-app:
I have pushed a commit that does not use the `x-teaser-timeline` library anymore.

Also, the apps team and I are thinking of a better variable name, instead of `alwaysShowTimestamp` - which I will probably update Monday.

components/x-teaser/src/RelativeTime.jsx
```diff
@@ -15,17 +15,17 @@ const displayTime = (date) => {
 	return `${hours} ${suffix}`
 }
 
-export default ({ publishedDate, firstPublishedDate }) => {
+export default ({ publishedDate, firstPublishedDate, override4hLimit = false }) => {
 	const relativeDate = getRelativeDate(publishedDate)
 	const status = getStatus(publishedDate, firstPublishedDate)
 
-	return isRecent(relativeDate) ? (
+	return override4hLimit === true || isRecent(relativeDate) ? (
```
erayb, line 22: `isRecent` function here abstracts away what recent is. It is 4 hours now, but can change in the future. If that happens your override variable name would need to be updated too.

`isRecent` is using currently a constant but it might make sense to pass an optional alternative.

components/x-teaser/src/AlwaysShowTimestamp.jsx
erayb: Ooh interesting. This definitely looks like what you are trying to do @rashedul-hassan-app.

rashedul-hassan-app:
Hi @apaleslimghost, thank you for all the comments and feedbacks. 

I have tested this PR locally running with Ft-App, however I have not tested it with FT.com. I realise this is a 'MUST' specification as I read in the template. If so, could you please throw some links for me to get it tested there? 

Otherwise, I think this PR is ready! :)

apaleslimghost:
[next-front-page](https://github.com/financial-times/next-front-page) would be a good app to try this out in, it should be straight forward to set up. you'll need to [install next-router](https://github.com/Financial-Times/next-router#setup) as well if you haven't already

rashedul-hassan-app:
Thank you @apaleslimghost, got it tested on next-front-page, behaving exactly as envisaged. PR is ready for your final review, if you have anything to add.

Plan to merge not on Friday :)

apaleslimghost (commented):
looks great, nice work with `o-date`!

(oops, sorry, wrong button, meant to approve)

apaleslimghost (approved):

PR #604, opened 2021-06-24
feat: Add support for Preact.Fragment in x-engine

maxcbc:
We currently support Fragments in react and hyperons, but not preact ([which did not support Fragments until the latest major version, 10.x](https://preactjs.com/guide/v10/whats-new/#fragments)).

This PR adds support via x-engine for `preact.Fragment`.

I'm unsure how this change should affect versioning. So would very much appreciate input on that.

### DIFF
--- a/packages/x-engine/src/concerns/presets.js
+++ b/packages/x-engine/src/concerns/presets.js
@@ -11,6 +11,7 @@ module.exports = {
 		runtime: 'preact',
 		factory: 'h',
 		component: 'Component',
+		fragment: 'Fragment',
 		render: 'render'
 	},
 	hyperons: {


apaleslimghost (approved):
looks good! i think this can go out in a minor version, it shouldn't break anything?

maxcbc:
@apaleslimghost I don't think it should.
If someone uses this with preact 8.x it should still work. I think this will only affect older apps anyway as 10.x has been out since early 2020.

apaleslimghost:
next-article is notably still on Preact 8. which component will be using fragments? is it used in next-article?

maxcbc:
@apaleslimghost next-discovery-pages. next-article shouldn't be able to use Fragments if its using preact v8.

maxcbc:
@apaleslimghost please can you merge this for me? I don't have merge perms

apaleslimghost:
dah, sorry, will  do

apaleslimghost:
released as 6.1!

PR #605, opened 2021-07-01
Fix gift article storybook to call activate action

ivomurrell:
Fixes an issue where the gift articles stories in storybook weren't showing how many credits were left.

As per the gift article docs, the `activate` action needs to be called to run the API call (which is mocked by `node-fetch` in this case.) Not exactly sure where the best place to do this would be so I just put it inside its own `actionRef` callback. Maybe this makes sense as we aren't showing the `GiftArticle` component conditionally, instead showing it as soon as it's mounted.

For reference:

- [Description of `actionRef`](https://github.com/Financial-Times/x-dash/tree/09fd985bd97ce907cb297ad586ac9989cee9c296/components/x-gift-article#usage)
- [Description of the `ref` pattern it's based on](https://reactjs.org/docs/refs-and-the-dom.html)

Screenshot of 'With gift credits' story before and after:

**Before**
Screenshot 2021-07-06 at 15 51 08
**After**
Story after fix

### DIFF
--- a/components/x-gift-article/storybook/free-article.js
+++ b/components/x-gift-article/storybook/free-article.js
@@ -19,12 +19,10 @@ exports.fetchMock = (fetchMock) => {
 	fetchMock
 		.restore()
 		.get('/article/gift-credits', {
-			credits: {
-				allowance: 20,
-				consumedCredits: 5,
-				remainingCredits: 15,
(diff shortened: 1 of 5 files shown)


AniaMakes (approved):
It looks like it should do the trick! Thank you!

PR #606, opened 2021-07-01
AT-4761 Update x-teaser status spacing

CLTPayne:
#### Context:
* Apps team UX noticed that the App is out of sync with FT.com and has less spacing between the Teaser status dot and the status text (FT.com currently has `5px`). 
* Investigation found that FT.com is using some older versions of `o-teaser` and `o-labels` - `next-front-page` uses `o-teaser "4.1.0-beta.1", "o-labels": "^5.0.0", "@financial-times/x-teaser": "^4.1.3"`, `ft-app` uses `o-teaser": "^5.2.2", o-labels": "^5.1.5", x-teaser": "5.2.0"`
* `o-labels` now (as of `v5.1.5`) only adds `margin-right: 2px` between these two elements. Change introduced [here](https://github.com/Financial-Times/o-labels/commit/cbdfb2a7d78427b691a439704613f3c2a6c34014#diff-369e718fbbdf6907bf9e45dfdf3234cce054c6e399981f0bca5cfa5a2926d1a8).
* Design spec for the status shows `4px` space between the elements:

#### Approach:
* Raised discussion with the Origami team to see if preference was to update `o-labels` to `margin-right: 4px` or update `x-teaser`. The Origami team were keen to leave `o-labels` as is (due to unknown number of consumers and wider impact). Full conversation here - https://financialtimes.slack.com/archives/C02FU5ARJ/p1624447956081800
* Adding a leading literal space creates the required visual. (`o-labels` demo also includes a leading space https://www.ft.com/__origami/service/build/v2/demos/o-labels@5.2.0/indicators?brand=master). 

#### Testing:
1. Updated the `x-teaser` storybook dependencies to use the newer `o-teaser` and `o-labels` as per this change of styles.
2. Tested locally in the FT App:
Updated Status 'New' in FT App - July 2021
Live Blog Status in FT App - July 2021
X-teaser Timeline Status in FT App - July 2021

3. Tested locally on FT.com front page (also had to update `o-teaser` and `o-labels` locally for for this test):
Update Status on FT com Next Front Page - July 2021

#### Proposed Next Steps:
- If approved, release changes as a "patch" change to `x-teaser`.
- I will contact the `next-front-page` maintainers to let them know about this change.
- Update `ft-app` to consume a new "patch" version of `x-teaser` containing this change.

### DIFF
--- a/components/x-teaser/src/LiveBlogStatus.jsx
+++ b/components/x-teaser/src/LiveBlogStatus.jsx
@@ -15,6 +15,6 @@ const LiveBlogModifiers = {
 export default ({ status }) =>
 	status && status !== 'closed' ? (
 		<div className={`o-teaser__timestamp o-teaser__timestamp--${LiveBlogModifiers[status]}`}>
-			<span className="o-teaser__timestamp-prefix">{LiveBlogLabels[status]}</span>
+			<span className="o-teaser__timestamp-prefix">{` ${LiveBlogLabels[status]} `}</span>
 		</div>
 	) : null
--- a/components/x-teaser/src/RelativeTime.jsx
+++ b/components/x-teaser/src/RelativeTime.jsx
@@ -21,7 +21,7 @@ export default ({ publishedDate, firstPublishedDate, showAlways = false }) => {
 
 	return showAlways === true || isRecent(relativeDate) ? (
 		<div className={`o-teaser__timestamp o-teaser__timestamp--${status}`}>
-			{status ? <span className="o-teaser__timestamp-prefix">{`${status} `} </span> : null}
+			{status ? <span className="o-teaser__timestamp-prefix">{` ${status} `} </span> : null}
 			<time
 				className="o-teaser__timestamp-date o-date"
 				data-o-component="o-date"
--- a/components/x-teaser/storybook/index.jsx
+++ b/components/x-teaser/storybook/index.jsx
@@ -4,9 +4,9 @@ import BuildService from '../../../.storybook/build-service'
 
 const dependencies = {
 	'o-date': '^4.2.0',
-	'o-labels': '^5.0.0',
+	'o-labels': '^5.2.0',
 	'o-normalise': '^2.0.0',
-	'o-teaser': '^4.0.0',
+	'o-teaser': '^5.2.3',
 	'o-typography': '^6.0.0',
 	'o-video': '^6.0.0'
 }


components/x-teaser/src/LiveBlogStatus.jsx
```diff
@@ -15,6 +15,6 @@ const LiveBlogModifiers = {
 export default ({ status }) =>
 	status && status !== 'closed' ? (
 		<div className={`o-teaser__timestamp o-teaser__timestamp--${LiveBlogModifiers[status]}`}>
-			<span className="o-teaser__timestamp-prefix">{LiveBlogLabels[status]}</span>
+			<span className="o-teaser__timestamp-prefix">{` ${LiveBlogLabels[status]} `}</span>
```
CLTPayne, line 18: Have added the trailing space too for consistency with the `<RelativeTime />` component.

umbobabo (approved):

kavanagh:
FT.com is deliberately using older Teaser styles to avoid getting the typography style changes.

The change of headline weight effects layout. The decision effectively boils down to not wanting to do any visible changes to the Front page while we're building a new one.

What's the solution to this I wonder?

umbobabo:
> The change of headline weight effects layout. The decision effectively boils down to not wanting to do any visible changes to the Front page while we're building a new one.

When the new Home page will be ready we will need anyway to do a general review and the same will apply when the new mobile App will be ready.
I do not see any other option considering we have diverged from the usage of x-dash components on the new Home page. At some point, we will need to converge and align with the rest of ft.com in some way.

CLTPayne:
Thanks both! That clarifies my question for `next-front-page`! :)  This will get the App closer to the current FT.com front page / those older styles so I'll get the App updated.

kavanagh:
I have some questions:

* If this x-teaser release is not a major, what's to stop someone mistakenly updating next-front-page to the latest x-teaser version?
* What if a future fix in x-teaser _needs_ to be deployed to next-front-page?
* What about next-stream-page, next-myft-page, next-article? Should they have the new teaser styles?

CLTPayne:
It seems to me that the FT.com apps are all at least one major version behind (x-teaser is currently on 5.2.0):

- Article Page "1.6.4" - https://github.com/Financial-Times/next-article/blob/master/package.json#L43
- MyFT Page "1.0.2" - https://github.com/Financial-Times/next-myft-page/blob/master/package.json#L39
- Stream Page "^4.1.3" - https://github.com/Financial-Times/next-stream-page/blob/master/package.json#L32
- Front Page "^4.1.3" - https://github.com/Financial-Times/next-front-page/blob/master/package.json#L25

And old `o-labels` and `o-teaser` styles too.

My thinking was that the original change to `o-labels` that introduced was only a minor release but makes sense re making the spacing change in x-teaser a `major` release - I will specify the required `o-labels` version to maintain the design spec in the release info. Unless there is a preferred route to how the App and x-teaser should be updated.

umbobabo:
> If this x-teaser release is not a major, what's to stop someone mistakenly updating next-front-page to the latest x-teaser version?
> What if a future fix in x-teaser needs to be deployed to next-front-page?

@kavanagh These points concern me the most cause we are already currently behind the latest and those are Plantinum services so we could struggle to release a hotfix if necessary.
We will evaluate the amount of work to do the upgrade in the repos we own and then decide what to do. @jkerr321 this discussion could interest your team too due to the involvement of the Article Page.

PR #610, opened 2021-09-03
CON-1123 - fix invalid use of aria role on x-teaser

gyss:
## Description

`aria-label` on a span element is not readable by screen readers. This PR adds a new hidden element to help screen readers to read the Premium label.

## Meta

- [Jira](https://financialtimes.atlassian.net/browse/CON-1123)

### DIFF
--- a/components/x-teaser/src/Title.jsx
+++ b/components/x-teaser/src/Title.jsx
@@ -28,9 +28,8 @@ export default ({ title, altTitle, headlineTesting, relativeUrl, url, indicators
 			{indicators && indicators.accessLevel === 'premium' ? (
 				<span>
 					{' '}
-					<span className={premiumClass} aria-label="Premium content">
-						Premium
-					</span>
+					<span className={premiumClass}>Premium</span>
+					<span className="o-normalise-visually-hidden">&nbsp;content</span>
 				</span>
 			) : null}
 		</div>


components/x-teaser/src/Title.jsx
```diff
@@ -28,9 +28,8 @@ export default ({ title, altTitle, headlineTesting, relativeUrl, url, indicators
 			{indicators && indicators.accessLevel === 'premium' ? (
 				<span>
 					{' '}
-					<span className={premiumClass} aria-label="Premium content">
-						Premium
-					</span>
+					<span className={premiumClass}>Premium</span>
+					<span className="o-normalise-visually-hidden">Premium content</span>
```
apaleslimghost, line 32: **question**: is `o-normalise` definitely going to be available on every page that x-teaser is used?

components/x-teaser/src/Title.jsx
```diff
@@ -28,9 +28,8 @@ export default ({ title, altTitle, headlineTesting, relativeUrl, url, indicators
 			{indicators && indicators.accessLevel === 'premium' ? (
 				<span>
 					{' '}
-					<span className={premiumClass} aria-label="Premium content">
-						Premium
-					</span>
+					<span className={premiumClass}>Premium</span>
```
apaleslimghost, line 31: **issue**: this element should be hidden from screenreaders, otherwise they'll read "Premium premium content"

components/x-teaser/src/Title.jsx
umbobabo, line 32: I belive they suggest to do this in the ticket. A screen reader should read it as a single phrase.
```suggestion
					<span className={premiumClass}>Premium</span>
					<span className="o-normalise-visually-hidden">&nbsp;content</span>
```

umbobabo: Good point. Only if they use o-teaser, that has it as [Peer dependency](https://github.com/Financial-Times/o-teaser/blob/e6e1ae48fcdc1895c06b13f5f67763313d9815fd/package.json#L35). I'm not aware of anything using x-teaser without o-teaser but I believe it's hard to give a precise answer without an easy way to know all the consumers of x-teaser.

apaleslimghost: +1 to this solution!

gyss: thanks @umbobabo @apaleslimghost . I've just tried it with Voice Over and reads correctly "Premium content"

apaleslimghost: well, x-teaser already assumes o-teaser's styles are available, so that's fine

gyss: I gathered a list of the repositories that uses x-teaser:

```
next-myft-page
next-stream-page
next-front-page
cp-chared-api-proposal
next-article
next-video-page
advertising
next-tour-page
n-native-ads
next-epaper
next-search-page
ft-app
next-measurement
```

Taking for example [next-tour-page](https://github.com/Financial-Times/next-tour-page/blob/main/bower.json) you can see that it requires `x-teaser` but doesn't include o-teaser at all (or even o-normalise). So far I guess there are 2 options:
- include o-normalise as a dependency of `x-teaser`. This would require all repositories that use x-teaser to import the new scss file generated with `o-normalise` bundled in it, right?
- review all repositories that use `x-teaser` and update them with an import of `o-teaser` (and update the docs specifying that `o-teaser` is required.

which one do you think it'd be the correct path?

apaleslimghost: nice sleuthing work 😀 in the case of next-tour-page, it looks like a mistake that it depends on x-teaser? i don't see it being used anywhere in the code. i don't think bundling o-normalise would be viable, and x-teaser already has an implicit dependency on o-teaser, so i think we're fine

umbobabo (approved):

components/x-teaser/src/Title.jsx
umbobabo: Although not explicitly stated as dependency, the components assume to work in conjunction of o-teaser, [see example here](https://github.com/Financial-Times/x-dash/blob/78abaac05295c702be47928b2883dbd4ebba8812/components/x-teaser/src/Container.jsx#L34) so I approved it but last call on @apaleslimghost

apaleslimghost (approved):

PR #617, opened 2021-09-24
CP-465 Added e2e test for hydration and serialisation

serena97:
runs a real express app in the e2e test, which involves creating a SSR button that is hydrated (ie. event listener added to the SSR template). puppeteer clicks the button to show that the component is properly hydrated. this gif illustrates it:
https://user-images.githubusercontent.com/20088576/134715395-be9a46af-90f8-4e3d-a2ea-259c3888209d.mov

### DIFF
--- a/components/x-interaction/readme.md
+++ b/components/x-interaction/readme.md
@@ -179,7 +179,7 @@ When rendered on the server side, components output an extra wrapper element, wi
 
 `x-interaction` exports a function `hydrate`. This should be called on the client side. It inspects the global serialisation data on the page, uses the identifiers to find the wrapper elements, and calls `render` from your chosen `x-engine` client-side runtime to render component instances into the wrappers.
 
-Before calling `hydrate`, you must first `import` any `x-interaction` components that will be rendered on the page. The components register themselves with the `x-interaction` runtime when imported; you don't need to do anything with the imported component. This will also ensure the component is included in your client-side bundle.
+Before calling `hydrate`, you must first `import` any `x-interaction` components that will be rendered on the page. The components register themselves with the `x-interaction` runtime when imported; you don't need to do anything with the imported component. This will also ensure the component is included in your client-side bundle. Similarly if the component that you're server side rendering is just a component that you've created through `withActions`, make sure you import that component along with its registerComponent invokation.
 
 Because `hydrate` expects the wrappers to be present in the DOM when called, it should be called after [`DOMContentLoaded`](https://developer.mozilla.org/en-US/docs/Web/Events/DOMContentLoaded). Depending on your page structure, it might be appropriate to hydrate the component when it's scrolled into view.
 
--- a/e2e/e2e.test.js
+++ b/e2e/e2e.test.js
@@ -1,0 +1,62 @@
+/**
+ * @jest-environment node
+ */
+
+const { h } = require('@financial-times/x-engine') // required for <GreetingComponent>
+const { Serialiser, HydrationData } = require('@financial-times/x-interaction')
+const puppeteer = require('puppeteer')
+const ReactDOMServer = require('react-dom/server')
+const express = require('express')
+import React from 'react'
+import { GreetingComponent } from './common'
+
+describe('x-interaction-e2e', () => {
+	let browser
+	let page
+	let app
+	let server
+
+	beforeAll(async () => {
+		app = express()
+		server = app.listen(3004)
+		app.use(express.static(__dirname))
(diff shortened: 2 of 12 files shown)


components/x-interaction/readme.md
```diff
@@ -179,7 +179,7 @@ When rendered on the server side, components output an extra wrapper element, wi
 
 `x-interaction` exports a function `hydrate`. This should be called on the client side. It inspects the global serialisation data on the page, uses the identifiers to find the wrapper elements, and calls `render` from your chosen `x-engine` client-side runtime to render component instances into the wrappers.
 
-Before calling `hydrate`, you must first `import` any `x-interaction` components that will be rendered on the page. The components register themselves with the `x-interaction` runtime when imported; you don't need to do anything with the imported component. This will also ensure the component is included in your client-side bundle.
+Before calling `hydrate`, you must first `import` any `x-interaction` components that will be rendered on the page. The components register themselves with the `x-interaction` runtime when imported; you don't need to do anything with the imported component. This will also ensure the component is included in your client-side bundle. Similarly if the component that you're server side rendering is just a component that you've created through `withActions`, make sure you import that component along with its registerComponent invokation.
```
apaleslimghost, line 182: **praise**: this is a great addition to the docs, i can see how it would be easily missed and cause hard-to-diagnose issues, nice!

e2e/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "@financial-times/x-interaction": "file:../packages/x-interaction"
+  },
+  "devDependencies": {
+    "@financial-times/x-rollup": "file:../../packages/x-rollup",
+    "@financial-times/x-test-utils": "file:../../packages/x-test-utils",
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2",
+    "webpack-cli": "^4.8.0",
+    "puppeteer": "^10.4.0"
+  },
```
apaleslimghost, line 25: **question**: how are these supposed to be installed? maybe this folder should be added to `workspaces` in the top-level `package.json` so Athloi can handle it?

e2e/e2e.test.js
```diff
@@ -0,0 +1,57 @@
...
+		const ClientComponent = () => {
+			// main.js is the transpiled version of index.js, which contains the registered GreetingComponent, and invokes hydrate
+			return <script type="module" src="./main.js" charset="utf-8"></script>
+		}
+
+		const serialiser = new Serialiser()
+		const htmlString =
+			ReactDOMServer.renderToString(<GreetingComponent serialiser={serialiser} />) +
+			ReactDOMServer.renderToString(<HydrationData serialiser={serialiser} />) +
+			ReactDOMServer.renderToString(ClientComponent())
```
apaleslimghost, line 36: **suggestion**: this could use a Fragment instead? i think it's more readable

```suggestion
		const htmlString = ReactDOMServer.renderToString(<>
			<GreetingComponent serialiser={serialiser} />
			<HydrationData serialiser={serialiser} />
			<ClientComponent />
		</>)
```

e2e/index.js
```diff
@@ -0,0 +1,4 @@
+import { hydrate } from '@financial-times/x-interaction'
+require('./common')
```
apaleslimghost, line 2: **suggestion**: use `import` here for consistency

```suggestion
import './common'
```

e2e/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "name": "e2e-x-dash",
+  "version": "0.0.0",
+  "description": "This module enables you to write x-dash components that respond to events and change their own data.",
+  "keywords": [
+    "x-dash"
+  ],
+  "author": "",
+  "license": "ISC",
+  "scripts": {
+    "server": "node server.mjs",
```
apaleslimghost, line 11: **question**: what's the purpose of this? it's not used by the tests, is it only for manual testing?

apaleslimghost (approved):
a handful of non-blocking suggestions and questions, this looks really good!

i think it makes most sense for the e2e tests to run in the workflow in parallel with the unit tests.

i seem to remember fixing the issue of closing Jest when running a server before, i'll see what i can dig up.

e2e/package.json
serena97: ah yes for this one, I was attempting to refactor the e2e test, so that the server runs in parallel and not in the same process as jest, to fix the issue of closing Jest when running a server 
https://stackoverflow.com/questions/51968480/jest-detects-open-handle-with-express-app

apaleslimghost (approved):
excellent work @serena97!

PR #621, opened 2021-11-01
added back to top link to liveBlogPost

noelenwenede:
## Back to top link on x-live-blog-post component

This PR adds a `back to top` link to the bottom of `x-live-blog-post` component. It is a minor feature and with it, a new props have been added to the component. `backToTop`  which can be a function or a string. The aim is to have either `type` passed to achieve same result. The function `type` was added for more fine-grained control of the scrolling with javascript if the consuming component desires that. Pass a string `top` or `#top` to allow the browser handle the scrolling back to top. A string also requires the an element in the `DOM` to have same value as the id attribute.

### DIFF
--- a/components/x-live-blog-post/readme.md
+++ b/components/x-live-blog-post/readme.md
@@ -52,5 +52,5 @@ Feature             | Type   | Notes
 `publishedTimestamp`| String | Deprecated - ISO timestamp of publish date
 `articleUrl`        | String | Url of the main article that includes this post
 `showShareButtons`  | Bool   | default: `false` - Shows social media share buttons when `true`
-`backToTop`  		| Function | String   | Shows the back to top link at the bottom of posts and manages navigating to `selected top` with a javascript function or a hashed link. Only pass in. Please call event.preventDefault() at the top level if this prop is a function. If prop is a string, add the string as the `id` to the element that represents the top.
+`backToTop`  		| Function | String   | Shows the back to top link at the bottom of posts and manages navigating to `selected top` with a javascript function or a hashed href. Please call event.preventDefault() at the top level if this prop is a function. If prop is a string, add the string as the `id` to the element that represents the top.
 
--- a/components/x-live-blog-post/readme.md
+++ b/components/x-live-blog-post/readme.md
@@ -52,3 +52,6 @@ Feature             | Type   | Notes
 `publishedTimestamp`| String | Deprecated - ISO timestamp of publish date
 `articleUrl`        | String | Url of the main article that includes this post
 `showShareButtons`  | Bool   | default: `false` - Shows social media share buttons when `true`
+`topRef`  			| String   | Shows the back to top link at the bottom of posts and navigates to the hashed section of the page. No defaults, example `#top`
+`backToTop`  		| Function   | Shows the back to top link at the bottom of posts and manages navigating to `selected top` with a javascript function. No defaults. Only pass in `topRef` or `backToTop` props to the component. Please call event.preventDefault() at the top level of this function.
+
--- a/components/x-live-blog-post/readme.md
+++ b/components/x-live-blog-post/readme.md
@@ -52,6 +52,5 @@ Feature             | Type   | Notes
 `publishedTimestamp`| String | Deprecated - ISO timestamp of publish date
 `articleUrl`        | String | Url of the main article that includes this post
 `showShareButtons`  | Bool   | default: `false` - Shows social media share buttons when `true`
-`backToTopRef`  			| String   | Shows the back to top link at the bottom of posts and navigates to the hashed section of the page. No defaults, example `#top`
-`backToTopFunction`  		| Function   | Shows the back to top link at the bottom of posts and manages navigating to `selected top` with a javascript function. No defaults. Only pass in `backToTopRef` or `backToTopFunction` props to the component. Please call event.preventDefault() at the top level of this function.
+`backToTop`  		| Function | String   | Shows the back to top link at the bottom of posts and manages navigating to `selected top` with a javascript function or a hashed link. Only pass in. Please call event.preventDefault() at the top level if this prop is a function. If prop is a string, add the string as the `id` to the element that represents the top.
 
--- a/components/x-live-blog-post/readme.md
+++ b/components/x-live-blog-post/readme.md
@@ -52,6 +52,6 @@ Feature             | Type   | Notes
 `publishedTimestamp`| String | Deprecated - ISO timestamp of publish date
 `articleUrl`        | String | Url of the main article that includes this post
 `showShareButtons`  | Bool   | default: `false` - Shows social media share buttons when `true`
-`topRef`  			| String   | Shows the back to top link at the bottom of posts and navigates to the hashed section of the page. No defaults, example `#top`
-`backToTop`  		| Function   | Shows the back to top link at the bottom of posts and manages navigating to `selected top` with a javascript function. No defaults. Only pass in `topRef` or `backToTop` props to the component. Please call event.preventDefault() at the top level of this function.
+`backToTopRef`  			| String   | Shows the back to top link at the bottom of posts and navigates to the hashed section of the page. No defaults, example `#top`
+`backToTopFunction`  		| Function   | Shows the back to top link at the bottom of posts and manages navigating to `selected top` with a javascript function. No defaults. Only pass in `backToTopRef` or `backToTopFunction` props to the component. Please call event.preventDefault() at the top level of this function.
 
--- a/components/x-live-blog-post/src/LiveBlogPost.jsx
+++ b/components/x-live-blog-post/src/LiveBlogPost.jsx
@@ -27,17 +27,17 @@ const LiveBlogPost = (props) => {
 
 	let backToTopProps = {}
 
-	if (backToTop) {
-		backToTopProps = {
-			...backToTopProps,
-			onClick: backToTop
-		}
-	}
-
 	if (topRef) {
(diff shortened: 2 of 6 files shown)


components/x-live-blog-post/src/LiveBlogPost.jsx
```diff
@@ -39,7 +58,21 @@ const LiveBlogPost = (props) => {
...
 			/>
-			{showShareButtons && <ShareButtons postId={id || postId} articleUrl={articleUrl} title={title} />}
+			<div className={styles['live-blog-post__bottom-controls']}>
+				{showShareButtons && <ShareButtons postId={id || postId} articleUrl={articleUrl} title={title} />}
+				{(Boolean(backToTop) || Boolean(topRef)) && (
+					// eslint-disable-next-line jsx-a11y/click-events-have-key-events
+					<a
+						{...backToTopProps}
+						aria-labelledby="Back to top"
+						{...backToTopProps}
```
GlynnPhillips, line 68: **Question:** Is this meant to be here twice?

components/x-live-blog-post/src/LiveBlogPost.scss
```diff
@@ -74,3 +70,22 @@
...
+
+.live-blog-post__bottom-controls .back-to-top {
+	@include oTypographySans($scale: 1);
+	color: oColorsByName('teal');
+	text-decoration: underline;
+	margin-left: auto;
+}
+
+.live-blog-post__bottom-controls .back-to-top:hover {
+	cursor: pointer;
```
GlynnPhillips, line 90: **Question:** Do you know why this is needed? shouldn't this be default behaviour for an anchor?

components/x-live-blog-post/src/LiveBlogPost.scss
```diff
@@ -74,3 +70,22 @@
...
+
+.live-blog-post__bottom-controls {
+	display: flex;
+	justify-content: space-between;
+	align-items: center;
+	flex-wrap: wrap;
+	margin-top: oSpacingByName('s6');
+}
+
+.live-blog-post__bottom-controls .back-to-top {
```
GlynnPhillips, line 82: **Talking point:** It feels like some of these classes don't really align to BEM very well and I would like to see what other people think.

`live-blog-post__bottom-controls` I feel like the `bottom` part of this is not required info and if its styles were different based on its position then that should be a modifer. So I am thinking this should just be `.live-blog-post__controls`. What do you think?

`back-to-top` suggests that this is a new block level and that it is not tied to its parent block (the controls element) however the nesting also suggests that they are related.

I have a funny relation with nesting and BEM I can never quite decide on how to combine them both. I think a single level of nesting with BEM can be useful to make sure that relationship is clear and defined but also BEM should do this as well via its naming.

In this instance I think this class should be something more like `live-blog-post-controls__back-to-top`

What do others think?

components/x-live-blog-post/readme.md
```diff
@@ -52,3 +52,6 @@ Feature             | Type   | Notes
 `publishedTimestamp`| String | Deprecated - ISO timestamp of publish date
 `articleUrl`        | String | Url of the main article that includes this post
 `showShareButtons`  | Bool   | default: `false` - Shows social media share buttons when `true`
+`topRef`  			| String   | Shows the back to top link at the bottom of posts and navigates to the hashed section of the page. No defaults, example `#top`
+`backToTop`  		| Function   | Shows the back to top link at the bottom of posts and manages navigating to `selected top` with a javascript function. No defaults. Only pass in `topRef` or `backToTop` props to the component. Please call event.preventDefault() at the top level of this function.
```
GlynnPhillips, line 56: **Question:** Are there reasons why we have decided to make this a function (maybe something in the app?) rather than just allowing the browser to handle the navigating to the top hash location?

GlynnPhillips (commented):
Some comments I would like to discuss before approving

components/x-live-blog-post/readme.md
GlynnPhillips: Sorry - I just remembered you mentioned this in the PR description (I forgot whilst reading code)

GlynnPhillips: **Question:** Is there a requirement for 2 type of behaviour at this stage?

**Feedback:** I am not sure the options (`topRef ` & `backToTop `) are obvious that they related bit provide different functionality. I think it might be nicer to make this clearer if we really need more sets of functionality

components/x-live-blog-post/src/LiveBlogPost.scss
noelenwenede: Yes, you're right. I'll fix them wiith your suggestions. Thanks 🙌🏽

components/x-live-blog-post/readme.md
noelenwenede: In terms of the browser controls scrolling to the top, I assumed that in itself may be the best user experience. Since the function is going to be defined by the consumer, I figured it wasn't much work to add a callback function as well. Also a thing Olga may appreciate too.

For the names, it's difficult to tell without reading the docs. So I'll try `backToTopRef` and `backToTopFunction`. That should give more context.

components/x-live-blog-post/src/LiveBlogPost.scss
noelenwenede: In this case it didn't work that way, so I had to add it in.

components/x-live-blog-post/src/LiveBlogPost.jsx
noelenwenede: 🤦🏾‍♂️ I'll clean that up now.

components/x-live-blog-post/src/LiveBlogPost.scss
noelenwenede: Not sure why that happened.

components/x-live-blog-post/readme.md
GlynnPhillips: That sounds reasonable - I just always question building features that may never get used. If they are not used its just extra functionality to have. I agree it might be nicer to let apps manage the scroll and I think in the app (might be wrong) that might have some other benefits about keeping that position which is why we are building this.

This is just a suggestion so feel free to ignore it. What if we just had 1 option called `backToTop`but could accept either a function or a string? I think this might make the docs clearer that theres a single back to top option but if you want to manage scrolling yourself you pass it a function instead of a string.

From reading the docs as they are right now I at first thought to get back to top functionality I needed to pass in both `topRef` & `backToTop`

noelenwenede: Agreed. It's a robust solution. I'll change the implementation. Thanks this helps a lot.

components/x-live-blog-post/src/LiveBlogPost.scss
GlynnPhillips: I am interested to try and understand what the problem is here rather than just apply the styling to fix it. Maybe we could look more into this before we merge

noelenwenede: Sure, I'll do some investigation. I do think it may have to do with this line `@include oTypographySans($scale: 1);` added to the styling.

components/x-live-blog-post/readme.md
```diff
@@ -52,3 +52,5 @@ Feature             | Type   | Notes
 `publishedTimestamp`| String | Deprecated - ISO timestamp of publish date
 `articleUrl`        | String | Url of the main article that includes this post
 `showShareButtons`  | Bool   | default: `false` - Shows social media share buttons when `true`
+`backToTop`  		| Function | String   | Shows the back to top link at the bottom of posts and manages navigating to `selected top` with a javascript function or a hashed href. Please call event.preventDefault() at the top level if this prop is a function. If prop is a string, add the string as the `id` to the element that represents the top.
```
GlynnPhillips, line 55: **Minor improvements:** Looks like the indentation is slight off here also some small improvements to the setence structure to help keep it clear.

```suggestion
`backToTop`                | String | Function   | Shows the back to top link at the bottom of posts and manages navigating to `selected top` with a javascript function or a hashed href (string). If this prop is a string it will rely on standard browser behaviour to navigate to the element `id` provided that represents the top. If this prop is a function then that function should control the experience of navigating/scrolling to the top position. When using a function please call event.preventDefault() at the top level.
```

components/x-live-blog-post/src/LiveBlogPost.scss
GlynnPhillips: Running this locally with storybook in chrome and firefox its works fine for me without these styles

noelenwenede: @GlynnPhillips  This has to do with the browser behaviour and the design choices I made here. When an `a` tag is missing the `href` attribute, the browser takes away the `cursor: pointer` `:hover` state and renders it like a regular text. So for situations when `backToTop` is a string, it should behave like a regular `a` tag but when it is a function it wouldn't. So this override is required ensure consistent behaviour.

GlynnPhillips: Thanks @noelenwenede - that makes sense but unfortunately made me aware of this I hadn't noticed it before. I wonder if we should switch it to be a button in the instance of the prop being a function because I guess its no longer an anchor. What do you think? its extra complexity

noelenwenede: Yes that makes sense for accessibility. 🙌🏽

GlynnPhillips (approved):
Thanks for taking on all the changes :)

PR #622, opened 2021-11-01
outline-offset and background colour change for the form submit save button

gitGremlin:
Ticket is [here](https://financialtimes.atlassian.net/jira/software/c/projects/ADSDEV/boards/803?modal=detail&selectedIssue=ADSDEV-898)

Following the example laid out by [this PR](https://github.com/Financial-Times/x-dash/pull/620) by @oliverturner 

We are using `teal-40` to make an offset around the button for when it is in focus. Originally I juggled with a few colours to see if I could find one that had a contrast ratio of 3:1 with both the `paper` and the `teal-40` colour, but there might just not be one.

This keeps the design consistent with Oliver's PR.

Changes are:
- button hover state changed to `teal-40`
- button outline matches hover state to be `teal-40`
- offset for outline of `3px` to make it more obvious

This should result in a contrast of 6.75:1 for the `teal-40` compared with the `paper` background.

Before
Screenshot 2021-11-01 at 19 27 09

After
Screenshot 2021-11-01 at 19 09 22

### DIFF
--- a/components/x-privacy-manager/src/privacy-manager.scss
+++ b/components/x-privacy-manager/src/privacy-manager.scss
@@ -74,6 +74,8 @@
 	padding: 0 oSpacingByName(m12);
 
 	&:focus-visible {
-		outline: 2px solid oColorsByName('teal-100');
+		outline: 2px solid oColorsByName('teal-40');
+		outline-offset: 3px;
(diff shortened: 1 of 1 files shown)


alevito (approved):
Good job!

PR #623, opened 2021-11-08
Hide back-to-top on first Live Blog Post

noelenwenede:
This is a follow-up PR from https://github.com/Financial-Times/x-dash/pull/621. The back to top button/link is not required on the first post. The initial plan was to let the consumers create overrides in their `css` to achieve this. Because the style names are changed on `build` (CSS modularisation), it becomes harder for the consumer of this component to override the styles as every new version released comes with a new class name.

This PR adds the styling direct to the component and also updates the storybook to reflect this. This CSS class names are also added directly to the styles without any modularisation so consumers can directly select these elements and modify them.

### DIFF
--- a/components/x-live-blog-post/src/LiveBlogPost.jsx
+++ b/components/x-live-blog-post/src/LiveBlogPost.jsx
@@ -32,16 +32,10 @@ const LiveBlogPost = (props) => {
 				return ref.includes('#') ? ref : `#${ref}`
 			}
 			BackToTopComponent = (
-				/**
-				 * (Ref: LBPC101) The class name `live-blog-post-controls__back-to-top-link` has been added to help consumers
-				 * of this component select and override css behaviours if it is required. Due to modularization.
-				 * the generated classnames will be differen on different releases so this addition provides a steady
-				 * selector for overrides
-				 */
 				<a
 					href={processTopRef(backToTop)}
 					aria-labelledby="Back to top"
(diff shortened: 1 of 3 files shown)


components/x-live-blog-post/src/LiveBlogPost.jsx
```diff
@@ -47,7 +47,7 @@ const LiveBlogPost = (props) => {
 				<button
 					onClick={backToTop}
 					aria-labelledby="Back to top"
-					className={styles['live-blog-post-controls__back-to-top-button']}
+					className={`live-blog-post-controls__back-to-top-button ${styles['live-blog-post-controls__back-to-top-button']}`}
```
GlynnPhillips, line 50: **Question:** Have you added it in this way so a consumer could decide to override the `display:none` if they wanted to?

noelenwenede: Yes, just exposing that option too.

GlynnPhillips: Ok, you know what comments coming next 😉 is there a requirement for this? the reason I ask is that if I saw this code in the future I wouldn't have a clue why we are applying the same class names twice. This could be solved with a code comment but if theres no need for it right now I would lean towards not doing it until its needed

GlynnPhillips (approved):

PR #624, opened 2021-11-10
removed fragment from live-blog-post component

noelenwenede:
This PR contains a minor change to the structure of the `live-blog-post` component. 

The current component uses `Fragment` from `@financial-times/x-engine` package which doesn't render properly on `ft-app`. I've replaced the Fragment with a `div` and adjusted the styles accordingly.

This specific feature has had 3 releases which is specifically bad for our versions and I take the blame for this. 

I have also tested locally with the local package on both `ft-app` and `next-article` to ensure this issue doesn't show up on the published package.

### DIFF
--- a/components/x-live-blog-post/src/LiveBlogPost.jsx
+++ b/components/x-live-blog-post/src/LiveBlogPost.jsx
@@ -1,5 +1,4 @@
-/* eslint-disable jsx-a11y/no-static-element-interactions */
-import { h, Fragment } from '@financial-times/x-engine'
+import { h } from '@financial-times/x-engine'
 import ShareButtons from './ShareButtons'
 import Timestamp from './Timestamp'
 import styles from './LiveBlogPost.scss'
@@ -35,7 +34,7 @@ const LiveBlogPost = (props) => {
 				<a
 					href={processTopRef(backToTop)}
 					aria-labelledby="Back to top"
-					className={styles['live-blog-post-controls__back-to-top-link']}
+					className={styles['live-blog-post-controls__right__back-to-top-link']}
 				>
 					Back to top
 				</a>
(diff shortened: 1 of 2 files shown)


boyewale:
Did you mean to add the package-lock.json files? They both seem to be new to this commit/PR.

noelenwenede:
> Did you mean to add the package-lock.json files? They both seem to be new to this commit/PR.

No, came in as part of the experiment.

boyewale (approved):

components/x-live-blog-post/src/LiveBlogPost.jsx
```diff
@@ -35,7 +34,7 @@ const LiveBlogPost = (props) => {
 				<a
 					href={processTopRef(backToTop)}
 					aria-labelledby="Back to top"
-					className={styles['live-blog-post-controls__back-to-top-link']}
+					className={styles['live-blog-post-controls__right__back-to-top-link']}
```
GlynnPhillips, line 37: **Suggestion:** I am getting very confused with these style names 😂 I know its hard to make BEM work when you also want to create components with prefixes, so this is a tricky setup.

I think how we get around that is to have something like this `prefix-block__element--modifier`

So in this case I am guessing the prefix is `live-blog-post` the block is `controls` and the element is `back-to-top-link`.

The `right` in the middle of all of this don't quite make sense. The question I would have is it the block or the element that should be positioned right?

Depending on the answer to that you could have the modifier on the block`live-blog-post-controls--right` or if its the element that is positioned right it could be `live-blog-post-controls__back-to-top--right`

I am guessing because you have put the `right` after the block then the first option is more relevant

noelenwenede: My BEM skills are 2/10 right now 😂.

Yes the first options is more relevant.  the entire block should be on the right.

I'll amend.

GlynnPhillips: Don't worry I also struggle with this all the time. Its a tricky naming to get your head around especially when combined with prefixes.

When I first saw it I hated it and was against using it but once I started to understand it more I started to see some benefits.. but I still mess it up regularly

PR #625, opened 2021-11-11
refactored the x-live-blog-component.

noelenwenede:
This PR refactors the `x-live-blog-component` and also reintroduced the use of Fragment within the component which was removed before.

The refactor was done mainly because the CSS wasn't BEM compliant.


GlynnPhillips (approved):
:chef-kiss:

PR #627, opened 2021-11-22
CPP-642 migrate from circleci/* to cimg/*

serena97:
These images will be EOL from Dec 31st, and we should migrate to the modern cimg/* equivalents.


serena97:
automatically closed 🤖

PR #629, opened 2021-11-22
CPP-675 Update to npm 8

ivomurrell:
We are moving to npm 8 so we are ready to migrate away from bower. There don't appear to be any breaking changes as part of the migration for this repository, so it was as simple as pinning a new npm version with Volta (you can also update your npm globally to test the changes) and making sure CircleCI uses the right npm version too. We have added a `preinstall` script to each package's `package.json` that will make sure that npm 8 is selected when running an install.

### DIFF
--- a/components/x-follow-button/package.json
+++ b/components/x-follow-button/package.json
@@ -9,7 +9,8 @@
   "scripts": {
     "prepare": "bower install && npm run build",
     "build": "node rollup.js",
-    "start": "node rollup.js --watch"
+    "start": "node rollup.js --watch",
+    "preinstall": "[ \"$INIT_CWD\" != \"$PWD\" ] || npm_config_yes=true npx check-engine"
   },
   "keywords": [],
   "author": "",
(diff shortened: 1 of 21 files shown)


components/x-follow-button/package.json
```diff
@@ -26,5 +27,12 @@
   "dependencies": {
     "@financial-times/x-engine": "file:../../packages/x-engine",
     "classnames": "^2.2.6"
+  },
+  "engines": {
+    "node": "12.x",
+    "npm": "7.x || 8.x"
```
serena97, line 33: **question:** Since we should be upgrading to npm 8 to get ready to migrate to bower, would it make sense to remove the 7.x for each engines object?

serena97 (dismissed):

package.json
```diff
@@ -40,6 +41,7 @@
     "@types/jest": "26.0.0",
     "@typescript-eslint/parser": "^3.0.0",
     "babel-loader": "^8.0.4",
+    "check-engines": "^1.5.0",
```
serena97, line 44: **question:** do we need this?

serena97 (dismissed):
same comment as the other PR - looks good to me overall, good spot on adding the preinstall script. think we'll need to remove the extra check-engines dev dependency

package.json
ivomurrell: See https://github.com/Financial-Times/dotcom-page-kit/pull/917#discussion_r758259368

serena97 (approved):
looks good to me, thanks for doing the check-engine change!

PR #630, opened 2021-11-24
Ci 890 live blog post tracker

noelenwenede:
The implementation of post level tracking for live blogs was previously done on `next-article`. As part of the process of implementing post level tracking on `ft-app`, the implementation of post-tracker has been moved to the `x-live-blog-wrapper` component. With this implementation, a new property (object) has been added to `x-live-blog-wrapper` component called `postTrackerConfig` (with the following properties/fields `onEntersViewport` [callback function], `onRead` [callback function], `onError` [callbackfunction], `usePostTracker` [boolean]).

The post-tracker class creates an intersection-observer instance and observes the visibility of posts that exist in the live-blog-wrapper component. When a new post is inserted, the wrapper component dispatches an event. The post tracker class listens for this event and updates the number of posts being observed.

### DIFF
--- a/components/x-live-blog-wrapper/src/LiveBlogWrapper.jsx
+++ b/components/x-live-blog-wrapper/src/LiveBlogWrapper.jsx
@@ -1,9 +1,19 @@
-import { h } from '@financial-times/x-engine'
+import { h, Component } from '@financial-times/x-engine'
 import { LiveBlogPost } from '@financial-times/x-live-blog-post'
 import { withActions } from '@financial-times/x-interaction'
 import { normalisePost } from './normalisePost'
 import { dispatchEvent } from './dispatchEvent'
 import { registerComponent } from '@financial-times/x-interaction'
+import { PostTracker } from './lib/post-tracker'
+
+/**
+ * @typedef PostTrackerConfig
+ * @type {object}
+ * @property {Function} onEntersViewport - function to be called when the criteria for `view` in Post tracker has been fulfiled
+ * @property {Function} onRead - function to be called when the criteria for `read` in Post tracker has been fulfiled
+ * @property {Function} onError - function to be called when PostTracker encounters an error
+ * @property {Boolean} usePostTracker - condition checked before creating an instance of post tracker
+ */
 
 const withLiveBlogWrapperActions = withActions({
 	insertPost(newPost, wrapper) {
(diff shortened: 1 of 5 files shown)


components/x-live-blog-wrapper/src/LiveBlogWrapper.jsx
```diff
@@ -20,45 +30,124 @@ const withLiveBlogWrapperActions = withActions({
...
+			minMillisecondsToReport: 5000,
+			returnVisibleElement: true,
+			observerUpdateEventString: 'LiveBlogWrapper.INSERT_POST',
+			liveBlogWrapperQuery: `div[data-live-blog-wrapper-id="${id}"]`,
+			liveBlogWrapper: this.props.liveBlogWrapperElementRef
+				? this.props.liveBlogWrapperElementRef.current
+				: undefined,
+			onEntersViewport: (event) => callbacks.onEntersViewport(event),
+			onRead: (event) => callbacks.onRead(event),
+			onError: (event) => callbacks.onError(event)
```
GlynnPhillips, line 107: **Suggestion/Feedback:** This could error if a consuming app doesn't pass these properties in as they won't be functions. I think we need some sanitising of whats passed in to make sure we can set up a post tracker

GlynnPhillips (changes requested):
I have made 1 small comment - if I am miss understanding it then let me know

components/x-live-blog-wrapper/src/LiveBlogWrapper.jsx
noelenwenede: So I wrote that as a `sub function` to the `setup function`. The setup function does the sanitisation here https://github.com/Financial-Times/x-dash/blob/bbe48a46699f09ca60c92a3a56e89273c83f0ded/components/x-live-blog-wrapper/src/LiveBlogWrapper.jsx#L65 and errors out here https://github.com/Financial-Times/x-dash/blob/bbe48a46699f09ca60c92a3a56e89273c83f0ded/components/x-live-blog-wrapper/src/LiveBlogWrapper.jsx#L71.

Would you prefer the sanitisation inside the sub function? I left all sanitisation inside the setup function.

GlynnPhillips: I knew I had probably missed something 😄 Sorry that looks fine - just hard to follow it in the PR

GlynnPhillips (approved):
Looks good to me - It might be worth checking with someone form the apps team that this will work in the app as I have no knowledge of how tracking works in the app

noelenwenede:
> Looks good to me - It might be worth checking with someone form the apps team that this will work in the app as I have no knowledge of how tracking works in the app

It should work fine on the app. The call back functions only reports local visibility tracking data, the consumers would manage sending tracking data and application context.

My worry is with `next-article` and how it hydrates on the frontend. I'll do some test on it to see if it works fine.

GlynnPhillips (approved):

PR #631, opened 2021-11-24
Update cross-component dependencies

rowanbeentje:
A few dep bumps...

The App currently installs many (most?) x-dash components, but due to various versions of different components referencing each other, we pull in several components twice at different versions, and x-engine four (!) times. To try to resolve this, I've gone through the various components still on branches and updated them, and then did a review of the components on `main`. I've found a few little things to update:

- x-podcast-launchers was still pulling in a prerelease version of x-follow-button, so I've updated that to use its sibling component from the current version instead.
- x-teaser-timeline was using an old prerelease version of x-article-save-button which was still using x-engine@^1.0.0-beta.7. I've updated this to use a new prerelease which updates to x-engine@^6.2.9.
- x-privacy-manager had a few pinned [dev- and prod] dependencies, which doesn't match the unpinned depedencies in use elsewhere, so I've unpinned those.

After these changes, storybook compilation was failing because it was trying to build the storybook for x-article-save-button, which was in the node_modules tree for x-teaser-timeline. This was failing because node_modules is explicitly excluded from the webpack loader transpilation setup, but I think the intention here is not to include node_modules storybooks anyway, so I've updated storybook/main.js to specify that.

### DIFF
--- a/components/x-teaser-timeline/package.json
+++ b/components/x-teaser-timeline/package.json
@@ -17,7 +17,7 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "@financial-times/x-article-save-button": "0.0.11",
+    "@financial-times/x-article-save-button": "0.0.13",
     "@financial-times/x-engine": "file:../../packages/x-engine",
     "@financial-times/x-teaser": "file:../x-teaser",
     "classnames": "^2.2.6",
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,4 +1,4 @@
 module.exports = {
-	stories: ['../components/**/storybook/index.jsx'],
+	stories: ['../components/**!(node_modules)/storybook/index.jsx'],
 	addons: ['@storybook/addon-essentials']
 }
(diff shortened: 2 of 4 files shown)


components/x-teaser-timeline/package.json
```diff
@@ -17,7 +17,7 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "@financial-times/x-article-save-button": "0.0.11",
+    "@financial-times/x-article-save-button": "0.0.13",
```
apaleslimghost, line 20: **question**: should this be a `file:` reference like the others?

rowanbeentje: This one is still on a prerelease branch with accompanying PR (https://github.com/Financial-Times/x-dash/pull/164/files) 😬 Not ideal!

apaleslimghost: oh nooooooo

apaleslimghost (approved):

rowanbeentje:
Thank you! 🎉

PR #632, opened 2021-11-25
moved tracking setup from componentDidMount hook to componentDidUpdate

noelenwenede:
As part of tracking post visibility. A PostTracker class was introduced to x-dash components. The PostTracker will be used by `x-live-blog-wrapper` component to manage tracking.

On `ft-app` the posts props doesn't get populated on the `componentDidMount` state hook. The setup for PostTracker was done in `componentDidMount` hook and since the posts didn't exist in the DOM, the IntersectionObserver doesn't pick them up.

This PR moves the setUp to the `componentDidUpdate` state hook. Within this hook, we check for the existence of a local PostTracker instance and also changes to the current number of posts in the component. When we have posts and no instance of PostTracker, we create a new PostTracker instance. When we have a PostTracker instance and the number of post changes (specifically for live updates) we trigger the PostTrackers `observeNewPosts()` function. This class also keep track of the actual DOM element being tracked so calling it multiple times has no side effects. The class also listens for `LiveBlogWrapper.INSERT_POST` events to update the elements being tracked.

NB: Because of the way ft.com hydrates (on the client after serverside rendering), this implementation doesn't work. I have exported the PostTracker class for ft.com to make it's own implementation.

### DIFF
--- a/components/x-live-blog-wrapper/readme.md
+++ b/components/x-live-blog-wrapper/readme.md
@@ -2,7 +2,6 @@
 
 This module displays a list of live blog posts using `x-live-blog-post` component. It also connects to an event stream which provides updates for the list. Based on these update events this component will add, remove and update `x-live-blog-post` components in the list.
 
-
 ## Installation
 
 This module is supported on Node 12 and is distributed on npm.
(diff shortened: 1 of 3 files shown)


GlynnPhillips:
> NB: Because of the way ft.com hydrates (on the client after serverside rendering), this implementation doesn't work. I have exported the PostTracker class for ft.com to make it's own implementation.

I am mainly interested by this as it feels like a shame to do things differently. Can you provide examples of how ft.com and ft-app would use this so I can get a better picture

noelenwenede:
> > NB: Because of the way ft.com hydrates (on the client after serverside rendering), this implementation doesn't work. I have exported the PostTracker class for ft.com to make it's own implementation.
> 
> I am mainly interested by this as it feels like a shame to do things differently. Can you provide examples of how ft.com and ft-app would use this so I can get a better picture

So ft-app just needs to pass the `postTrackerConfig` as a prop to `LiveBlogWrapper` component and everything works fine since it's fully clientside rendered.

ft.com will continue as it was before, only replacing the local IntersectionObserverTracker class with PostTracker class. All references to IntersectionObserverTracker will be replaced by PostTracker imported from @financial-times/x-live-blog-wrapper in this file https://github.com/Financial-Times/next-article/blob/457a9979feead9831b4b374ced150b5125430b42/client/components/live-blogs/live-blog-tracker.js#L2. I'm not sure how hydration works or how it affects the lifecycle hooks but I noticed that most of our jsx components on ft.com don't use them and a majority of the app state on the client is handled by plain javascript functions.

GlynnPhillips:
Thanks @noelenwenede - what I am trying to decide is whether its worth having 2 ways of doing it when we only have 2 consumers. I assuming ft-app could do it the same way that ft.com does? if thats the case are there any benefits to ft-app doing it differently? Thats not a trick question I don't know the answers to this :)

noelenwenede:
> Thanks @noelenwenede - what I am trying to decide is whether its worth having 2 ways of doing it when we only have 2 consumers. I assuming ft-app could do it the same way that ft.com does? if thats the case are there any benefits to ft-app doing it differently? Thats not a trick question I don't know the answers to this :)

My thought process was more inline with encapsulating the logic and having the `x-live-blog-wrapper` component just emit the events required for tracking. So `ft-app` would be a reference to how it is expected to be done and `ft.com` would (show an alternative way to do it without hydration) manage it's own instance until server state can be transferred to the client without losing functions and state hooks. At some point we should have hydration working in a reactjs way with components on the client having the entire server state.

noelenwenede:
@GlynnPhillips also on `ft-app` live updates don't work same so tracking newly inserted post is impossible with events from the `liveBlogWrapper ref` (as this is not dispatched) which is what next-article does. The componentDidUpdate lifecycle hooks in the Wrapper component also helps handle tracking newly inserted posts.

GlynnPhillips (approved):

PR #633, opened 2021-12-08
Add support for redirectUrl prop in place of referrer

oliverturner:
When a user's choice has been saved we display an onward link. This change allows us to make it point back to the page from which they came.

Additionally, this PR moves the confirmation message to the area below the form controls to make the progression clearer to users.

Screenshot 2021-12-14 at 10 56 19

### DIFF
--- a/components/x-privacy-manager/package.json
+++ b/components/x-privacy-manager/package.json
@@ -27,15 +27,15 @@
   "dependencies": {
     "@financial-times/x-engine": "file:../../packages/x-engine",
     "@financial-times/x-interaction": "file:../x-interaction",
-    "classnames": "^2.2.6"
+    "classnames": "2.3.1"
   },
   "devDependencies": {
     "@financial-times/x-rollup": "file:../../packages/x-rollup",
     "@financial-times/x-test-utils": "file:../../packages/x-test-utils",
-    "bower": "^1.8.8",
-    "check-engine": "^1.10.1",
-    "fetch-mock-jest": "^1.3.0",
-    "sass": "^1.26.5"
+    "bower": "1.8.13",
+    "check-engine": "1.10.1",
+    "fetch-mock-jest": "1.5.1",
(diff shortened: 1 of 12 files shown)


components/x-privacy-manager/package.json
```diff
@@ -27,15 +27,15 @@
...
     "@financial-times/x-rollup": "file:../../packages/x-rollup",
     "@financial-times/x-test-utils": "file:../../packages/x-test-utils",
-    "bower": "^1.8.8",
-    "check-engine": "^1.10.1",
-    "fetch-mock-jest": "^1.3.0",
-    "sass": "^1.26.5"
+    "bower": "1.8.13",
+    "check-engine": "1.10.1",
+    "fetch-mock-jest": "1.5.1",
+    "sass": "1.44.0"
```
apaleslimghost, line 38: **question**:  were these pinned deliberately? our new Renovate config doesn't pin things any more, so i think it'll be annoyed about this

apaleslimghost (approved):
looks good!

**suggestion**: as it is, this is a breaking change, which would require a major release of every x-dash component. that's fine! but keeping it backwards compatible and releasing as a minor version would be a little less hassle :)

components/x-privacy-manager/package.json
oliverturner: Not exactly deliberately: this made me realise that my ~/.npmrc file has `save-exact=true`

oliverturner:
So, on the matter of the breaking nature of this change... it's a feature that has never actually been implemented in either of the 2 locations the component is used... so squinting a bit one _might_ conclude that it's safe to release as a minor version?

apaleslimghost:
if a breaking change falls in the woods...

apaleslimghost:
go ahead and release it as minor 😁

apaleslimghost (approved):

PR #634, opened 2021-12-13
FTDCS-36 - Fix Sentry error

JSRedondo:
Sentry error [Cannot read property 'items' of undefined](https://sentry.io/organizations/financial-times/issues/2770847631)
itemGroups without group led to a TypeError:
Cannot read property 'items' of undefined

### DIFF
--- a/components/x-teaser-timeline/src/lib/transform.js
+++ b/components/x-teaser-timeline/src/lib/transform.js
@@ -135,11 +135,13 @@ const getItemGroups = ({
 const getGroupAndIndex = (groups, position) => {
 	if (position > 0) {
 		const group = groups.findIndex((g) => g.items.some((item) => item.articleIndex === position - 1))
-		const index = groups[group].items.findIndex((item) => item.articleIndex === position - 1)
+		if(groups[group]){
+			const index = groups[group].items.findIndex((item) => item.articleIndex === position - 1)
 
-		return {
-			group: group,
-			index: index + 1
+			return {
+				group: group,
+				index: index + 1
+			}
 		}
 	}
 


components/x-teaser-timeline/src/lib/transform.js
```diff
@@ -135,11 +135,13 @@ const getItemGroups = ({
...
+		if(groups[group]){
+			const index = groups[group].items.findIndex((item) => item.articleIndex === position - 1)
 
-		return {
-			group: group,
-			index: index + 1
+			return {
+				group: group,
+				index: index + 1
+			}
```
gyss, line 144: what happens if `groups[group]` is undefined? I guess this function will return undefined and the exception will be thrown somewhere else? Have you tested it?

JSRedondo: If groups[group] is undefined then is returned:	
```
{
    group: 0,
    index: 0
}
```
This is a suggestion as I am not aware of the business logic; anyway, this is a simple way to avoid the TypeError.
Apparently, the only way to get groups[group] as undefined is that there isn't any match, so group (index of the match) is -1. Therefore groups[-1] is undefined.
Probably this code needs proper testing to make sure of it, but it's out of the scope of ft.com team, sorry! 😉

JSRedondo: If groups[group] is undefined then is returned:	
```
{
    group: 0,
    index: 0
}
```
This is a suggestion as I am not aware of the business logic; anyway, this is a simple way to avoid the TypeError.
Apparently, the only way to get groups[group] as undefined is that there isn't any match, so group (index of the match) is -1. Therefore groups[-1] is undefined.
Probably this code needs proper testing to make sure of it, but it's out of the scope of ft.com team, sorry! 😉

PR #635, opened 2021-12-22
ENT-503 Add Enterprise Sharing to Gift Article

fabiosikansi:
This PR adds the enterprise sharing option to x-gift-article.

We are moving enterprise sharing from its own [micro-site](https://enterprise-sharing.ft.com/) to FT.com. This change has been discussed with the accounts team in advance.

## Summary of Changes

* x-gift article will make a request to the enterprise-sharing API and add an new option to the sharing modal if the user is a B2B user:
image

* Users will be able to compare and choose the best sharing option for their use case.

* There is no breaking changes, If property `enterpriseApiBaseUrl` is not included or enterprise API request fails the modal will just work as usual, without the enterprise option.

I added 4 stories to the x-gift-article storybook, covering the main feature and some edge cases:

| Story | Use Case |
|-------|----------|
| With enterprise sharing | User has access to enterprise sharing, all interactions with ES API are mocked in the storybook |
| With enterprise sharing (no credits) | User has access but B2B licence has run out of credits, user can't use the feature but they will see a message to request more credits  |
| With enterprise sharing (first time user) | User has access to enterprise sharing but has never used the feature, they will see a message explaining its purpose |
| With enterprise sharing (request access) | User has a B2B licence without access to enterprise sharing, they will see a message explaining the feature and how to get access |
| With enterprise sharing (link generated) | An example of with the link generated |

** I also updated the story `With gift credits` to mock a `404` response from the Enterprise Sharing API. This is the default behaviour when a B2B user opens the link.

## How it works

* When opening the modal, a request to the enterprise sharing API will return if the user should see the enterprise option and some other properties like the number of shares allowed for their licence and if it is a first time user or not. Default values are set to not show the option if the API fails.
* The gift article component will make calls to different APIs based on the selected option (gift link or enterprise link)
* Enterprise and Gift links use different states, so the user can, in theory, create both links in the same session/view.
* Enterprise Link doesn't use the URL shortener service (we don't want to add a dependency to that and we are planning to implement something similar in our backend and return a shorter link by default)

## Tracking

Code changes include the broadcast of events when an enterprise link is created, based on the same payload for gift links but changing the `linkType` property.

## Testing

We created stories for all states, and the first one (`With enterprise sharing`) includes a mock for the response of our API when the user clicks on "create enterprise link".

### DIFF
--- a/components/x-gift-article/src/GiftArticle.jsx
+++ b/components/x-gift-article/src/GiftArticle.jsx
@@ -9,6 +9,7 @@ import EnterpriseApiClient from './lib/enterpriseApi'
 import { copyToClipboard, createMailtoUrl } from './lib/share-link-actions'
 import tracking from './lib/tracking'
 import * as updaters from './lib/updaters'
+import { ShareType } from './lib/constants'
 
 const isCopySupported =
 	typeof document !== 'undefined' && document.queryCommandSupported && document.queryCommandSupported('copy')
@@ -148,6 +149,7 @@ const withGiftFormActions = withActions(
 						if (giftCredits > 0 || giftCredits === 0) {
 							return {
 								...updaters.setAllowance(giftCredits, monthlyAllowance, nextRenewalDate),
+								shareType: enabled ? ShareType.enterprise : ShareType.gift,
 								enterpriseEnabled: enabled,
 								enterpriseLimit: limit,
 								enterpriseHasCredits: hasCredits,
--- a/components/x-gift-article/src/GiftArticle.jsx
+++ b/components/x-gift-article/src/GiftArticle.jsx
@@ -5,6 +5,7 @@ import Loading from './Loading'
 import Form from './Form'
 
 import ApiClient from './lib/api'
+import EnterpriseApiClient from './lib/enterpriseApi'
 import { copyToClipboard, createMailtoUrl } from './lib/share-link-actions'
 import tracking from './lib/tracking'
 import * as updaters from './lib/updaters'
@@ -21,10 +22,18 @@ const withGiftFormActions = withActions(
 			protocol: initialProps.apiProtocol,
 			domain: initialProps.apiDomain
 		})
+		const enterpriseApi = new EnterpriseApiClient({
+			protocol: initialProps.apiProtocol,
+			domain: initialProps.enterpriseApiDomain
+		})
 
 		return {
 			showGiftUrlSection() {
 				return updaters.showGiftUrlSection
+			},
+
+			showEnterpriseUrlSection() {
+				return updaters.showGiftEnterpriseSection
 			},
 
 			showNonGiftUrlSection() {
@@ -108,12 +117,20 @@ const withGiftFormActions = withActions(
 						}
 					} else {
 						const { giftCredits, monthlyAllowance, nextRenewalDate } = await api.getGiftArticleAllowance()
-
+						const { limit, hasCredits } = await enterpriseApi.getEnterpriseArticleAllowance()
 						// avoid to use giftCredits >= 0 because it returns true when null and ""
 						if (giftCredits > 0 || giftCredits === 0) {
-							return updaters.setAllowance(giftCredits, monthlyAllowance, nextRenewalDate)
+							return {
+								...updaters.setAllowance(giftCredits, monthlyAllowance, nextRenewalDate),
+								enterpriseLimit: limit,
+								enterpriseHasCredits: hasCredits
+							}
 						} else {
-							return { invalidResponseFromApi: true }
+							return {
+								invalidResponseFromApi: true,
+								enterpriseLimit: limit,
+								enterpriseHasCredits: hasCredits
+							}
 						}
 					}
 				}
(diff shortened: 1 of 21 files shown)


nickramsbottom (approved):
Looks great from our team's perspective 🚀

components/x-gift-article/storybook/with-gift-credits.js
```diff
@@ -38,4 +39,5 @@ exports.fetchMock = (fetchMock) => {
 			redemptionUrl: articleUrlRedeemed,
 			remainingAllowance: 1
 		})
+		.get(`https://enterprise-sharing-api.ft.com/v1/users/me/allowance`, 404)
```
polkwc, line 42: **Question:** What about this kind of endpoint? Do you have a way to use environment variables?

components/x-gift-article/storybook/with-gift-credits.js
```diff
@@ -12,7 +12,8 @@ exports.args = {
 		title: 'Title Title Title Title'
 	},
 	showMobileShareLinks: true,
-	id: 'base-gift-article-static-id'
+	id: 'base-gift-article-static-id',
+	enterpriseApiBaseUrl: `https://enterprise-sharing-api.ft.com`
```
polkwc, line 16: Question: What about this kind of endpoint? Do you have a way to use environment variables?

components/x-gift-article/storybook/with-enterprise.js
```diff
@@ -0,0 +1,52 @@
...
+			redemptionUrl: articleUrlRedeemed,
+			redemptionLimit: 3,
+			remainingAllowance: 1
+		})
+		.get(`https://enterprise-sharing-api.ft.com/v1/users/me/allowance`, {
+			limit: 120,
+			hasCredits: true,
+			firstTimeUser: false
+		})
+		.post(`https://enterprise-sharing-api.ft.com/v1/shares`, {
```
polkwc, line 48: Question: What about this kind of endpoint? Do you have a way to use environment variables?

components/x-gift-article/src/GiftArticle.jsx
```diff
@@ -56,6 +63,17 @@ const withGiftFormActions = withActions(
...
 			},
 
+			async createEnterpriseUrl() {
+				const { redemptionUrl, redemptionLimit } = await enterpriseApi.getESUrl(initialProps.article.id)
+
+				if (redemptionUrl) {
+					tracking.createESLink(redemptionUrl)
+					return updaters.setGiftUrl(redemptionUrl, redemptionLimit, false, true)
+				} else {
+					// TODO do something
```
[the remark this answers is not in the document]
fabiosikansi: If the request fails to create the Gift/Enterprise link it doesn't do anything. Not a good solution but It is an event unlikely to happen (unless we have an outage with the API). The code itself is a copy of the current gift link https://github.com/Financial-Times/x-dash/blob/main/components/x-gift-article/src/GiftArticle.jsx#L46-L57 that also fails silently.
I am happy to change it and show some error message and I do the same with the existing implementation if you agree

components/x-gift-article/storybook/with-gift-credits.js
fabiosikansi: Specifically for this line, it's just mocking the enterprise sharing URL, other fetches mocked in this test only include the endpoint:
```
.get('/article/gift-credits',
```
and
```
.get(`/article/shorten-url/${encodeURIComponent(articleUrlRedeemed)}`, {
```
but it only works because the api is [initialised](https://github.com/Financial-Times/x-dash/blob/main/components/x-gift-article/src/GiftArticle.jsx#L20-L23) with domain empty. I decided to mock the full enterprise sharing URL so it is clear the difference between the gift api and the enterprise-sharing one. Otherwise someone that look at this code later would think that both services are provided by the same API

components/x-gift-article/storybook/with-gift-credits.js
fabiosikansi: My understanding is that x-dash should be a standalone package, therefore not relying in the environment it is used. So the API URL is passed as parameter in the component `<GiftArticle />`. When we implement it in next-article it will get the API from an env var (if agreed by the storytelling team). Worth saying that it is the same approach used in the current implementation, but the url is empty in tests, so not visible.

components/x-gift-article/storybook/with-enterprise.js
fabiosikansi: Like the answer above, I'm mocking the full endpoint for testing instead of leaving the api domain empty, that could cause a misinterpretation that gift and enterprise sharing are served by the same API

components/x-gift-article/src/GiftArticle.jsx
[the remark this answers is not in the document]
fabiosikansi: I've added the error message when the request fails (fc51c7ab4c24101f35f8ad212d673d506bdbac01)
image

polkwc (approved):
It's all good.

binaryberry (approved):

PR #636, opened 2022-01-05
ADSDEV-963 insert custom slots only if related teaser position is found

alevito:
[ADSDEV-963](https://financialtimes.atlassian.net/browse/ADSDEV-963)

A bug was reported where the X-Teaser-Timeline component attempts to insert ad slots even if the specified position is greater than the number of available teasers.

This PR improves the functionality for inserting the custom slots in between teasers in the teasers timeline, by making sure that the custom slot is inserted only when a suitable position is successfully identified.

### Description of the bug
This issue mainly impacted the FT App, and the FT App team added a temporary patch in the PR https://github.com/Financial-Times/ft-app/pull/1729

Below I replicated the error by passing a custom slot position value - 15 - greater than the length of the array of items - 10. The X-Teaser-Timeline should skip the insertion of the out-of-bound slot, while in this case it errors out.
```javascript
TypeError: Cannot read property 'items' of undefined

      136 | 	if (position > 0) {
      137 | 		const group = groups.findIndex((g) => g.items.some((item) => item.articleIndex === position - 1))
    > 138 | 		const index = groups[group].items.findIndex((item) => item.articleIndex === position - 1)
          | 		                            ^
      139 |
      140 | 		return {
      141 | 			group: group,

      at getGroupAndIndex (components/x-teaser-timeline/src/lib/transform.js:138:31)
      at interleaveAllSlotsWithCustomSlots (components/x-teaser-timeline/src/lib/transform.js:160:18)
      at buildModel (components/x-teaser-timeline/src/lib/transform.js:192:16)
      at Object.<anonymous> (components/x-teaser-timeline/__tests__/lib/transform.test.js:350:19)
```
Specifically, the `getGroupAndIndex()` function takes as arguments an array of group of items - each group is basically a list of news that happened on the same date - and it attempts to identify the correct group and article index, based on its linear position - ie., where the custom slot needs to appear within the teaser timeline whole list.
In this case an error happens because the `getGroupAndIndex()` functionality - which is responsible for adding the custom slot to the correct group of items - fails to identify that the provided position is out-of bounds - meaning that the position is either < 0 or >= total items length.

Although not strictly necessary, a deep copy of the grouped items data structure is created to avoid mutating the items in the groups list passed as argument to the `interleaveAllSlotsWithCustomSlots()` function. Creating a deep copy makes the function `interleaveAllSlotsWithCustomSlots()` more easily testable. Let me know what you think about deep copying the structure; if any concern arises I can remove it and change the unit tests accordingly.

### DIFF
--- a/components/x-teaser-timeline/__tests__/lib/transform.test.js
+++ b/components/x-teaser-timeline/__tests__/lib/transform.test.js
@@ -1,4 +1,4 @@
-import { buildModel } from '../../src/lib/transform'
+import { interleaveAllSlotsWithCustomSlots, buildModel } from '../../src/lib/transform'
 
 const items = [
 	{
@@ -372,5 +372,51 @@ describe('buildModel', () => {
 				}
 			])
 		})
+	})
+})
+
+describe('interleaveAllSlotsWithCustomSlots', () => {
+	test('inserts custom slots in group-article indexes', () => {
+		const customSlotContent = [{ foo: 1 }, { bar: 2 }]
+		const customSlotPosition = [0, 10]
+
+		const interleavedGroupedItems = interleaveAllSlotsWithCustomSlots(
+			customSlotContent,
+			customSlotPosition,
+			groupedItems,
+			items
+		)
+
+		expect(interleavedGroupedItems).toEqual([
+			{
+				...groupedItems[0],
+				items: [
+					{ foo: 1 },
+					groupedItems[0].items[0],
+					groupedItems[0].items[1],
+					groupedItems[0].items[2],
+					groupedItems[0].items[3]
+				]
+			},
+			groupedItems[1],
+			{
+				...groupedItems[2],
+				items: [...groupedItems[2].items, { bar: 2 }]
+			}
+		])
+	})
+
+	test('does not insert custom slots in out-of-bound group-article indexes', () => {
+		const customSlotContent = [{ foo: 1 }, { bar: 2 }]
+		const customSlotPosition = [-5, 15]
+
+		const interleavedGroupedItems = interleaveAllSlotsWithCustomSlots(
+			customSlotContent,
+			customSlotPosition,
+			groupedItems,
+			items
+		)
+
+		expect(groupedItems).toEqual(interleavedGroupedItems)
 	})
 })
--- a/components/x-teaser-timeline/src/TeaserTimeline.jsx
+++ b/components/x-teaser-timeline/src/TeaserTimeline.jsx
@@ -6,6 +6,21 @@ import { getDateOnly } from './lib/date'
 import styles from './TeaserTimeline.scss'
 import classNames from 'classnames'
 
+/**
+ * @typedef {Object} Props
+ * @property {string} csrfToken
+ * @property {boolean} showSaveButtons
+ * @property {Array<Object> | Array<string> | Object | string} customSlotContent
+ * @property {Array<number> | number} customSlotPosition
+ * @property {Array<Item>} items
+ * @property {number} timezoneOffset
+ * @property {string} localTodayDate
+ * @property {string} latestItemsTime,
+ * @property {number} latestItemsAgeHours
+ *
+ * @param {Props} props
+ * @returns
+ */
 const TeaserTimeline = (props) => {
 	const now = new Date()
 	const {
--- a/components/x-teaser-timeline/src/lib/transform.js
+++ b/components/x-teaser-timeline/src/lib/transform.js
@@ -1,5 +1,51 @@
 import { getLocalisedISODate, getTitleForItemGroup, getDateOnly } from './date'
 
+/**
+ * @typedef { Array<string> | Array<Object> | Object | string } CustomSlotContent
+ * @typedef { Array<number> | number } CustomSlotPosition
+ */
+
+/**
+ * A news item (an article).
+ * @typedef {Object} Item
+ * @property {string} id e.g., '01f0b004-36b9-11ea-a6d3-9a26f8c3cba4'
+ * @property {string} title e.g.,'Europeans step up pressure on Iran over nuclear deal'
+ * @property {string} publishedDate e.g., '2020-01-14T11:10:26.000Z'
+ */
+
+/**
+ * Extra props added to an item in groups.
+ * @typedef {Object} ItemInGroupInfo
+ * @property {number} articleIndex
+ * @property {string} localisedLastUpdated e.g., '2020-01-14T11:10:26.000+00:00'
+ */
+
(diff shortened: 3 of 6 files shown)


components/x-teaser-timeline/src/lib/transform.js
```diff
@@ -120,52 +187,116 @@ const getItemGroups = ({
...
+ * @property {number} timezoneOffset
+ * @property {string} localTodayDate e.g., '2020-01-14'
+ * @property {string} latestItemsTime e.g., '2020-01-13T10:00:00+00:00'
+ * @property {number} latestItemsAgeHours e.g., 36
+ *
+ * Builds the XTeaserTimeline data model.
+ * The list of news items passed as argument is divided into groups of items,
+ * where each group identifies a specific date when the subset of news was published.
+ * Custom slots - e.g., ads - are also inserted into their expected position.
+ * @param {buildModelProps}
```
oliverturner, line 297:
```suggestion
 * @typedef {Object} BuildModelProps
 * @property {Array<Item>} items
 * @property {CustomSlotContent} customSlotContent
 * @property {CustomSlotPosition} customSlotPosition
 * @property {number} timezoneOffset
 * @property {string} localTodayDate e.g., '2020-01-14'
 * @property {string} latestItemsTime e.g., '2020-01-13T10:00:00+00:00'
 * @property {number} latestItemsAgeHours e.g., 36
 *
 * Builds the XTeaserTimeline data model.
 * The list of news items passed as argument is divided into groups of items,
 * where each group identifies a specific date when the subset of news was published.
 * Custom slots - e.g., ads - are also inserted into their expected position.
 * @param {BuildModelProps} props
```
Small tweak: `@param` definitions need an accompanying reference (here `props`)

components/x-teaser-timeline/src/TeaserTimeline.jsx
```diff
@@ -6,6 +6,21 @@ import { getDateOnly } from './lib/date'
...
+ * @property {Array<Object> | Array<string> | Object | string} customSlotContent
+ * @property {Array<number> | number} customSlotPosition
+ * @property {Array<Item>} items
+ * @property {number} timezoneOffset
+ * @property {string} localTodayDate
+ * @property {string} latestItemsTime,
+ * @property {number} latestItemsAgeHours
+ *
+ * @param {Props} props
+ * @returns
```
oliverturner, line 22:
```suggestion
 * @returns {(props: Props) => JSX.Element}
```

Suggestion [non-blocking]: I'd always recommend explicitly typing the return signature: it guards against refactors inadvertently changing the behaviour of the method

(a classic one is when an arrow function is refactored into a plain function declaration and the developer forgets the `return` keyword)

components/x-teaser-timeline/__tests__/lib/transform.test.js
```diff
@@ -374,3 +374,49 @@ describe('buildModel', () => {
...
+		const interleavedGroupedItems = interleaveAllSlotsWithCustomSlots(
+			customSlotContent,
+			customSlotPosition,
+			groupedItems,
+			items
+		)
+
+		expect(groupedItems).toEqual(interleavedGroupedItems)
+	})
+})
```
oliverturner, line 422: Praise: this is _super_ helpful to understanding the intended behaviour 👍

components/x-teaser-timeline/src/lib/transform.js
```diff
@@ -28,7 +74,12 @@ const groupItemsByLocalisedDate = (
 	}))
 }
 
-const splitLatestItems = (items, localTodayDate, latestItemsTime) => {
+/**
+ * @param {Array<Item>} items
+ * @param {string} latestItemsTime
+ * @returns {Array<Array<ItemInGroup>,Array<Item>>}
```
oliverturner, line 80:
```suggestion
 * @returns {[ItemInGroup[], Item[]]}
```
Suggestion [non-blocking]: It can be helpful to know that the returned value has a fixed number of items rather than an unbounded list

components/x-teaser-timeline/src/lib/transform.js
```diff
@@ -1,5 +1,51 @@
...
+ * @property {string} date e.g., '2020-01-14'
+ * @property {Array<ItemInGroup>} items An array of news articles.
+ */
+
+/**
+ * @typedef {Object} PositionInGroup
+ * @property {number} group
+ * @property {number} index
+ */
+
```
oliverturner, line 40: Praise: I found this really helpful... thank you!

Personally I tend to write  types in a separate `.d.ts` file (mainly because the syntax is nicer!) but it's actually kinda nice to have them right here too 😀

components/x-teaser-timeline/src/lib/transform.js
```diff
@@ -75,18 +142,18 @@ const isLatestNewsSectionAllowed = (localTodayDate, latestItemsTime, latestItems
...
  * Will include a "Latest News" group if allowed by `latestItemsTime` and `latestItemsAgeRange`.
  *
- * @param {Item[]} items  An array of news articles.
+ * @param {Array<Item>} items  An array of news articles.
  * @param {number} timezoneOffset  Minutes ahead (negative) or behind UTC
  * @param {string} localTodayDate  Today's date in client timezone. ISO Date string format.
  * @param {string} latestItemsTime  Cutoff time for items to be treated as "Latest News". ISO Date string format.
  * @param {number} latestItemsAgeRange  Maximum age allowed for items in "Latest News". Hours.
- * @returns An array of group objects, each containing the group's title, date and items.
+ * @returns {Array<GroupOfItems>} An array of group objects, each containing the group's title, date and items.
```
oliverturner, line 156:
```
 * @param {Object} props
 * @param {Item[]} props.items  An array of news articles.
 * @param {number} props.timezoneOffset  Minutes ahead (negative) or behind UTC
 * @param {string} props.localTodayDate  Today's date in client timezone. ISO Date string format.
 * @param {string} props.latestItemsTime  Cutoff time for items to be treated as "Latest News". ISO Date string format.
 * @param {number} props.latestItemsAgeHours  Maximum age allowed for items in "Latest News". Hours.
 * @returns {GroupOfItems[]} An array of group objects, each containing the group's title, date and items.
```
Comment [non-blocking]: as well as defining a method's signature in a type you can also declare it directly using the dot operator for properties

Note: this is just a potentially interesting bit of trivia, not a request for a change!

oliverturner (approved):
Great work, and thank you so much for taking so much effort over the type definitions: made understanding the code so much easier!

On your question about the deep clone... I think that it works! Happy to approve this PR and revisit if it's felt that a different approach would provide any benefits.

oliverturner (approved):

PR #637, opened 2022-01-10
ADSDEV 963/externalise type definitions

oliverturner:
Moves type definitions into a dedicated definitions file


alevito:
Fantastic! Thanks for this @oliverturner!
This component is now well documented with jsdoc and its common types are nicely placed in a separate file. 🙌

PR #641, opened 2022-01-31
fix x-topic-search result font size

apaleslimghost:


serena97 (approved):
right, I see that we're changing the font-size from 12px -> 16px according to the v5 scale size table

fenglish (approved):
Thank you for fixing it!

PR #643, opened 2022-01-31
don't build individual components on install

apaleslimghost:
a) it doesn't work because bootstrapping b) they get built in order after install anyway


apaleslimghost:
and yes, this is completely temporary when we unbower it anyway, just need to get it building for the prerelease components when i merge them which i'll need to do to do the unbowering so

serena97 (approved):

PR #661, opened 2022-03-23
use peer dependencies for origami component css

apaleslimghost:
which entails:

- no longer bundling CSS; consumers are expected to import the SCSS directly
- no longer using CSS modules (since we're no longer bundling CSS), so i've used (sort of) [BEM naming](http://getbem.com/naming/) for classes instead
- importing styles in the storybooks instead of appending `<link>` elements to the compiled CSS
- updating to Build Service v3 and the updating the versions of components included with it
- tidying up a few of the more scuffed ways that the classes were being used (e.g. using `classnames` etc for single classes
- probably some other stuff i've been working on this for weeks, my god

this is going to be a nightmare to review lol. you probably wanna just skim it.

i'm aware i'm not strictly following BEM. i don't much care as long as the class names are specific enough


components/x-gift-article/src/Buttons.jsx
```diff
@@ -20,19 +13,23 @@ export default ({
 	if (isGiftUrlCreated || shareType === ShareType.nonGift) {
 		if (nativeShare) {
 			return (
-				<div className={ButtonsClassName}>
-					<button className={ButtonWithGapClassNames} type="button" onClick={actions.shareByNativeShare}>
+				<div className="x-gift-article__buttons">
+					<button
+						className="js-copy-link x-gift-article__button x-gift-article-button--with-gap"
```
serena97, line 18: question: should this be 'x-gift-article-button--gap' instead? I can't find 'x-gift-article-button--with-gap' in the components/x-gift-article/src/Buttons.scss

components/x-live-blog-post/src/LiveBlogPost.jsx
```diff
@@ -1,91 +1,80 @@
...
 			<div
-				className={`${styles['live-blog-post__body']} n-content-body article--body`}
+				className="x-live-blog-post__body n-content-body article--body"
 				dangerouslySetInnerHTML={{ __html: bodyHTML || content }}
 			/>
-			<div className={styles['live-blog-post__controls']}>
+			<div className="x-live-blog-post__controls">
 				{showShareButtons && <ShareButtons postId={id || postId} articleUrl={articleUrl} title={title} />}
-				{Boolean(BackToTopComponent) && <Fragment>{BackToTopComponent}</Fragment>}
+				<BackToTop backToTop={backToTop} />
```
serena97, line 77: question- do we need to use boolean just in case it shows null in the html?

components/x-teaser-timeline/src/TeaserTimeline.jsx
```diff
@@ -47,18 +45,18 @@ const TeaserTimeline = (props) => {
...
+						<ul className="x-teaser-timeline__items">
 							{group.items.map((item) => {
 								if (item.id) {
 									return (
-										<li key={item.id} className={styles.item}>
+										<li key={item.id} className="x-teaser-timeline__item">
 											<Teaser {...item} {...presets.SmallHeavy} modifiers="timeline-teaser" />
 											{showSaveButtons && (
-												<div className={classNames(styles.itemActions)}>
+												<div className="x-teaser-timeline__item-ctions">
```
serena97, line 59: nitpick: typo

components/x-live-blog-post/src/LiveBlogPost.jsx
apaleslimghost: `return null` from a React component won't output anything

serena97 (approved):
this looks good to me, it's ready to merge once the typo is fixed

PR #664, opened 2022-04-07
Update readme versions of node and npm to match package engines key

IkeLutra:
As per the title, I have updated the readme so it matches the current [engines](https://github.com/Financial-Times/x-dash/blob/main/package.json#L86-L89) key in package.json. This caught me out while setting it up locally. New to the FT so any feedback is most welcome.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -27,8 +27,8 @@ To get started with x-dash, you'll need to make sure you have the following soft
 
 1. [Git](https://git-scm.com/)
 2. [Make](https://www.gnu.org/software/make/)
-3. [Node.js](https://nodejs.org/en/) (version 12)
-4. [npm](http://npmjs.com/)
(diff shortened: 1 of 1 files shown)


ivomurrell:
Hey, this looks good to me, thanks for the catch! The dangers of automating a migration 😅

Unfortunately, our tests are not set up to run for forked versions of the repository at the moment (I've made a note to review that next week.) Seeing as you already have write access to this repository, would you be able to instead create a branch in the upstream `Financial-Times` repository instead of your fork, and remake this PR? Sorry for the hassle though at least you've highlighted an issue which we'll want to look into further if that's any consolation...

IkeLutra:
Thanks @ivomurrell . I thought I didn't have write access but I appear to. Have re-created the PR https://github.com/Financial-Times/x-dash/pull/665

PR #665, opened 2022-04-07
Update readme versions of node and npm to match package engines key

IkeLutra:
Non-fork version of this PR: https://github.com/Financial-Times/x-dash/pull/664

As per the title, I have updated the readme so it matches the current [engines](https://github.com/Financial-Times/x-dash/blob/main/package.json#L86-L89) key in package.json. 

This caught me out while setting it up locally. New to the FT so any feedback is most welcome.


ivomurrell (approved):
Thanks again for sorting this!

ivomurrell:
Merged now, thanks again @IkeLutra

PR #668, opened 2022-07-21
Adds refer a friend feature to gift article

jorbuedo:
### Description
This PR adds an additional section to the gift article component.
The section is an optional Refer a Friend feature with a title, description and copyable url.
The intention is to run it as an AB test on `next-article`. 
I've tried keeping the code as isolated as possible, so it can be removed or improved (dynamic urls) later, while also seemingly integrating it with the existing code.

### Ticket
https://financialtimes.atlassian.net/browse/RS-17

### Design
https://www.figma.com/file/YUc6OErd2vMIQocCrt8n0t/Refer-a-friend-(RAF)?node-id=37%3A514

### Screenshot
Screenshot 2022-07-21 at 14 12 07

### DIFF
--- a/components/x-gift-article/storybook/free-article.js
+++ b/components/x-gift-article/storybook/free-article.js
@@ -3,7 +3,9 @@ const nonGiftArticleUrl = `${articleUrl}?shareType=nongift`
 
 exports.args = {
 	title: 'Share this article (free)',
+	isRafActive: false,
 	isFreeArticle: true,
+	raf: { title: 'Custom title' },
 	article: {
 		title: 'Title Title Title Title',
 		id: 'base-gift-article-static-id',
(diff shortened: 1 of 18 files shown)


components/x-gift-article/storybook/free-article.js
```diff
@@ -3,7 +3,9 @@ const nonGiftArticleUrl = `${articleUrl}?shareType=nongift`
 
 exports.args = {
 	title: 'Share this article (free)',
+	isRafActive: false,
 	isFreeArticle: true,
+	raf: { title: 'Custom title' },
```
ludrob, line 8: if `isRafActive: false,`
`raf: { title: 'Custom title' },` would do nothing?

Should it be:
if `isRafActive: true,`
`raf: { title: 'Custom title' },`

ludrob (commented):
Shall we have a storybook entry for this component?

components/x-gift-article/storybook/free-article.js
jorbuedo: Args can be changed in storybook, just by having it here as false, it can be toggled later.

jorbuedo:
I could add an entry in storybook with isRafActive: true, but basically any of the current ones allows you to toggle it on. And since this can be added to any of the other variants, that's why I didn't create a specific entry for it.

ludrob (approved):

PR #676, opened 2022-10-27
[CI-1368] fix accessibility errors

AlbertoCuberoNavas:
After a user activates the share option, and chooses the ‘FT subscribers only’ option in the modal, they are presented with a form field containing the link for the article. However, the form field has not been provided with a visual label to inform users of the purpose of the field.
This also means that voice activation users are unable to discern what voice command is required to access the form field.
Users were not presented instructions or labels to identify the controls in a form and describe what input data is expected. Those with cognitive, language, and learning disabilities may find labels help them to enter information correctly. Labels may also help to prevent users from making submission errors. For voice activation users, visible labels can be used to indicate the accessible name by which this component may be referenced to assistive technology. Otherwise, users are required to use general commands such as ‘click box’. 

[ticket](https://financialtimes.atlassian.net/browse/CI-1368)

BEFORE
image

AFTER
image

### DIFF
--- a/components/x-gift-article/src/ReferAFriend.jsx
+++ b/components/x-gift-article/src/ReferAFriend.jsx
@@ -12,7 +12,9 @@ export default ({ rafTitle, rafDescription, urls, actions }) => {
 				data-trackable={UrlType.raf + 'Link'}
 			>
 				<span className="o-forms-input o-forms-input--text">
-					<label for="refer-friend-link">{rafTitle}</label>
+					<label className="x-gift-article__label-link" htmlFor="refer-friend-link">
+						Gift 2 month free access link
+					</label>
 					<input
 						id="refer-friend-link"
 						type="text"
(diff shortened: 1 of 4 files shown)


components/x-gift-article/src/ReferAFriend.jsx
```diff
@@ -12,13 +12,14 @@ export default ({ rafTitle, rafDescription, urls, actions }) => {
 				data-trackable={UrlType.raf + 'Link'}
 			>
 				<span className="o-forms-input o-forms-input--text">
+					<label for="refer-friend-link">{rafTitle}</label>
```
juanSanchezAlcala, line 15: This will add the same text twice, isn't it ? I think this is not the final solution due as i understand the view has changed and the content of rafTitle is shown twice.

juanSanchezAlcala (changes requested):

components/x-gift-article/src/ReferAFriend.jsx
juanSanchezAlcala: maybe in this case this is not an accessibility issue and we need to discuss with they

asugar13 (commented):
I am ready to approve but just have one request. Could you please include a screenshot or a video capture of the frontend changes in the PR description?

asugar13 (approved):
looks good thank you for this

PR #680, opened 2022-11-09
fix: handle live blog posts without a byline

adgad:
Turns out `typeof null === 'object'`, so this would throw an error when that happens

### DIFF
--- a/components/x-live-blog-post/src/LiveBlogPost.jsx
+++ b/components/x-live-blog-post/src/LiveBlogPost.jsx
@@ -70,7 +70,7 @@ const LiveBlogPost = ({
 			/>
 		)
 	}
-	if (typeof byline === 'object' && 'tree' in byline) {
+	if (byline && typeof byline === 'object' && 'tree' in byline) {
 		postByline = (
 			<p className="x-live-blog-post__byline">
 				<RichText structuredContent={byline} />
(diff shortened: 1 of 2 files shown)


nickramsbottom (approved):

components/x-live-blog-post/src/LiveBlogPost.jsx
```diff
@@ -70,7 +70,7 @@ const LiveBlogPost = ({
 			/>
 		)
 	}
-	if (typeof byline === 'object' && 'tree' in byline) {
+	if (byline && typeof byline === 'object' && 'tree' in byline) {
```
apaleslimghost, line 73: **thought**: this suggests our types aren't up to scratch if typescript didn't mind this 😬

adgad: i don't think there are any types in x-dash?

apaleslimghost: frick i thought i was in the other repo

adgad: although curiously I just tried it in our repo and it isn't caught there either.
```
(property) byline?: {
    readonly __typename?: "StructuredContent" | undefined;
    readonly tree: any;
} | undefined
```

That's the generated type, but the data we're getting from the API is:
```
        {
          "byline": null
        }
```

apaleslimghost: frick

adgad: ah from our codegen config:

```
# make types better-compatible with React, which doesn't like null
      maybeValue: T
```

apaleslimghost: 😶

PR #682, opened 2022-11-30
fix: improve x-engine server error message

apaleslimghost:
ht @nickramsbottom for pointing out the current unhelpful message


nickramsbottom (approved):
beautiful

joelcarr (approved):

PR #684, opened 2022-12-14
ENTST-139 Add Enterprise Sharing to FTR Articles

willhoward:
This PR tweaks the behaviour of `x-gift-article` to make a few adjustments concerning Enterprise Sharing. The main difference is that when sharing a free to read article with Enterprise Sharing enabled, users will now see the option to share with Enterprise Sharing alongside the default share option.

This is necessary because articles commonly remain free to read for a short period of time before going behind the paywall and we've had feedback from several Enterprise Sharing clients expressing frustration that the standard links to the article they've shared stop working when this happens. Enterprise Sharing clients can also track how many times their links have been opened and therefore some would prefer to share free to read articles via Enterprise Sharing to have visibility over this.

This behaviour is covered by ticket [ENTST-139](https://financialtimes.atlassian.net/browse/ENTST-139?atlOrigin=eyJpIjoiYTM0ODA1NGMzM2I5NGFkMmI3YjU1NDdlZDY0YTdiM2MiLCJwIjoiaiJ9).

I've added a 'Free with enterprise sharing' option to the Storybook to make this behaviour available to test.

### Styling Tweaks

This PR also makes a few minor styling changes to the `x-gift-article`:

- The 'Enterprise' label has been removed from next to the 'Enterprise Sharing' radio button. Ticket [ENTST-152](https://financialtimes.atlassian.net/browse/ENTST-152?atlOrigin=eyJpIjoiNmEwOThiYzA1NzU1NDJkODkzNjk4ZGJmYjc3NjdkOWEiLCJwIjoiaiJ9).
- The sharing link url field label has been corrected to read 'Enterprise Sharing link' instead of 'Gift article shareable link' when the link has been generated with Enterprise Sharing. Ticket [ENTST-154](https://financialtimes.atlassian.net/browse/ENTST-154?atlOrigin=eyJpIjoiNjhlZTBkOTI0YWEyNDUzM2JmOGI5NzVmZDRkYjRmYjMiLCJwIjoiaiJ9).
- A 'View all Enterprise Links' link which takes users to the [Enterprise Sharing Dashboard](https://enterprise-sharing-dashboard.ft.com/) has been added to the messaging when the Enterprise Sharing option is selected. Ticket [ENTST-135](https://financialtimes.atlassian.net/jira/software/projects/ENTST/boards/1569?selectedIssue=ENTST-135).

[ENTST-139]: https://financialtimes.atlassian.net/browse/ENTST-139?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ENTST-139]: https://financialtimes.atlassian.net/browse/ENTST-139?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ENTST-152]: https://financialtimes.atlassian.net/browse/ENTST-152?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/x-gift-article/src/Form.jsx
+++ b/components/x-gift-article/src/Form.jsx
@@ -12,20 +12,19 @@ export default (props) => (
 			<div role="group" arialabelledby="gift-article-title">
 				<Title {...props} />
 
-				{!props.isFreeArticle && (
-					<RadioButtonsSection
-						shareType={props.shareType}
-						isArticleSharingUxUpdates={props.isArticleSharingUxUpdates}
-						showGiftUrlSection={props.actions.showGiftUrlSection}
-						showEnterpriseUrlSection={props.actions.showEnterpriseUrlSection}
-						showNonGiftUrlSection={props.actions.showNonGiftUrlSection}
-						enterpriseLimit={props.enterpriseLimit}
-						enterpriseHasCredits={props.enterpriseHasCredits}
-						enterpriseRequestAccess={props.enterpriseRequestAccess}
-						enterpriseAlert={!props.enterpriseHasCredits && !props.enterpriseRequestAccess}
-						enterpriseEnabled={props.enterpriseEnabled}
-					/>
-				)}
+				<RadioButtonsSection
+					shareType={props.shareType}
+					isArticleSharingUxUpdates={props.isArticleSharingUxUpdates}
+					showGiftUrlSection={props.actions.showGiftUrlSection}
+					showEnterpriseUrlSection={props.actions.showEnterpriseUrlSection}
+					showNonGiftUrlSection={props.actions.showNonGiftUrlSection}
+					enterpriseLimit={props.enterpriseLimit}
+					enterpriseHasCredits={props.enterpriseHasCredits}
+					enterpriseRequestAccess={props.enterpriseRequestAccess}
+					enterpriseAlert={!props.enterpriseHasCredits && !props.enterpriseRequestAccess}
+					enterpriseEnabled={props.enterpriseEnabled}
+					isFreeArticle={props.isFreeArticle}
+				/>
 
 				<UrlSection {...props} />
 			</div>
--- a/components/x-gift-article/src/GiftArticle.jsx
+++ b/components/x-gift-article/src/GiftArticle.jsx
@@ -150,27 +150,38 @@ const withGiftFormActions = withActions(
 
 			activate() {
 				return async (state) => {
+					const { enabled, limit, hasCredits, firstTimeUser, requestAccess } =
+						await enterpriseApi.getEnterpriseArticleAllowance()
+
+					if (enabled) {
+						tracking.initEnterpriseSharing(
+							requestAccess
+								? 'enterprise-request-access'
+								: !hasCredits
+								? 'enterprise-no-credits'
+								: 'enterprise-enabled'
+						)
+					} else {
+						tracking.initEnterpriseSharing('enterprise-disabled')
+					}
+
 					if (initialProps.isFreeArticle) {
 						const { url, isShortened } = await api.getShorterUrl(state.urls.nonGift)
 
 						if (isShortened) {
-							return updaters.setShortenedNonGiftUrl(url)(state)
+							updaters.setShortenedNonGiftUrl(url)(state)
+						}
+						return {
+							invalidResponseFromApi: true,
+							enterpriseEnabled: enabled,
+							enterpriseLimit: limit,
+							enterpriseHasCredits: hasCredits,
+							enterpriseFirstTimeUser: firstTimeUser,
+							enterpriseRequestAccess: requestAccess
 						}
 					} else {
 						const { giftCredits, monthlyAllowance, nextRenewalDate } = await api.getGiftArticleAllowance()
-						const { enabled, limit, hasCredits, firstTimeUser, requestAccess } =
-							await enterpriseApi.getEnterpriseArticleAllowance()
-						if (enabled) {
-							tracking.initEnterpriseSharing(
-								requestAccess
-									? 'enterprise-request-access'
-									: !hasCredits
-									? 'enterprise-no-credits'
-									: 'enterprise-enabled'
-							)
-						} else {
-							tracking.initEnterpriseSharing('enterprise-disabled')
-						}
+
 						// avoid to use giftCredits >= 0 because it returns true when null and ""
 						if (giftCredits > 0 || giftCredits === 0) {
 							return {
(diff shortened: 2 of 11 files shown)


AnnaDraganova (approved):

fabiosikansi (changes requested):
I think there are some undesired side-effects in these changes:

1. Button text has now the same colour as background
2. I'd also suggest add a spacing between the label and the warning icon
image

components/x-gift-article/src/RadioButtonsSection.jsx
```diff
@@ -63,5 +57,54 @@ export default ({
...
+		if (enterpriseEnabled) {
+			return (
+				<div
+					className="o-forms-input o-forms-input--radio-round o-forms-field x-gift-article__radio_buttons"
+					role="group"
+					aria-labelledby="article-share-options"
+				>
+					<span className="x-gift-article--visually-hidden" id="article-share-options">
+						Article share options
+					</span>
```
umbobabo, line 86: Suggestion:
All this part is duplicated in [line97](https://github.com/Financial-Times/x-dash/pull/684/files#diff-4b3d233b646176226bcb2b4b467761b8775262f3b727dfdf7254d39b33b44c08R97). You can structure the composition in a way that avoids duplications.

Duplications, in this case, will just make it hard to refactor this component where needed.

components/x-gift-article/src/Form.jsx
```diff
@@ -12,20 +12,19 @@ export default (props) => (
...
-						showNonGiftUrlSection={props.actions.showNonGiftUrlSection}
-						enterpriseLimit={props.enterpriseLimit}
-						enterpriseHasCredits={props.enterpriseHasCredits}
-						enterpriseRequestAccess={props.enterpriseRequestAccess}
-						enterpriseAlert={!props.enterpriseHasCredits && !props.enterpriseRequestAccess}
-						enterpriseEnabled={props.enterpriseEnabled}
-					/>
-				)}
+				<RadioButtonsSection
+					shareType={props.shareType}
```
umbobabo, line 16: (Out-of-scope for this PR). The usage of props without PropTypes will easily lead to zombie properties on refactoring like this. Better would be to explicitly declare all the props as function parameters as done in the components.

components/x-gift-article/src/GiftArticle.jsx
```diff
@@ -150,27 +150,38 @@ const withGiftFormActions = withActions(
...
-							return updaters.setShortenedNonGiftUrl(url)(state)
+							updaters.setShortenedNonGiftUrl(url)(state)
+						}
+						return {
+							invalidResponseFromApi: true,
+							enterpriseEnabled: enabled,
+							enterpriseLimit: limit,
+							enterpriseHasCredits: hasCredits,
+							enterpriseFirstTimeUser: firstTimeUser,
+							enterpriseRequestAccess: requestAccess
```
umbobabo, line 180: Thought:
This seems to be the same return we have [here](https://github.com/Financial-Times/x-dash/pull/684/files#diff-7082a9e3cb317538577308da954ff347e6c58e103516faaa15ec7c405cba1161R198). If they need to stay in sync it is probably better to have a CONST object to return in both cases. You can name it as a significant state.

components/x-gift-article/src/GiftArticle.scss
```diff
@@ -50,12 +50,6 @@
 	);
 }
 
-.x-gift-article__enterprise-label {
-	background-color: oColorsByName('black-10');
-	color: oColorsByName('black');
-	margin: 0px 6px;
-}
```
umbobabo, line 57: Thought: Can this change be related to the bug Fabio reported?

components/x-gift-article/src/Message.jsx
```diff
@@ -151,24 +153,31 @@ export default ({
...
 					<p>
-						Request more credits and our Enterprise team will get in touch with the admin of your FT
-						subscription to arrange a top-up of sharing credits.
+						Request more credits and our team will get in touch with the admin of your FT subscription to
+						arrange a top-up of sharing credits.
 					</p>
 					<a
-						href="https://enterprise.ft.com/ft-enterprise-sharing-request-access/?segmentId=c87259e0-7073-3ea8-7f83-b988f05c3f94"
-						target="_blank"
+						href="mailto:customer.success@ft.com"
```
umbobabo, line 160: Thought: I would rather not use an e-mail directly. However, this part should be visible only to B2B users, is that right?

umbobabo (commented):
There are a lot of changes in this PR, no tests, some repetitions in the code, and some components do not explicitly declare the properties they need. Some of these is anyway related to already existing code. It seems there is a wide margin for improvement of the existing code to make it more robust and easier to refactor in the future. 
I haven't spotted any functional issues but there are some comments.

components/x-gift-article/src/Message.jsx
willhoward: The requirement of the ticket in this case is to include a `mailto` link directly ([ENTST-153](https://financialtimes.atlassian.net/jira/software/projects/ENTST/boards/1569?selectedIssue=ENTST-153)) but yes, this button will only be visible to B2B users and only when they have run out of sharing credits

components/x-gift-article/src/GiftArticle.scss
willhoward: Not quite! This one's [to remove an old label](https://financialtimes.atlassian.net/jira/software/projects/ENTST/boards/1569?selectedIssue=ENTST-152). I've added additional styles to fix the styling bug 😄

components/x-gift-article/src/Form.jsx
willhoward: Agreed. This should be applied repo-wide though!

willhoward:
> I think there are some undesired side-effects in these changes:
> 
> 1. Button text has now the same colour as background
> 2. I'd also suggest add a spacing between the label and the warning icon
>    image

This is now fixed

components/x-gift-article/src/GiftArticle.jsx
willhoward: I'm not quite sure I follow - which bit is the repeated return?

umbobabo: What we are returning seems identical to the part [here](https://github.com/Financial-Times/x-dash/pull/684/files#diff-7082a9e3cb317538577308da954ff347e6c58e103516faaa15ec7c405cba1161R197)

fabiosikansi (approved):

PR #688, opened 2023-01-17
Prevent circular dependency with cp-content-pipeline-ui

apaleslimghost:
This should never have worked but sort of does because of some npm workspace magic I don't understand. To prevent us having to import `cp-content-pipeline-ui`, have consumers pass the `RichText` component in as a prop. This also means we don't have to update this component every time `cp-content-pipeline-ui` is updated.

This is a breaking change.

### DIFF
--- a/components/x-live-blog-post/package.json
+++ b/components/x-live-blog-post/package.json
@@ -17,7 +17,6 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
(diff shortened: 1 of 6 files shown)


adgad:
Ooh sorry I just thought of something - how will this work with the client-side rendering? IIRC I was having trouble getting the component prop to be passed to (i wanna say) x-interaction when it rerenders client side...

PR #693, opened 2023-01-25
Bugfix: Gift Article Radio Buttons

willhoward:
This PR fixes a bug whereby B2C users were erroneously seeing the 'Enterprise Sharing' option in the Gift Article radio buttons. This was happening because the specific logic controlling which radio buttons to display was quite confusing.

I've now separated out that logic. There's some minor repetition of code which I think is justified by making the logic clearer and easier to understand. I've also added a few code comment to explain it.


asugar13 (changes requested):
Can you please share screenshots of the differences before/after? Other than that, ready to approve

asugar13 (approved):
looks good! thank you

PR #696, opened 2023-02-28
Update x-follow-button to support inverse monochrome

phil-hunt-ft:
The inverse-monochrome theme has been added to the follow button in n-myft-ui - this PR updates the x-follow-button to support this


magsallen (approved):
Looks good, thanks for adding this @phil-hunt-ft

PR #697, opened 2023-03-03
CPP-1250: Migrate *mostly* to npm workspaces

joelcarr:
Removing the use of [Athloi](https://github.com/Financial-Times/athloi) for managing tasks in each package of this monorepo and replacing it with [npm workspaces](https://docs.npmjs.com/cli/v8/using-npm/workspaces). This allows us to have a `package-lock.json` for the root of the project and makes installing dependencies for all packages a lot faster. This will allow us to migrate this repo to Tool Kit when it supports monorepos.

This PR removes any reference to the `tools/` directory. It was removed a few years ago in https://github.com/Financial-Times/x-dash/commit/bf3342971bb1dc001856a4c8c3f5cd9388f46101.

### Update

A refactor for how we resolve webpack for the `x-engine` webpack plugin was needed to be able to run the e2e tests. This refactor should aline how npm expects to resolve dependencies whether it is linked to a local version of x-engine or downloaded as a module dependency.


.circleci/config.yml
```diff
@@ -133,7 +133,7 @@ jobs:
           command: npx athloi version ${CIRCLE_TAG}
       - run:
           name: NPM publish
-          command: npx athloi publish -- --access=public
+          command: npm publish --workspaces
```
apaleslimghost, line 136: **question**: will this have the same problem re: `file:` dependencies that Page Kit had?

apaleslimghost (approved):
nice! one question but looks good

joelcarr:
There is unfortunately a bug with how `packages/x-engine/src/webpack.js` works. It tries to resolve the dependency `webpack` by getting the version of `webpack` [from the current working directory](https://github.com/Financial-Times/x-dash/blob/a4d6d145e6a6bb314373b56ebad6442356e724b3/packages/x-engine/src/concerns/resolve-peer.js#L2). But now workspaces is handling all the dependencies, it doesn't install a executable webpack to `e2e/node_modules` so it [cannot be found](https://app.circleci.com/pipelines/github/Financial-Times/x-dash/4456/workflows/41b456e6-4390-4cb0-99b5-9f80be92b75c/jobs/8571?invite=true#step-102-7). I think we can now rely on setting webpack as a peer dependency for `@financial-times/x-engine` and removing the need for `resolvePeer`. Does this sound reasonable?

packages/x-engine/src/webpack.js
```diff
@@ -18,25 +16,10 @@ module.exports = function () {
 	// 3. format the configuration we've loaded
 	const config = formatConfig(raw)
 
-	// 4. if this module is a linked dependency then resolve Webpack & runtime to CWD
-	const webpackResolution = resolvePeer('webpack')
-	const runtimeResolution = resolvePeer(config.runtime)
-	const renderResolution = resolvePeer(config.renderModule)
-
-	const webpack = require(webpackResolution)
+	const webpack = require('webpack')
```
apaleslimghost, line 19: this is probably fine, but it will need testing with a real app

apaleslimghost (approved):

joelcarr:
I have tested this locally with [next-myft-page](https://github.comFinancial-Times/next-myft-page) and everything seems to be working as expected. I did `make build` and `make run` and looked through each link of the myft subnavigation.

packages/x-engine/package.json
```diff
@@ -32,5 +29,11 @@
   },
   "devDependencies": {
     "check-engine": "^1.10.1"
+  },
+  "peerDependencies": {
+    "webpack": "^4.42.0"
```
ivomurrell, line 34: **question**: is there a reason why `x-engine` can only support `webpack@4.42.0` and up? you tend to want peer dependencies to be as lenient as possible so you don't cause npm installations to fail.
```suggestion
    "webpack": "4.x"
```

.storybook/webpack.config.js
```diff
@@ -1,45 +1,11 @@
...
-		{
-			test: /\/components\/x-[^\/]+\/src\//,
-			...xBabelConfig()
-		}
-	]
-
-	// HACK: there is a bug in babel-plugin-minify-simplify which cannot
-	// handle how Babel transpiles restful destructing assignment so remove it.
-	// e.g. const { foo, ...qux } = { foo: 0, bar: 1, baz: 2 }
-	babelConfig.options.presets = babelConfig.options.presets.filter((preset) => {
```
ivomurrell, line 38: **question**: why were hacks like this removed? I can't immediately see the relation between this polyfill issue and (now resolved) monorepo duplication issues

ivomurrell (approved):
got some whiplash after going through all your meticulous commits and seeing how small the final diff actually is 😄 this looks good and makes sense to me (though all the dependency resolution hacks that were previously in place are a little beyond me) just a few non-blocking clarifying q's!

.storybook/webpack.config.js
joelcarr: The hacks that were not related to solve the monorepo duplication issue seemed redundant. As far as I can tell, we do not use `@babel/preset-minify` anymore (tested by running `npm why @babel/preset-minify` and getting a not found error). I would expect removing this hack would break storybook, the tests, and the link test with `next-myft-page` however no of these broke.

What do you think, should I add it back in?

packages/x-engine/src/webpack.js
apaleslimghost: @joelcarr did you get chance to test this?

package.json
```diff
@@ -7,17 +7,17 @@
...
   "scripts": {
     "clean": "git clean -fxdi",
-    "build": "athloi run build --concurrency 3 --preserve-order",
-    "build-only": "athloi run build",
+    "build": "npm run build --workspaces --if-present",
     "jest": "jest -c jest.config.js",
+    "update-snapshots": "jest -c jest.config.js --updateSnapshot",
     "test": "npm run lint && npm run jest",
-    "e2e": "cd e2e && ./node_modules/.bin/webpack && jest -c jest.e2e.config.js",
+    "e2e": "npm test -w e2e",
```
apaleslimghost, line 14: **praise**: this is so much cleaner!

apaleslimghost (approved):
very nice!

jkerr321 (approved):
🤖 YAY 🥳

PR #709, opened 2023-05-30
X-Gift article highlights sharing Release

MMSaleh-FT:
## Description
This PR introduces two features
- Highlights checkbox that creates a share with highlights included
- A new message is introduced to B2B recipients in the x-gift-article prompting the users to save their highlights in order to re-share the article with the received highlights.

Changes:

- Updated the entrepriseApi
- Created a separate highlightsApi client to be used only for requests to the user-annotations-api
- Created HighlightsSection component to be a wrapper of the highlights checkbox, the recipient message for saving highlights and the success message afterwards.
- Created a HighlightCheckbox component, which will appear after the user has saved the shared highlights or when the user has created highlights on their own.
- Created a new message HighlightRecipientMessage which will be shown to B2B users only if they open a shared article with highlights( highlights query parameter should be in the url)
- Created a new success message HighlightSuccessMessage which will appear after the user clicks save the highlights.
- add an action to check if user have highlights ,this action will be used by the next-article project to update the hasHighlights status when user creates highlights after the x-gift-article component initializes 
- remove organization share credit message
- update storybook to show share highlights options

## Jira Tickets 

Highlights checkbox ticket
https://financialtimes.atlassian.net/browse/ENTST-352

Recipient message ticket
https://financialtimes.atlassian.net/browse/ENTST-397

### Sharing article with existing highlights
https://github.com/Financial-Times/x-dash/assets/22678655/b17b8ca1-494c-4ab1-b18a-b7fb4353cbcc

### Sharing article with no highlights
Screenshot 2023-05-04 at 15 28 14

### Sharing article with received highlights

https://github.com/Financial-Times/x-dash/assets/22678655/539e4301-7fb2-4487-9a01-50d71b142882

### DIFF
--- a/components/x-gift-article/package.json
+++ b/components/x-gift-article/package.json
@@ -28,7 +28,7 @@
     "@financial-times/o-loading": "^5.2.1",
     "@financial-times/o-message": "^5.2.1",
     "@financial-times/o-normalise": "^3.2.2",
-    "@financial-times/o-typography": "^7.2.2",
+    "@financial-times/o-typography": "^7.4.1",
     "@financial-times/x-rollup": "file:../../packages/x-rollup",
     "check-engine": "^1.10.1",
     "sass": "^1.49.0"
@@ -41,12 +41,14 @@
     "extends": "../../package.json"
   },
   "peerDependencies": {
+    "@financial-times/o-banner": "^4.4.9",
     "@financial-times/o-buttons": "^7.2.2",
+    "@financial-times/o-colors": "^6.6.0",
     "@financial-times/o-forms": "^9.2.3",
     "@financial-times/o-labels": "^6.2.2",
     "@financial-times/o-loading": "^5.2.1",
     "@financial-times/o-message": "^5.2.1",
     "@financial-times/o-normalise": "^3.2.2",
-    "@financial-times/o-typography": "^7.2.2"
+    "@financial-times/o-typography": "^7.4.1"
   }
 }
--- a/components/x-gift-article/src/HighlightRecipientMessage.jsx
+++ b/components/x-gift-article/src/HighlightRecipientMessage.jsx
@@ -1,0 +1,26 @@
+import { h } from '@financial-times/x-engine'
+
+export default ({ handleSaveAnnotations, handleCloseRecipientMessage }) => {
+	return (
+		<div className="x-gift-article__highlight-recipient-message">
+			<p className="x-gift-article__highlight-recipient-message__title">Share the highlighted article</p>
+			<div className="x-gift-article__highlight-recipient-message__divider"></div>
+			<p className="o-typography-professional o-typography-inverse x-gift-article__highlight-recipient-message__description">
+				If you wish to share the highlighted article,
+				<button
+					type="button"
+					className="o-typography-link x-gift-article__highlight-recipient-message__save-highlights-button"
+					onClick={handleSaveAnnotations}
+				>
+					save the highlights
+				</button>
+				as your own before creating the link.
+			</p>
+			<button
+				className="x-gift-article__highlight-recipient-message__close"
+				type="button"
+				onClick={handleCloseRecipientMessage}
+			></button>
+		</div>
+	)
+}
--- a/components/x-gift-article/src/HighlightRecipientMessage.jsx
+++ b/components/x-gift-article/src/HighlightRecipientMessage.jsx
@@ -19,6 +19,7 @@ export default ({ handleSaveAnnotations, handleCloseRecipientMessage }) => {
 			<button
 				className="x-gift-article__highlight-recipient-message__close"
 				type="button"
+				content="Close"
 				onClick={handleCloseRecipientMessage}
 			></button>
 		</div>
--- a/components/x-gift-article/src/HighlightRecipientMessage.jsx
+++ b/components/x-gift-article/src/HighlightRecipientMessage.jsx
@@ -6,14 +6,14 @@ export default ({ handleSaveAnnotations, handleCloseRecipientMessage }) => {
 			<p className="x-gift-article__highlight-recipient-message__title">Share the highlighted article</p>
 			<div className="x-gift-article__highlight-recipient-message__divider"></div>
 			<p className="o-typography-professional o-typography-inverse x-gift-article__highlight-recipient-message__description">
-				If you wish to share the highlighted article,
+				If you wish to share the highlighted article,{' '}
 				<button
 					type="button"
 					className="o-typography-link x-gift-article__highlight-recipient-message__save-highlights-button"
 					onClick={handleSaveAnnotations}
 				>
 					save the highlights
-				</button>
+				</button>{' '}
 				as your own before creating the link.
 			</p>
 			<button
--- a/components/x-gift-article/src/HighlightRecipientMessage.jsx
+++ b/components/x-gift-article/src/HighlightRecipientMessage.jsx
@@ -2,14 +2,14 @@ import { h } from '@financial-times/x-engine'
 
 export default ({ handleSaveAnnotations, handleCloseRecipientMessage }) => {
 	return (
-		<div className="x-gift-article__highlight-recipient-message">
-			<p className="x-gift-article__highlight-recipient-message__title">Share the highlighted article</p>
-			<div className="x-gift-article__highlight-recipient-message__divider"></div>
-			<p className="o-typography-professional o-typography-inverse x-gift-article__highlight-recipient-message__description">
+		<div className="x-gift-article-highlight-recipient-message">
+			<p className="x-gift-article-highlight-recipient-message__title">Share the highlighted article</p>
+			<div className="x-gift-article-highlight-recipient-message__divider"></div>
+			<p className="o-typography-professional o-typography-inverse x-gift-article-highlight-recipient-message__description">
 				If you wish to share the highlighted article,{' '}
 				<button
 					type="button"
-					className="o-typography-link x-gift-article__highlight-recipient-message__save-highlights-button"
+					className="o-typography-link x-gift-article-highlight-recipient-message__save-highlights-button"
 					onClick={handleSaveAnnotations}
 				>
 					save the highlights
@@ -17,11 +17,12 @@ export default ({ handleSaveAnnotations, handleCloseRecipientMessage }) => {
 				as your own before creating the link.
 			</p>
 			<button
-				className="x-gift-article__highlight-recipient-message__close"
+				className="x-gift-article-highlight-recipient-message__close"
 				type="button"
-				content="Close"
 				onClick={handleCloseRecipientMessage}
-			></button>
+			>
+				<span className="o-normalise-visually-hidden">Close</span>
+			</button>
 		</div>
 	)
 }
--- a/components/x-gift-article/src/HighlightRecipientMessage.scss
+++ b/components/x-gift-article/src/HighlightRecipientMessage.scss
@@ -53,6 +53,7 @@
     .x-gift-article__highlight-recipient-message__save-highlights-button {
       background: none;
       border: none;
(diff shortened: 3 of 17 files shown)


components/x-gift-article/package.json
```diff
@@ -28,7 +28,7 @@
     "@financial-times/o-loading": "^5.2.1",
     "@financial-times/o-message": "^5.2.1",
     "@financial-times/o-normalise": "^3.2.2",
-    "@financial-times/o-typography": "^7.2.2",
+    "@financial-times/o-typography": "^7.4.1",
```
asugar13, line 31: **Non-blocking comment**: I believe the `^` already brings you to the latest major version

components/x-gift-article/src/lib/highlightsApi.js
```diff
@@ -0,0 +1,56 @@
...
+				credentials: 'include',
+				headers: {
+					'Content-Type': 'application/json'
+				}
+			},
+			additionalOptions
+		)
+
+		const response = await fetch(url, options)
+		return await response.json()
```
asugar13, line 39: **Non-blocking comment**: I'd prefer to wrap `await response.json()` in a variable and then return it.
```
const output = await response.json()
return output
```

components/x-gift-article/src/lib/highlightsApi.js
```diff
@@ -0,0 +1,56 @@
...
+		try {
+			const json = await this.fetchJson('/copy-annotations', {
+				method: 'POST',
+				body: JSON.stringify({ highlightsToken })
+			})
+
+			return {
+				annotationsCopyResult: json.annotationsCopyResult
+			}
+		} catch (e) {
```
asugar13, line 52: **Suggestion**: Why not console.log the error here?

asugar13 (approved):
looks good to me. feel free to action my suggestions.

components/x-gift-article/src/lib/highlightsApi.js
MMSaleh-FT: I didn't want to log errors to the client side console

components/x-gift-article/package.json
MMSaleh-FT: yeah it shouldn't make a deference but just for the sake of clarification since the new version contains the new professional branding

components/x-gift-article/src/lib/highlightsApi.js
MMSaleh-FT: we can do that but i don't think that adding a new variable for the sake of just returning it's value is best , since this code does the same without adding it

components/x-gift-article/src/HighlightRecipientMessage.scss
```diff
@@ -0,0 +1,60 @@
...
+
+@include oTypography();
+
+.x-gift-article__highlight-recipient-message {
+    position: relative;;
+    background-color: oColorsByName('slate');
+    color: oColorsMix('ft-grey', 'white', 20);
+    padding: oSpacingByName('s3');
+  
+    .x-gift-article__highlight-recipient-message__title {
```
GlynnPhillips, line 13: **Change request:** The classnames in this file look a bit like [BEM](https://getbem.com/naming/) but seems to be suggesting multiple elements levels. I think it might be useful to refactor this to follow the BEM naming

components/x-gift-article/src/HighlightRecipientMessage.scss
```diff
@@ -0,0 +1,60 @@
+@import '@financial-times/o-colors/main';
+@import '@financial-times/o-spacing/main';
+@import '@financial-times/o-typography/main';
+
+@include oTypography();
+
+.x-gift-article__highlight-recipient-message {
+    position: relative;;
```
GlynnPhillips, line 8: **Change request:** 

```suggestion
    position: relative;
```

components/x-gift-article/src/HighlightRecipientMessage.scss
```diff
@@ -0,0 +1,60 @@
...
+    
+    .x-gift-article__highlight-recipient-message__divider{
+      padding-top: oSpacingByName('s2');
+      padding-bottom: oSpacingByName('s3');
+      &::after {
+        border-top: oSpacingByName('s1') solid oColorsByName('mint');
+        content: "";
+        position: absolute;
+        width: 40px;
+        content: "";
```
GlynnPhillips, line 32: **Change request:** Duplicate property

```suggestion

```

components/x-gift-article/src/HighlightRecipientMessage.scss
```diff
@@ -0,0 +1,60 @@
...
+      margin: 0;
+      color: oColorsByName('white');
+    }
+  
+    .x-gift-article__highlight-recipient-message__description {
+      @include oTypographySans($scale: 0);
+      margin: 0;
+    }
+    
+    .x-gift-article__highlight-recipient-message__divider{
```
GlynnPhillips, line 24: **Change Request:** Please add a space between the selector and `{`

```suggestion
    .x-gift-article__highlight-recipient-message__divider {
```

components/x-gift-article/src/HighlightRecipientMessage.scss
```diff
@@ -0,0 +1,60 @@
...
+      background-image: url('https://www.ft.com/__origami/service/image/v2/images/raw/fticon-v1%3Across?source=editor&tint=%23FFFFFF&bgcolor=%23FFFFFF');
+      background-repeat: no-repeat;
+      background-size: contain;
+      background-position: 50%;
+      background-color: transparent;
+      width: oSpacingByName('s6');
+      height: oSpacingByName('s6');
+      border: none;
+      padding: 0;
+      content: "";
```
GlynnPhillips, line 47: **Change Request:** I don't think `content` is needed here

```suggestion

```

components/x-gift-article/src/HighlightSavedMessage.jsx
```diff
@@ -0,0 +1,25 @@
...
+		>
+			<div className="o-message__container">
+				<div className="o-message__content">
+					<p className="o-message__content-main">
+						<span className="o-message__content-highlight">Highlights saved.</span>
+					</p>
+					<button
+						type="button"
+						className="o-message__close"
+						aria-label="close"
```
GlynnPhillips, line 17: **Suggestion:** Instead of relying on `aria-label` here it might be nicer to add text to the content but visually hide that text. That will provide more reliable results for assistive tech

components/x-gift-article/src/HighlightSection.scss
```diff
@@ -0,0 +1,38 @@
...
+		)
+	);
+}
+
+.x-gift-article__form .o-forms-field.x-gift-article__checkbox {
+	margin: 0;
+}
+
+.x-gift-article__checkbox-span::before {
+	background-color: inherit !important;
```
GlynnPhillips, line 28: **Change request:** Please avoid using `!important`

GlynnPhillips (changes requested):
I have left some comments that I believe need to be actioned before this can be merged

components/x-gift-article/src/HighlightSection.scss
MMSaleh-FT: Unfortunately this is the only way i found to enforce this style , any other way gets override by the form class the is imported in next article project for some reason , will discuss with the team if anyone have a solution for this

components/x-gift-article/src/HighlightRecipientMessage.jsx
```diff
@@ -19,6 +19,7 @@ export default ({ handleSaveAnnotations, handleCloseRecipientMessage }) => {
 			<button
 				className="x-gift-article__highlight-recipient-message__close"
 				type="button"
+				content="Close"
```
GlynnPhillips, line 22: **Change Request:** This isn't what I expected when I asked for the button to have text and as far as I am aware the `content` attribute isn't valid on a `button` element.

Please add the text to the button element eg `<button>Close</button>` and then use CSS to visually hide that text so sighted users can't see it but its still accessible to screen readers / assistive tech. We have a helper class for doing this called `o-normalise-visually-hidden` you can see it used [here](https://github.com/Financial-Times/x-dash/blob/4926cfbbf8de7b83615b1dd72413292cf29d3b48/components/x-teaser/src/Title.jsx#LL32C23-L32C50)

components/x-gift-article/src/HighlightSavedMessage.jsx
```diff
@@ -14,7 +14,7 @@ export default ({ handleCloseSuccessMessage }) => {
 					<button
 						type="button"
 						className="o-message__close"
-						aria-label="close"
+						content="close"
```
GlynnPhillips, line 17: Same response as [here](https://github.com/Financial-Times/x-dash/pull/709/commits/6d9dd17f5487368607a7892da0c606bd533ffc05#r1221533368)

components/x-gift-article/src/HighlightRecipientMessage.scss
```diff
@@ -5,56 +5,52 @@
...
-      border: none;
-      padding: 0;
-    }
-}
+	position: relative;
+	background-color: oColorsByName('slate');
+	color: oColorsMix('ft-grey', 'white', 20);
+	padding: oSpacingByName('s3');
+
+	.x-gift-article__highlight-recipient-message__title {
```
GlynnPhillips, line 13: **Change request:**** The classnames used through out this PR still don't conform to the BEM naming convention. This currently reads as `BLOCK__ELEMENT__ELEMENT` Maybe something like this would better match BEM `x-gift-article-highlight-recipient-message__title`

components/x-gift-article/src/HighlightRecipientMessage.scss
```diff
@@ -0,0 +1,60 @@
+@import '@financial-times/o-colors/main';
+@import '@financial-times/o-spacing/main';
+@import '@financial-times/o-typography/main';
+
+@include oTypography();
+
+.x-gift-article__highlight-recipient-message {
```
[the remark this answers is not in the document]
GlynnPhillips: The rest of the styles in this document still appear to be nested within this selector. When using BEM we shouldn't need to nest in this situation

GlynnPhillips (changes requested):
Some of the previous requests haven't been actions correctly. Feel free to reach out if you need further clarity

components/x-gift-article/src/HighlightSection.scss
```diff
@@ -0,0 +1,38 @@
...
+	);
+}
+
+.x-gift-article__form .o-forms-field.x-gift-article__checkbox {
+	margin: 0;
+
+	.o-forms-input__label.x-gift-article__checkbox-span::before {
+		background-color: inherit;
+	}
+}
```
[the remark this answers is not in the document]
jkerr321: Good point Glynn, thanks. @MMSaleh-FT let's remove this background colour style.

In an upcoming sprint we will be rebranding this modal with some new Origami styles anyway, which will conform to Origami rules.

components/x-gift-article/src/HighlightRecipientMessage.scss
jkerr321: For more context: https://getbem.com/naming/ . So yes I think we should precede each of the elements with (`__title`,` __description`,` __divider`) with `x-gift-article-highlight-recipient-message`

components/x-gift-article/src/HighlightSection.scss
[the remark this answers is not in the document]
MMSaleh-FT: i will discuss this with our designer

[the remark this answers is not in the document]
MMSaleh-FT: removed

GlynnPhillips (approved):
Thanks for actioning the feedback :)

PR #718, opened 2023-06-22
Make backToTop a Wrapper level prop for live blogs

magsallen:
Summary: Adds a `backToTop` prop to x-live-blog-wrapper

`backToTop` is currently an x-live-blog-post level argument but we don't really have a use case for this being post-level rather than live-blog level. I'm adding it in now so that I can use it in the new Content Pipeline UI that currently being developed - [relevant PR](https://github.com/Financial-Times/cp-content-pipeline/pull/383)

The prop can be either a string or a function to support the different implementations currently in place for dotcom and the App

Change is additive and non-breaking 🤞


apaleslimghost (approved):
nice, i like how straightforward this was in the end!

PR #739, opened 2023-09-06
change twitter icon to X

serena97:
Screenshot 2023-09-06 at 12 56 53


components/x-live-blog-post/src/svgIcons.jsx
```diff
@@ -6,8 +6,8 @@ Once this is resolved, these components should be replaced with o-share https://
  */
 export function TwitterSVG() {
```
AliceArmstrong, line 7: **Non-blocking suggestion:** Should we update this function name as well? Side-note, `X` is such an awkward name to use in functions 🙈

AliceArmstrong (commented):
Looks good, one non-blocking suggestion!

components/x-live-blog-post/src/svgIcons.jsx
AliceArmstrong: Ah, just realise this is probably used externally, so probably a lot of effort to rename it!

AliceArmstrong (approved):

emortong (approved):
Looks good, thanks!

ivomurrell (approved):
**question (non-blocking)**: should we change this copy to "Share on X" now? https://github.com/Financial-Times/x-dash/blob/da2f6ade0a9c162ebe51155e5b38dad826944410/components/x-gift-article/src/SocialShareButtons.jsx#L50 https://github.com/Financial-Times/x-dash/blob/da2f6ade0a9c162ebe51155e5b38dad826944410/components/x-live-blog-post/src/ShareButtons.jsx#L37

PR #744, opened 2023-10-02
Modify close button

AnnaDraganova:
## Description

- Since we don't use `o-overlay` for the sharing modal, we had to get rid of the `o-overlay__close` style for the close button. 
  - Created a custom style class and copied the style of the close button from `o-overlay` so the style of the button does not break in the future when the o-overlay gets updated. 
- The modal is closed with an event in `next-article`
  - Link to the PR in `next-article` https://github.com/Financial-Times/next-article/pull/5281

## Link to the Jira ticket
https://financialtimes.atlassian.net/browse/ENTST-644

### DIFF
--- a/components/x-gift-article/src/ShareArticleDialog.scss
+++ b/components/x-gift-article/src/ShareArticleDialog.scss
@@ -179,3 +179,39 @@
 		@include oTypographySans($weight: 'medium');
 	}
 }
+
+.share-article-modal__close {
+	display: inline-block;
+	background-repeat: no-repeat;
+	background-size: contain;
+	background-position: 50%;
+	background-color: transparent;
+	vertical-align: baseline;
+	width: 20px;
+	height: 20px;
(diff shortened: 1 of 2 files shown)


abshirahmed (approved):

components/x-gift-article/src/ShareArticleDialog.scss
```diff
@@ -179,3 +179,39 @@
 		@include oTypographySans($weight: 'medium');
 	}
 }
+
+.share-article-modal__close {
```
jkerr321, line 183: I think we should leave a short comment in the css here explaining that these styles have been copied directly from o-overlay close as we're no longer using o-overlay but want to preserve close functionality and a11y

AnnaDraganova: Here is the PR for the comment https://github.com/Financial-Times/x-dash/pull/746

PR #752, opened 2023-11-07
Update to date-fns version two

benwallman:
Updated to date-fns version two. This was fairly tricky, I'd recommend TDD-ing future date util libraries as it will force simplicity.

A really nice feature of date-fns version 2 is that it stops you handling dates, numbers & strings interchangeably, and forces you to use either dates or numbers. This is much easier to grep, and they provide utility modules to get a dates format in whichever string you like.

The previous code has a hack in, which is to get the first 10 characters of a datestring and use that to get the date. This is not the correct behaviour, as timezones are a thing. I've kept that hack in one place, but have included a small comment for future maintainers.

If this is your first `x-dash` pull request please familiarise yourself with the [contribution guide](https://github.com/Financial-Times/x-dash/blob/HEAD/contribution.md) before submitting.

## 

- Discuss features first
- Update the documentation
- Reference other issues
- Follow the code style


components/x-teaser-timeline/src/lib/transform.js
```diff
@@ -19,7 +19,11 @@ const groupItemsByLocalisedDate = (
 
 	items.forEach((item, index) => {
```
cerdfied, line 20: **question:** I can't see `groupItemsByLocalisedDate` being used anywhere..do we even need it? 🤷‍♀️ 
https://github.com/search?q=org%3AFinancial-Times%20groupItemsByLocalisedDate&type=code

cerdfied (approved):

components/x-teaser-timeline/src/lib/transform.js
```diff
@@ -19,7 +19,11 @@ const groupItemsByLocalisedDate = (
...
 		const localDateTime = getLocalisedISODate(item.publishedDate, timezoneOffset)
-		const localDate = getDateOnly(localDateTime)
+		// N.B. This is to keep a previous hack that existed. Changing this may break existing dates.
```
gyss, line 22: sorry Ben, we are not familiar with this codebase. To which hack are you referring?
the commit linked is just a refactor to be able to reuse `getDateOnly()`

components/x-teaser-timeline/src/lib/transform.js
```diff
@@ -19,7 +19,11 @@ const groupItemsByLocalisedDate = (
...
+		// https://github.com/Financial-Times/x-dash/commit/b4fcedcd3bcee721a7a03877b9bbcea18fdd4e1c
+		// This date won't represent the timezone. Using date-fns respects the timezone, so it's not used here.
+
+		const localDate = localDateTime.substring(0, 10)
```
gyss, line 26: How come this might break existing date if it's doing the same as before?

before we had this
```
export const getDateOnly = date => date.substr(0, 10);
```

benwallman: Sorry will clarify!

This PR moves the hack to use a substring to get the date to only be used once, in this file. Previously it was used wherever getDateOnly was used.

The reason I've kept it in this file is that removing this changes the snapshot test, and I'm not certain what the desired behaviour should be. If this is SSR, and we're not serving from the users region, then I'd rather not mess with it. If this is client side rendered, then I'm fairly confident with updating it.

gyss: I've done a [search in Github ](https://github.com/search?q=org%3AFinancial-Times+x-teaser-timeline&type=code) and there are a couple of `x-teaser` clients:

- ft-app - client
- next-myft-page - server

So `next-myft-page` would be affected, isn't it?

benwallman: Seemingly so!

I mistakenly did the same thing, but just with `x-teaser` and found it used on both the client and the server in next-article..

With this change I'll include some notes in the release notes, but do you think it's worth me checking / getting approval from the next-myft-page and the ft-app team?

gyss: next-myft-page it's us (content-disco) and ft-app that's the App team.

I was thinking, could you test this update to check how's working in `next-myft-page`?

benwallman: Apologies @gyss I'm unable to get the next-myft-page running!

It's showing the 'you need to install next-router' error, despite next-router seemingly being set up correctly (make test works in next-router directory).

Is this just a 'me' issue?

gyss: that's easy to fix. Just install next-router in local with the same `npm` and `node` versions than `next-myft-page`. Then npm link next-router into next-myft-page

benwallman: Ok so I've got past that issue with help from this thread: https://financialtimes.slack.com/archives/C041V9QA7/p1655480220646879

But that's just leading me to the next issue which is, make .env failing due to vault issues, despite being logged in.

Either way, how necessary is testing this in next-myft-page? It feels that testing middleware in each of it's consumers leads to a cascading level of effort.

Alternatively, are there any tests I could add / fix either here or in next-myft-page that would reassure us that these modules are working as expected?

gyss: Why would `make .env` have any error? did you run this command before in other projects?

Well, I'm not asking to test every one of the consumers, just one to check the component continues working as expected.

I'm concerned about this PR for some reasons:
- Not sure what's the benefit of this update
- Neither me or my teammates have enough background in this repo to safely approve it. That's why I'd like to make sure by testing
- Deploying a possible issue in the `ft-app` could be problematic since the release system is slower 

Apart from that, looks like there are plenty of snapshots tests in `x-teaser` and `x-teaser-timeline`. But, I was thinking if you could add some unit tests for `getLocalisedISODate()` and `getTitleForItemGroup()` so we can make sure at a lower level that the results are the same

benwallman: Not sure if I've ran `make .env` successfully, haven't needed to AFAIK

I've added a few tests for both those functions, and I ran these tests successfully against the previous version - Please take the horrendous screenshot below for proof!

image

For the benefits, it's hard to say without being able to build these apps. In next-article, version 1 of date-fns seems to be responsible for a large proportion of the client side JS:
iundle js

Adding tests to increase reliability and confidence is something I'm happy to do, so please let me know if there's any more I can add :)

gyss (approved):
That looks great! thanks Ben for adding the tests and the reasoning behind this PR. Not it looks much clearer

PR #753, opened 2023-11-14
Privacy Manager - Setting permutiveAds as a consent category value

andrewgeorgiou1981:
In preparation for the rollout of the new Sourcepoint based Consent Management System (CMP) there is a small update required to ensure that the CCPA consent toggle choices are respected in the ads-stack.
Specifically, when the new CMP is live, the ad-stack will check for the `permutiveAds` value in the FTCONSENT cookie / SCS when determining wether or not to run Permutive code.

This small PR ensures that the CCPA block/allow toggles will set the new 'permutiveAds' category value in the SCS and FTCONSENT cookie.


oliverturner (approved):
LGTM 👍

PR #757, opened 2023-12-12
include highlights to non-gift & gift shares

MMSaleh-FT:
This PR updates the behaviour of x-gift-article component to add the option to include highlights when using gift  & non-gift sharing for B2B users. The new logic is as follows:

- Option to 'include highlights' has been added to the share modal for gift article and non-gift share
- If a B2B user selects 'include highlights', a highlight share token is generated via the Professional user-annotations-api
- The highlights token is appended as a query param to the share url, along with the article share token, and is then shortened by bit.ly before being returned to the user

## Code changes

this PR only changes code in the x-gift-article component

### UI changes 

| Before | After |
| -------- | -------- |
| B2B users who have highlights in their articles can see the include highlights check box only when they use the advance share | B2B users who have highlights in their articles can see the include highlights check box for all sharing options
| Screenshot from 2023-12-18 15-34-52 | Screenshot from 2023-12-18 15-34-23

### Code changes 

API client Update:
- update the `getGiftUrl`  API to include an optional `highlightsToken` that will be used to add highlights when creating the Gift share link 
- add new function in the 'highlightsApi' to generate highlight share token that will be used for both Gift & non-gift shares

UI upates:
- Moved the include highlights checkbox from the advanced sharing options component to the parent share type selector component 
- change the condition show the highlights checkbox to be visible to all B2B users who have highlights in the article 

other updates:
- add logic to generate and include Highlight token in the action for creating  both Gift & non-Gift Share

### Gira 
[gira ticket](https://financialtimes.atlassian.net/browse/ENTST-353)

### DIFF
--- a/components/x-gift-article/src/AdvancedSharingOptions.jsx
+++ b/components/x-gift-article/src/AdvancedSharingOptions.jsx
@@ -4,16 +4,7 @@ import { NoCreditAlert } from './NoCreditAlert'
 import { ReceivedHighlightsAlert } from './ReceivedHighlightsAlert'
 
 export const AdvancedSharingOptions = (props) => {
-	const {
-		shareType,
-		actions,
-		showHighlightsCheckbox,
-		includeHighlights,
-		enterpriseHasCredits,
-		giftCredits,
-		showHighlightsRecipientMessage,
-		hasHighlights
-	} = props
+	const { shareType, actions, enterpriseHasCredits, giftCredits, showHighlightsRecipientMessage } = props
 	const onValueChange = (event) => {
 		if (event.target.value === ShareType.enterprise) {
 			actions.showEnterpriseUrlSection(event)
@@ -23,17 +14,12 @@ export const AdvancedSharingOptions = (props) => {
 		}
 	}
 
-	const includeHighlightsHandler = (event) => {
-		actions.setIncludeHighlights(event.target.checked)
-	}
-
 	return (
 		<div>
 			<div
 				className="o-forms-field o-forms-field--optional o-forms-field--professional share-article-dialog__advanced-sharing-options"
 				role="group"
-				aria-labelledby="radio-group-title"
-			>
+				aria-labelledby="radio-group-title">
 				<span className="o-forms-input o-forms-input--radio-round">
 					<span className="o-forms-input--radio-round__container">
 						<label htmlFor="share-with-multiple-people-radio">
@@ -70,23 +56,6 @@ export const AdvancedSharingOptions = (props) => {
 				</NoCreditAlert>
 			)}
 			{showHighlightsRecipientMessage && <ReceivedHighlightsAlert {...props} />}
-			{showHighlightsCheckbox && hasHighlights && (
-				<div className="o-forms-input o-forms-input--checkbox o-forms-field share-article-dialog__include-highlights">
-					<label htmlFor="includeHighlights">
-						<input
-							type="checkbox"
-							id="includeHighlights"
-							name="includeHighlights"
-							value={includeHighlights}
-							checked={includeHighlights}
-							onChange={includeHighlightsHandler}
-							disabled={shareType !== ShareType.enterprise}
-							data-trackable="make-highlights-visible"
-						/>
-						<span className="o-forms-input__label x-gift-article__checkbox-span">Include highlights</span>
-					</label>
-				</div>
-			)}
 		</div>
 	)
 }
--- a/components/x-gift-article/src/AdvancedSharingOptions.jsx
+++ b/components/x-gift-article/src/AdvancedSharingOptions.jsx
@@ -19,7 +19,7 @@ export const AdvancedSharingOptions = (props) => {
 			<div
 				className="o-forms-field o-forms-field--optional o-forms-field--professional share-article-dialog__advanced-sharing-options"
 				role="group"
-				aria-labelledby="radio-group-title">
+			>
 				<span className="o-forms-input o-forms-input--radio-round">
 					<span className="o-forms-input--radio-round__container">
 						<label htmlFor="share-with-multiple-people-radio">
--- a/components/x-gift-article/src/GiftArticle.jsx
+++ b/components/x-gift-article/src/GiftArticle.jsx
@@ -72,14 +72,13 @@ const withGiftFormActions = withActions(
 					}
 
 					const nonGiftUrl = new URL(state.urls.nonGift)
-						const { highlightsAccessToken } = await highlightsApiClient.shareHighlights(
-							initialProps.article.id,
-							state.includeHighlights
-						)
-						if (highlightsAccessToken) {
-							nonGiftUrl.searchParams.append('highlights',highlightsAccessToken);
-						}
-					
+					const { highlightsAccessToken } = await highlightsApiClient.shareHighlights(
+						initialProps.article.id,
+						state.includeHighlights
+					)
+					if (highlightsAccessToken) {
+						nonGiftUrl.searchParams.append('highlights', highlightsAccessToken)
+					}
 
 					const { url, isShortened } = await api.getShorterUrl(nonGiftUrl.toString())
 					tracking.createNonGiftLink(url, state.urls.nonGift)
--- a/components/x-gift-article/src/GiftArticle.jsx
+++ b/components/x-gift-article/src/GiftArticle.jsx
@@ -8,6 +8,7 @@ import * as updaters from './lib/updaters'
 import { ShareType } from './lib/constants'
 import ShareArticleDialog from './ShareArticleDialog'
 import { parsedSavedAnnotationsFromLocalStorage } from './lib/highlightsHelpers'
+import HighlightsApiClient from './lib/highlightsApi'
 
 const isCopySupported =
 	typeof document !== 'undefined' && document.queryCommandSupported && document.queryCommandSupported('copy')
@@ -19,6 +20,8 @@ const withGiftFormActions = withActions(
 			domain: initialProps.apiDomain
 		})
 		const enterpriseApi = new EnterpriseApiClient(initialProps.enterpriseApiBaseUrl)
+
+		const highlightsApiClient = new HighlightsApiClient()
 
 		return {
 			showGiftUrlSection() {
@@ -42,15 +45,22 @@ const withGiftFormActions = withActions(
 			},
 
 			async createGiftUrl() {
-				const { redemptionUrl, redemptionLimit } = await api.getGiftUrl(initialProps.article.id)
+				return async (state) => {
+					let response
+					const { highlightsAccessToken } = await highlightsApiClient.shareHighlights(
+						initialProps.article.id,
+						state.includeHighlights
+					)
+					response = await api.getGiftUrl(initialProps.article.id, highlightsAccessToken)
+					const { redemptionUrl, redemptionLimit } = response
+					if (redemptionUrl) {
+						const { url, isShortened } = await api.getShorterUrl(redemptionUrl)
+						tracking.createGiftLink(url, redemptionUrl)
 
-				if (redemptionUrl) {
-					const { url, isShortened } = await api.getShorterUrl(redemptionUrl)
-					tracking.createGiftLink(url, redemptionUrl)
-
-					return updaters.setGiftUrl(url, redemptionLimit, isShortened)
-				} else {
-					return updaters.setErrorState(true)
+						return updaters.setGiftUrl(url, redemptionLimit, isShortened)(state)
+					} else {
+						return updaters.setErrorState(true)
+					}
 				}
 			},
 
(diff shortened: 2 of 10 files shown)


components/x-gift-article/src/SharedLinkTypeSelector.jsx
```diff
@@ -70,14 +75,29 @@ export const SharedLinkTypeSelector = (props) => {
...
-						data-trackable="enterprise-out-of-credits"
-					>
+						data-trackable="enterprise-out-of-credits">
 						contact support
 					</a>
 					.
 				</NoCreditAlert>
 			)}
 			{showAdvancedSharingOptions && <AdvancedSharingOptions {...props} />}
+			{hasHighlights && (
```
jkerr321, line 85: **Question:** this should be limited just to B2B users - only B2B users should see the option to share highlights... is that the case? I know only B2B users can create highlights at the moment, but if a B2C user receives an article that contains highlights and then re-shares it, what would happen? Would they get the option to share the highlights? Would that cause an error if they did?

MMSaleh-FT: yeah, you are right, this would allow B2C customers to share highlights if it was shared to them , and it may cause issues since the owner of the highlights share  token and the article shared would be different in this case which will cause issues with redemption count for both of them

components/x-gift-article/src/SharedLinkTypeSelector.jsx
```diff
@@ -70,14 +75,29 @@ export const SharedLinkTypeSelector = (props) => {
...
-						data-trackable="enterprise-out-of-credits"
-					>
+						data-trackable="enterprise-out-of-credits">
 						contact support
 					</a>
 					.
 				</NoCreditAlert>
 			)}
 			{showAdvancedSharingOptions && <AdvancedSharingOptions {...props} />}
+			{hasHighlights && advancedSharingEnabled && (
```
jkerr321, line 85: I believe this should actually be `enterpriseEnabled` and not `advancedSharingEnabled` as we want _all_ b2b users to share highlights, not just advanced sharing users

jkerr321 (changes requested):

components/x-gift-article/src/SharedLinkTypeSelector.jsx
MMSaleh-FT: so the `advancedSharingEnabled` checks if `enterpriseEnabled` & `enterpriseRequestAccess` which i don't exactly know what does it mean , i will check  if it is only related to advance sharing , I will update the code to only check for `enterpriseEnabled`

jkerr321 (approved):

components/x-gift-article/src/lib/highlightsApi.js
```diff
@@ -54,4 +56,20 @@ export default class HighlightsApiClient {
...
+		try {
+			const response = await this.fetchJson('/create-token', {
+				method: 'POST',
+				body: JSON.stringify({ articleId })
+			})
+
+			if (response) {
+				return response
+			}
+			return {}
```
GlynnPhillips, line 70: **Question:** Given how the `fetchJson` function behaves is there any scenario where `response` would not be defined other than in a error scenario? https://github.com/Financial-Times/x-dash/blob/05680b0268774dbf50a4bfeb233390f7c44a5a47/components/x-gift-article/src/lib/highlightsApi.js#L28-L43

components/x-gift-article/src/lib/highlightsApi.js
```diff
@@ -54,4 +56,20 @@ export default class HighlightsApiClient {
...
+		} catch (error) {
+			return { shareToken: undefined }
```
GlynnPhillips, line 72: **Question:** Why would you return something different here to what you return above on line 70? This comment might become redundant depending on my [question on line 70 ](https://github.com/Financial-Times/x-dash/pull/757/files#r1439310018) but it would be good to keep consistency so raising it anyway

components/x-gift-article/src/lib/api.js
```diff
@@ -43,9 +43,16 @@ export default class ApiClient {
...
 	}
 
-	async getGiftUrl(articleId) {
+	async getGiftUrl(articleId, highlightsToken = undefined) {
 		try {
-			const json = await this.fetchJson('/article/gift-link/' + encodeURIComponent(articleId))
+			let json
+			if (highlightsToken) {
+				json = await this.fetchJson(
+					'/article/gift-link/' + encodeURIComponent(articleId) + `?highlightsToken=${highlightsToken}`
```
GlynnPhillips, line 51: **Change request:** Here we are using a mixture of string concatenation (`'' + ''`) and template literals (``${}``) please could we keep this consistent and only use a single way of producing the string. I would suggest template literals

 ```suggestion
					`/article/gift-link/${encodeURIComponent(articleId)}?highlightsToken=${highlightsToken}`
```

File: components/x-gift-article/src/lib/api.js [hunk 6]
```diff
@@ -43,9 +43,16 @@ export default class ApiClient {
...
+				)
+			} else {
+				json = await this.fetchJson('/article/gift-link/' + encodeURIComponent(articleId))
```
GlynnPhillips, line 54: **Note:** Along with the above comment on like 51 lets also make this consistent with whatever we choose for that comment

components/x-gift-article/src/SharedLinkTypeSelector.jsx
```diff
@@ -11,7 +11,9 @@ export const SharedLinkTypeSelector = (props) => {
 		enterpriseRequestAccess,
 		showAdvancedSharingOptions,
 		enterpriseHasCredits,
-		giftCredits
+		giftCredits,
+		hasHighlights,
+		includeHighlights
```
GlynnPhillips, line 16: **Question:** Could you please show me where these new props come from? I have tried to trace them back but I can't quite find the starting point. As they are used for stuff like `checked` and `value` attributes I am wondering if we need to be defaulting them or type checking them to ensure the values are correct

components/x-gift-article/src/GiftArticle.jsx
```diff
@@ -59,7 +72,18 @@ const withGiftFormActions = withActions(
 						state.showFreeArticleAlert = false
 						return state
 					}
-					const { url, isShortened } = await api.getShorterUrl(state.urls.nonGift)
+
+					let nonGiftUrl = state.urls.nonGift
```
GlynnPhillips, line 76: **Change Request:** Here I think we can avoid the mutation of `let`  and ensure the type never changes by instead use a `URL` constructor to produce the final url.

This would also avoid issues in the future should the search params change such as the original `state.urls.nonGift` not including a param

Something like the below

```js
const nonGiftUrl = new URL(state.urls.nonGift);

if (state.includeHighlights) {
	const { highlightsAccessToken } = await highlightsApiClient.shareHighlights(
		initialProps.article.id
	)
	if (highlightsAccessToken) {
		nonGiftUrl = `${nonGiftUrl}&highlights=${highlightsAccessToken}`
	}
}

const { url, isShortened } = await api.getShorterUrl(nonGiftUrl.toString())
```

components/x-gift-article/src/GiftArticle.jsx
```diff
@@ -41,15 +44,25 @@ const withGiftFormActions = withActions(
...
+					let response
+					if (state.includeHighlights) {
+						const { highlightsAccessToken } = await highlightsApiClient.shareHighlights(
+							initialProps.article.id
+						)
+						response = await api.getGiftUrl(initialProps.article.id, highlightsAccessToken)
+					} else {
+						response = await api.getGiftUrl(initialProps.article.id)
+					}
+					const { redemptionUrl, redemptionLimit } = response
```
GlynnPhillips, line 57: **Suggestion:** I would if this could be simplified if `shareHighlights` accepted `state.includeHighlights` as a prop and exited early if it was falsey? That would look something like the below, what do you think?

This function

```js
const { highlightsAccessToken } = await highlightsApiClient.shareHighlights(
	initialProps.article.id
);

const { redemptionUrl, redemptionLimit } = await api.getGiftUrl(initialProps.article.id, highlightsAccessToken);

```

shareHighlights function could have this at the start

```js
async shareHighlights({ articleId, includeHighlights} = props) {
    if (!includeHighlights) {
        return {};
    }
    // rest of function
}
```

GlynnPhillips: Or another alternative would be to move the includesHighlights logic into the `getGiftUrl` function.

That function would become something like

```js
async getGiftUrl(articleId, includeHighlights = false) {
    
    let highlightsToken = undefined;
    if (includeHighlights) {
         const { highlightsAccessToken } = await highlightsApiClient.shareHighlights(
		initialProps.article.id
	);
	
	highlightsToken = highlightsAccessToken;
    }
    
    // rest of code
```

and this function would become just

```js
const { redemptionUrl, redemptionLimit } = await api.getGiftUrl(initialProps.article.id, state.includeHighlights);
```

GlynnPhillips (changes requested):
i have left some comments, happy to discuss them further with you if you would like as I might have misunderstood some of it

components/x-gift-article/src/SharedLinkTypeSelector.jsx
MMSaleh-FT: these are not new props , I just moved them from the `AdvancedSharingOptions` to this file since the highlights are no longer exclusive to advance sharing users.

As for where these props come from they are assign when the tool-tip is initialized on next-article
https://github.com/Financial-Times/next-article/blob/42b329f65709ec05984422cc463663d6903e36b6/client/components/gift-article/index.jsx#L125-L143

they also have default values in the initalState in the Gift article component https://github.com/Financial-Times/x-dash/blob/e9062b0215b4593155cdaec985880bf86d67a62f/components/x-gift-article/src/GiftArticle.jsx#L264-L266

components/x-gift-article/src/GiftArticle.jsx
MMSaleh-FT: I think the first option is better since in the second options we will be making two api calls in one function which is not ideal

components/x-gift-article/src/lib/highlightsApi.js
MMSaleh-FT: Hmm, I don't think so , since if the `response` is undefined it will through when calling  `response.json()` , will remove the return statement and the if condition then

components/x-gift-article/src/GiftArticle.jsx
GlynnPhillips: Sorry I made a mistake when writing this comment. Just editing the suggestion now so it makes sense. I forgot to change something that I copy and pasted from your code

GlynnPhillips: That should be better now

components/x-gift-article/src/lib/highlightsApi.js
MMSaleh-FT: yeah, i think just using `return;` should be good

components/x-gift-article/src/SharedLinkTypeSelector.jsx
GlynnPhillips: Thanks

components/x-gift-article/src/lib/highlightsApi.js
GlynnPhillips: @MMSaleh-FT if you change it to `return;` without also changing how its used [here](https://github.com/Financial-Times/x-dash/pull/757/files#diff-7082a9e3cb317538577308da954ff347e6c58e103516faaa15ec7c405cba1161R49) it will result in a destructuring error.

components/x-gift-article/src/GiftArticle.jsx
MMSaleh-FT: yeah, that looks good, will update the PR , thanks alot!

components/x-gift-article/src/AdvancedSharingOptions.jsx
```diff
@@ -23,17 +14,12 @@ export const AdvancedSharingOptions = (props) => {
...
-	}
-
 	return (
 		<div>
 			<div
 				className="o-forms-field o-forms-field--optional o-forms-field--professional share-article-dialog__advanced-sharing-options"
 				role="group"
-				aria-labelledby="radio-group-title"
-			>
+				aria-labelledby="radio-group-title">
```
GlynnPhillips, line 22: **Change request:** Sorry I think you might be getting unlucky here but your change highlighting an already existing issue.

The `aria-labelledby` attribute seems to be pointing to an non-existence element. I can't see anything in this repo with an `id` of `radio-group-title`

GlynnPhillips: I think we can probably just remove the attribute

components/x-gift-article/src/lib/highlightsApi.js
```diff
@@ -54,4 +56,19 @@ export default class HighlightsApiClient {
...
+			}
+			return await this.fetchJson('/create-token', {
+				method: 'POST',
+				body: JSON.stringify({ articleId })
+			})
+
+		} catch (error) {
+			return{};
+		}
+	}
```
GlynnPhillips, line 73: **Question / change request** Has linting correctly been run on these changes? I would have expected some linting errors here based on this projects eslint config.

I would have expected the `keyword-spacing` rules to fail here on stuff like `if(` and `return{}`

GlynnPhillips: If you run linting on this and it passes then fine to leave it but if it fails then please update it

MMSaleh-FT: hmmm , yeah you are right , for some reason the pre-commit script didn't run , take a look at this

components/x-gift-article/src/AdvancedSharingOptions.jsx
MMSaleh-FT: we probably removed the element with that Id when we updated the gift article component a some time ago , will check the commit history for more info and remove this lable for now

GlynnPhillips (approved):
Thanks for your patience

PR #764, opened 2024-02-05
ENTST 788 | Remove save message for highlighted article with different content

ob6160:
## What does this change?

- Don't show the highlights recipient message if the article content has changed since the shared article highlights were first added
- Adds a test to ensure that the save the highlights message is not displayed when the article content has changed since highlights were added

## How was this tested?

- Locally using an instance of `next-article` with the `x-gift-article` package linked.

### Before

Screenshot 2024-02-05 at 16 42 19

### After

Screenshot 2024-02-05 at 16 44 08

### DIFF
--- a/components/x-gift-article/src/GiftArticle.jsx
+++ b/components/x-gift-article/src/GiftArticle.jsx
@@ -276,9 +276,19 @@ const withGiftFormActions = withActions(
 		const userHasNotYetSavedSharedAnnotations =
 			!parsedSavedAnnotationsFromLocalStorage().includes(highlightsToken)
 
-		// We only want to display the highlights recipient message if the user is a
-		// highlights recipient and has not yet saved the shared annotations for this highlight token.
-		const showHighlightsRecipientMessage = userIsAHighlightsRecipient && userHasNotYetSavedSharedAnnotations
+		// We use the absence of the missing-highlight-message class to determine
+		// if the article has not changed since the highlights were shared.
+		const articleHasNotChangedSinceHighlightsWereShared =
+			document.querySelector('.missing-highlight-message') === null
+
+		// We only want to display the highlights recipient message if:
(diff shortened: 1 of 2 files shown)


components/x-gift-article/src/GiftArticle.jsx
```diff
@@ -276,9 +276,19 @@ const withGiftFormActions = withActions(
 		const userHasNotYetSavedSharedAnnotations =
 			!parsedSavedAnnotationsFromLocalStorage().includes(highlightsToken)
 
-		// We only want to display the highlights recipient message if the user is a
-		// highlights recipient and has not yet saved the shared annotations for this highlight token.
-		const showHighlightsRecipientMessage = userIsAHighlightsRecipient && userHasNotYetSavedSharedAnnotations
+		// We use the absence of the missing-highlight-message class to determine
+		// if the article has not changed since the highlights were shared.
```
MMSaleh-FT, line 280: Just a note on this, I think that  the `missing-highlight-message` doesn't detect if the article changed but if the section that was highlighted was removed,
and here is the check that displays the `missing-highlight-message`.
https://github.com/Financial-Times/user-annotations-editor/blob/d43f62e0c6ffa925d1a169eab21c160a9e8cce8d/src/locateAnnotation.ts#L36-L42

So, maybe changing the comment and variable name to reflect that , just to avoid any confusion in the future

ob6160: good shout - I'll update the code to indicate this detail :)

ob6160: I've pushed a commit that updates the variable name and comment here @MMSaleh-FT - sorry for delay, got sidetracked by tracking 🛤️

ytcleon (approved):

PR #765, opened 2024-02-07
add optional prop to make article save button icon only

MMSaleh-FT:
PR to add a property to `x-article-save-button` to make it icon only 

| Before  | After |
| ------------- | ------------- |
|Screenshot from 2024-02-07 18-16-20|Screenshot from 2024-02-07 19-21-02|

### DIFF
--- a/components/x-article-save-button/src/ArticleSaveButton.jsx
+++ b/components/x-article-save-button/src/ArticleSaveButton.jsx
@@ -41,7 +41,7 @@ export const ArticleSaveButton = (props) => {
 				aria-pressed={props.saved}
 			>
 				<span className="x-article-save-button__icon" />
-				{props.saved ? 'Saved' : 'Save'}
+				{props.iconOnly ? '' : props.saved ? 'Saved' : 'Save'}
 			</button>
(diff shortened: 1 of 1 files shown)


components/x-article-save-button/src/ArticleSaveButton.jsx
```diff
@@ -41,7 +41,7 @@ export const ArticleSaveButton = (props) => {
 				aria-pressed={props.saved}
 			>
 				<span className="x-article-save-button__icon" />
-				{props.saved ? 'Saved' : 'Save'}
+				{!props.iconOnly ? (props.saved ? 'Saved' : 'Save') : ''}
```
ob6160, line 44:
```suggestion
				{props.iconOnly ? '' : (props.saved ? 'Saved' : 'Save')}
```

_nit:_ might be a bit clearer if we invert the logic here and avoid having the logical not operator against `!props.iconOnly`?

ob6160 (approved):
totally optional comment from me :shipit:

joelcarr (approved):

PR #806, opened 2024-12-11
HETD-99 - add context tracking for storygroups

gyss:
## Description

add context link tracking for storygroups  teasers
 
- story heading
- story standfirst
- story image

Author is not a link, so can't be tracked
-  story author


metart43 (commented):
If we already have this information in `data-trackable` why do we need an exatra property called `data-trackable-context-story-link`?

gyss:
@metart43  this is needed because the storygroup tracking data is stored inside the `context` object

In this screenshot you can see the expected data by Spoort. The `story-link` prop is inside the `context`
Screenshot 2024-12-11 at 11 48 18

metart43 (approved):

PR #819, opened 2025-03-18
chore: remove unused live blog components

apaleslimghost:
these components have been moved into `cp-content-pipeline` and are no longer in use.


olusoladavid (approved):
In Kara we trust 😄 - changes look good, I believe changes have been tested as well. Thank you 🥂

umbobabo (approved):

PR #824, opened 2025-04-09
Reinstate and migrate x-privacy-manager

apaleslimghost:
it proved to hard to unpick where it's being used in `ft-app` and it causes peer dep conflicts so

probably easiest to review commit-by-commit


danieleruiz (approved):
LGTM!

PR #826, opened 2025-05-08
fix: Fix x-privacy-manager to respond to interaction

Nazehs:
Fixed radio button component in privacy manager  by reverting the class that was removed during o3 migration for it to respond to user interaction.

This was spotted during the work of https://financialtimes.atlassian.net/browse/CPO3-39

### DIFF
--- a/components/x-privacy-manager/src/components/radio-btn.jsx
+++ b/components/x-privacy-manager/src/components/radio-btn.jsx
@@ -19,7 +19,7 @@ export function RadioBtn({ name, type, checked, trackingKeys, buttonText, onChan
 	return (
 		<div className="x-privacy-manager-radio-button">
 			<input
-				className="o3-visually-hidden"
+				className="o3-visually-hidden x-privacy-manager-radio-button__input"
 				id={id}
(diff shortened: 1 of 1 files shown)


ivomurrell (commented):
**question**: i can't see anywhere in the code where adding `x-privacy-manager-radio-button__input` will change the element's behaviour. i'm rubbish at CSS but none of the selectors [here](https://github.com/Financial-Times/x-dash/blob/c12fccb30797b93d3a0c871c168a418e821c5e95/components/x-privacy-manager/src/components/radio-btn.scss) seem to match this element to me? am i missing something?

ivomurrell (approved):
ah sorry i think i understand the CSS now 😅 i guess those are independent selectors in the nested blocks and don't have to be children? sorry my CSS is rubbish like i said!

Nazehs:
> ah sorry i think i understand the CSS now 😅 i guess those are independent selectors in the nested blocks and don't have to be children? sorry my CSS is rubbish like i said!

That make the 2 of us 😊

